PR #89, opened 2014-11-14
Added method to get a column from a table node.

PurpleBooth:
Used by: https://github.com/Behat/Behat/pull/654

Allows you to get a column of a table node by index.

For example:

The Gherkin

``` gherkin
Given I am user:
  | username | ever.zet |
  | age      | 22       |
```

What that looks like in PHP

``` php
$table = new TableNode(array(
        array('username', 'ever.zet'),
        array('age', '22'),
    ));
$table->getColumn(0);
//=> array('username', 'age')
```

### DIFF
--- a/src/Behat/Gherkin/Node/TableNode.php
+++ b/src/Behat/Gherkin/Node/TableNode.php
@@ -157,6 +157,31 @@ class TableNode implements ArgumentInterface, IteratorAggregate
     }
 
     /**
+     * Returns specific column in a table.
+     *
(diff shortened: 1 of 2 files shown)


rajeshtaneja:
This is causing our case to fail, we have a step which handles rows where 1st column is the field with rest of the column as value and they may vary.
It looks obvious to have table with equal rows, but as tableNode is passed to step, it should not be forced but should be handled by Step.

Just for your consideration.

Cheers.

PR #99, opened 2015-11-18
Ensure keywords are trimmed when syncing translations

stof:
The lexer parses keywords with spaces following them, and then trims the result to get only the keyword. This assumes that the keyword itself was trimmed, otherwise the detection of the keyword type breaks (and the tests break too).
The Tamil translations were including spaces in the translations. This trims them during the translation update (better than trimming them each time they are loaded, even though it would be the safer way).

Note that I have not updated translation here (this would be done separately). I ran the script by altering a bit to load our local i18n.php file instead of the cucumber i18n.json one.Only the ta locale is actually updated (and BC for users as spaces are allowed in any number after step keywords). Most changes in the i18n.php file are actually due to the fact that PHP sorting is not stable for strings which have the same length


stof:
@everzet what do you think about this solution ?

everzet:
I like the approach and it makes sense. Are there any blockers?

stof:
no. I only wanted to know whether you preferred another approach ensuring things works fine all the time, including when people load translations from a different source (i.e. trimming at load time).
The testsuite is green, which proves that the fix is working (testsuite is broken in master)

PR #100, opened 2015-12-18
Add a reference to OutlineNode for ExampleNode

higrow:
It is currently not possible to get the ScenarioOutline from an ExampleNode (like in a Scenario Hook)

This change adds a reference to the OutlineNode into the ExampleNode so there is a possibility to get the ScenarioOutline if it is needed so you could to this:

```php
public function afterScenario(AfterScenarioScope $scope)
    {
        if ($scope->getScenario() instanceof ExampleNode) {
            $title = $scope->getScenario()->getOutline()->getTitle();
        }
    }
```

there is a issue with the same basic requirements here, however with another approach: behat/behat#780

### DIFF
--- a/src/Behat/Gherkin/Node/ExampleNode.php
+++ b/src/Behat/Gherkin/Node/ExampleNode.php
@@ -41,23 +41,29 @@ class ExampleNode implements ScenarioInterface
      * @var null|StepNode[]
      */
     private $steps;
+    /**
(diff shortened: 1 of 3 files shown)


everzet:
This introduces cyclic reference into the Gherkin tree - something we precisely tried to eliminate in 3.0 for performance reasons

PR #101, opened 2015-12-31
Fix the handling of file names in CucumberKeywords

stof:
The Yaml parser does not accept file names anymore in Symfony 3 (and this is deprecated in Symfony 2).
File names are now handled in the CucumberKeywords class directly now, copying the logic available in symfony/yaml 2.x, as we want to continue accepting a filename.

This was the reason of the test failure for the Symfony 3 jobs (and for other jobs after the addition of the phpunit-bridge)


stof:
Yeah, testsuite is green again.

PR #102, opened 2016-01-28
Update i18n.php

mirsoo:
Added language support for Bosnian

### DIFF
--- a/i18n.php
+++ b/i18n.php
@@ -59,6 +59,21 @@
     'then' => 'اذاً|ثم',
     'when' => 'عندما|متى',
   ),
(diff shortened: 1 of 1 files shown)


stof:
We don't accept contributions on the i18n.php file itself (as explained in our contributing guidelines), because this file is generated automatically based on the translations available in cucumber's gherkin project. Translation contributions should go there instead: https://github.com/cucumber/gherkin/blob/master/gherkin-languages.json
Can you send a PR to this repo instead ?

mirsoo:
I can try.

PR #107, opened 2016-06-16
Azerbaijani translation has been added

nalekberov:


stof:
We don't accept contributions on the i18n.php file itself (as explained in our contributing guidelines), because this file is generated automatically based on the translations available in cucumber's gherkin project. Translation contributions should go there instead: https://github.com/cucumber/gherkin/blob/master/gherkin-languages.json

stof:
Please send a PR to the cucumber/gherkin repo instead

i18n.php
```diff
@@ -59,6 +59,21 @@
     'then' => 'اذاً|ثم',
     'when' => 'عندما|متى',
   ),
+  'az' => 
+  array (
+    'and' => 'Və/Həm',
```
stof: it is expected to have a ``/`` here ? shouldn't it be 2 alternatives (and so a ``|``) ?

But this allowed me to catch a bug in gherkin btw.

stof:
I'm closing this PR as it cannot be accepted here

i18n.php
nalekberov: My bad. It is not allowed I know. Didn't check. By the way I sent PR to cucumber/gherkin and they merged it.

Issue #103, opened 2016-02-06
Translations repository


dudemelo:
The current [URL to i18n.json](https://github.com/Behat/Gherkin/blob/master/bin/update_i18n#L4) is broken because the gherkin2 repository has been moved. Will we use the new gherkin2 repository or keep using the gherkin3?

stof:
We should update our script to read translations from https://github.com/cucumber/gherkin/blob/master/gherkin-languages.json instead of the gherkin2 repo. All gherkin 2 translations have been synchronized into gherkin 3, but gherkin 2 is now unmaintained and so translations won't be updated anymore.
Note that the gherkin 3 repo uses a different format for their language file, so it is not just a matter of updating the URL.

I will do this update for the next release of the library (as I will need to run the script anyway).

PR #111, opened 2016-06-23
Add asterisk support , flowing gherkin specification.

lxbzmy:


src/Behat/Gherkin/Lexer.php
```diff
@@ -325,7 +325,7 @@ protected function getKeywords($type)
                 foreach (explode('|', $keywords) as $keyword) {
                     $padded[] = false !== mb_strpos($keyword, '<', 0, 'utf8')
                         ? mb_substr($keyword, 0, -1, 'utf8') . '\s*'
-                        : $keyword . '\s+';
+                        : preg_quote($keyword) . '\s+';
```
stof: please don't duplicate the work of the existing PR, especially given that your fix is incomplete

i18n.php
```diff
@@ -921,7 +921,7 @@
     'but' => 'Nhưng',
     'examples' => 'Dữ liệu',
     'feature' => 'Tính năng',
-    'given' => 'Biết|Cho',
+    'given' => '*|Biết|Cho',
```
stof: this file must not be edited by hand. We have a script generating it (which needs to be updated btw, which is on my TODO list). Manual updates on this file will be rejected

src/Behat/Gherkin/Lexer.php
lxbzmy: Test case passed.

i18n.php
lxbzmy: update_i18n 
https://raw.github.com/cucumber/gherkin/master/lib/gherkin/i18n.json 
404 not found.

lxbzmy: change to this?https://github.com/cucumber/gherkin/blob/master/ruby/lib/gherkin/gherkin-languages.json

src/Behat/Gherkin/Lexer.php
stof: yeah, for the asterisk. But it misses some escaping

i18n.php
stof: See the discussion in https://github.com/Behat/Gherkin/issues/103

I will update the script this weekend to read the new i18n file.

everzet:
Can you rebase, please?

stof:
Closing because #112 actually made it (and with full support)

PR #112, opened 2016-06-25
Rewrite the translation update script

### LINKED ISSUE #103: Translations repository

stof:
It now reads translations from the gherkin 3 repository rather than the gherkin 2 repository.
The format of the i18n.php file is kept the same, for backward compatibility reasons. The next major release of Gherkin is expected to change the KeywordsInterface to return an array of keywords, and keeping the trailing space in step keywords instead of the old gherkin2 syntax.

Closes #103

this also preserves the ``*`` keyword, see #111


stof:
@everzet to keep tests running, I had to skip asterisks in the i18n tests (because the keyword type is ambiguous in this case), which implied skipping them in the dumper

everzet:
Awesome! TY!!!

PR #118, opened 2016-11-14
Add outlineTitle to ExampleNode

duxet:
It's needed for BeforeScenarioScope to get the title of the scenario, as at the moment there is only example title provided.

### DIFF
--- a/src/Behat/Gherkin/Node/ExampleNode.php
+++ b/src/Behat/Gherkin/Node/ExampleNode.php
@@ -41,23 +41,29 @@ class ExampleNode implements ScenarioInterface
      * @var null|StepNode[]
      */
     private $steps;
+    /**
+     * @var string
+     */
+    private $outlineTitle;
 
     /**
      * Initializes outline.
      *
-     * @param string     $title
-     * @param string[]   $tags
-     * @param StepNode[] $outlineSteps
-     * @param string[]   $tokens
-     * @param integer    $line
+     * @param string      $title
+     * @param string[]    $tags
+     * @param StepNode[]  $outlineSteps
+     * @param string[]    $tokens
+     * @param integer     $line
+     * @param string|null $outlineTitle
      */
-    public function __construct($title, array $tags, $outlineSteps, array $tokens, $line)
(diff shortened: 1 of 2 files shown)


sskorc:
LGTM :+1:

stof:
How is cucumber's gherkin generating the scenario title for outline-based scenarios ? It would be better to match their title than exposing the outline title separately IMO.

duxet:
@stof ExampleNode is being passed directly to the Behat hook scope:

sskorc:
@stof It seems like their scenario title (or scenario name, how they call it) is the specific outline example value (so, it's the same value as it is now in Behat in `$title` field). But they also have a separate scenario outline name field for the whole scenario title (just like `$outlineTitle` in @duxet's change). 

Here's the StackOverflow questions that proves it: http://stackoverflow.com/questions/11209066/feature-and-scenario-outline-name-in-cucumber-before-hook/11209068#11209068

duxet:
@stof Does @sskorc reply explains everything you wanted to know?

I can add from myself that this PR solves Behat/Behat#780 issue.

tikolakin:
It would be good to have both:
- get ScenarioOtline title
- keep getting examples

sskorc:
@tikolakin Sure, and this PR covers it. You can get an example title by `getTitle()` (as it is now) and you can also get a scenario outline title by calling `getOutlineTitle()`.

everzet:
Can you make `$outlineTitle` argument parameter `null`able for BC?

duxet:
@everzet done :)

duxet:
Tests have failed because of deprecation warnings caused by changes in Symfony YAML parser:

```
Duplicate key "type" detected on line 18 whilst parsing YAML. Silent handling of duplicate mapping keys in YAML is deprecated since version 3.2 and will throw \Symfony\Component\Yaml\Exception\ParseException in 4.0: 1x
    1x in ParserTest::testParser from Tests\Behat\Gherkin
```

They are not related to changes made in this pull request, so it can be safely merged anyway.

everzet:
@duxet Thank you so much for hard work!

PR #120, opened 2017-01-30
Add TableNode::fromList() method

TravisCarden:
I frequently want to create a `TableNode`, for comparison, from a flat list (or a one-dimensional array). A simple little factory method would be very helpful.


everzet:
Thank you!

PR #121, opened 2017-02-01
Handle non-two-dimensional arrays in TableNode

TravisCarden:
The UX around constructing a `TableNode()` could be improved in terms of ensuring two-dimensional arrays. Presently the errors are not very illuminating:

Given a one-dimensional array:
```
Invalid argument supplied for foreach()
 vendor/symfony/phpunit-bridge/DeprecationErrorHandler.php:73
 src/Behat/Gherkin/Node/TableNode.php:55
```

Given a three-dimensional (or deeper) array:
```
mb_strlen() expects parameter 1 to be string, array given
 vendor/symfony/phpunit-bridge/DeprecationErrorHandler.php:73
 src/Behat/Gherkin/Node/TableNode.php:65
```

I propose the much clearer `NodeException('Table is not two-dimensional.')`.

### DIFF
--- a/src/Behat/Gherkin/Node/TableNode.php
+++ b/src/Behat/Gherkin/Node/TableNode.php
@@ -35,8 +35,8 @@ class TableNode implements ArgumentInterface, IteratorAggregate
      * Initializes table.
      *
      * @param array $table Table in form of [$rowLineNumber => [$val1, $val2, $val3]]
-     * 
-     * @throws NodeException If the number of columns is not the same in each row
+     *
+     * @throws NodeException If the given table is invalid
      */
     public function __construct(array $table)
     {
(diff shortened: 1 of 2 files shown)


src/Behat/Gherkin/Node/TableNode.php
```diff
@@ -52,11 +52,19 @@ public function __construct(array $table)
...
+            if (!is_array($row)) {
+                throw new NodeException('Table is not two-dimensional.');
+            }
+
             foreach ($row as $column => $string) {
                 if (!isset($this->maxLineLength[$column])) {
                     $this->maxLineLength[$column] = 0;
                 }
 
+                if (is_array($string)) {
```
stof: what about ``!is_string`` ?

TravisCarden: @stof I tried `!is_string()` first, but a whole bunch of unrelated unit tests failed because they were passing integers. I could change the variable name from `$string` to `$value`, since the assumption isn't necessarily valid.

stof: hmm, maybe using ``is_scalar`` rather than ``is_string`` then ?

Your code does not forbid putting arrays

TravisCarden: Yeah, that works. And all the tests pass. Updating PR...

TravisCarden: Better, @stof?

everzet:
This is perfect! Sorry for late response.

PR #123, opened 2017-02-16
Updated reference to cucumber/gherkin’s translation file

chunfeilung:
According to CONTRIBUTING.md, the i18n.php file is still generated from
https://github.com/cucumber/gherkin/blob/master/lib/gherkin/i18n.json,
which no longer exists.

Because the actual translation file used by update_i18n is
https://raw.github.com/cucumber/gherkin/master/gherkin-languages.json,
I’ve updated CONTRIBUTING.md so that it points contributors to the correct
JSON file again.


stof:
the raw.github.com URL is the one to get the raw content rather than the Github UI showing it. We don't want to send users there

PR #126, opened 2017-07-03
Use realpath, so the feature receives the cwd prefixed.

glennunipro:
See the PR in the Behat repo for more context: behat/behat#1056.

As the path filter checks that the feature's file path _starts_ with a given whitelist of directories, we wrap the feature file in a realpath call -- this will either:

1. return false if the Feature file path is relative, but does not exist
2. return the full path to the Feature file, if it exists within the `cwd`
3. return the same thing if the Feature file path is already absolute and pointing to an existing file

This fixes a bug raised in the behat/behat repo, with another PR to implement part 2 of the fix.

### DIFF
--- a/src/Behat/Gherkin/Filter/PathsFilter.php
+++ b/src/Behat/Gherkin/Filter/PathsFilter.php
@@ -50,7 +50,7 @@ class PathsFilter extends SimpleFilter
     public function isFeatureMatch(FeatureNode $feature)
     {
         foreach ($this->filterPaths as $path) {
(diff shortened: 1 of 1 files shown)


everzet:
@glennunipro thank you very much!

pfrenssen:
It seems to be the intention that `$feature->getFile()` always contains a relative path, so this is not the right solution.

pfrenssen:
It is not clear if this paths is supposed to be absolute or relative, apparently some tests assume it is relative and some assume it is absolute, ref #129.

PR #127, opened 2017-08-09
Add belarus translation

gernovich:


ciaranmcnulty:
This looks good but the file is auto-generated from the upstream Cucumber translations so that we stay in sync:
https://raw.github.com/cucumber/gherkin/master/gherkin-languages.json

Can you propose your changes there instead?

stof:
Actually, the file has been moved to https://github.com/cucumber/cucumber/blob/master/gherkin/gherkin-languages.json

Issue #128, opened 2017-09-13
Gherkin cache invalid since v4.5


mihaeu:
Hey everyone, thanks for this awesome project!

We noticed that our Behat tests on Jenkins were slowly dying and after a couple of hours of debugging we noticed that the problem has something to do with outdated caches since the 4.5 update.

What happened was that you could run the whole unit as one, but when you tried to launch a single .feature file Behat was unable to find any specifications.

I deleted the caches and downgraded to Gherkin v4.4 and everything is back to normal. Did we miss something or was there a breaking change?

Cheers,
Mike

pfrenssen:
I'm having the same issue. I cannot run individual tests, getting this error:

```
 [Behat\Testwork\Tester\Exception\WrongPathsException]                                                                                           
  No specifications found at path(s) `features/authentication.feature`. This might be because of incorrect paths configuration in your `suites`.
```

pfrenssen:
Note for people who may not be able to run their tests even after rolling back to 4.4.5: delete the Gherkin cache, this is located in your temporary storage, e.g. in `/tmp/behat_gherkin_cache/` or similar.

pfrenssen:
I've been experimenting with this. The bug is apparently not in `behat/gherkin` but in `behat/behat`. The bug only occurs if I am using Behat 3.4.0, regardless of the version of Gherkin.

pfrenssen:
I did a `git bisect` and the bug was introduced in this commit from `behat/behat`: https://github.com/Behat/Behat/commit/08802b7f2d736b88df339e5411a1c722e92adbe3

This was part of https://github.com/Behat/Behat/pull/1056. We should move this discussion to `behat/behat`.

pfrenssen:
It appears that the bug is in Gherkin after all. I proposed a fix in #129 but this needs some input because I am looking at the Gherkin code base for the first time and don't know its intended architecture.

kwoxer:
Thanks. Indeed this is a version Issue. For me it worked to have this version setup active:
    "behat/behat": "3.3.0",
    "behat/gherkin": "4.4.5",
    "drupal/drupal-extension": "3.3.0",
    "behat/symfony2-extension": "2.1.1"

phil-davis:
Is this fixed by #166 release 4.6.2 ?

PR #129, opened 2017-09-14
Correctly filter features, now that the base path is correctly set

pfrenssen:
### Update + quick fix

This is a long standing issue which is affecting many people that are using Behat and want to run individual tests, either using the CLI or using PHPStorm. The issue is caused by a confusion between relative and absolute paths.

If you are affected by this you can solve it by adding the following lines to your `composer.json`:

```
"require-dev": {
    "behat/gherkin": "^4.6.0",
    "cweagans/composer-patches": "^1.6.7",
},
"patches": {
    "behat/gherkin": {
        "Correctly filter features, now that the base path is correctly set": "https://github.com/Behat/Gherkin/compare/v4.6.0...a7d84eb6ba245ea268a304312d14660788bf6c29.patch"
    }
}
```

### Original summary

This is a proposed fix for https://github.com/Behat/Behat/issues/1076 and https://github.com/Behat/Gherkin/issues/128.

In the latest version of Behat the base path is always set in Gherkin. There seems to have been some confusion recently whether `FeatureNode::getFile()` should return an absolute or a relative path. For example there recently was a fix committed (https://github.com/Behat/Gherkin/pull/126) that assumes it contains absolute paths.

`PathsFilter::isFeatureMatch()` thinks that `$feature->getFile()` returns an absolute path, and will always return FALSE if it actually contains a relative path such as `feature/homepage.feature`.

From the documentation of `FeatureNode::getFile()` it is not explicitly defined whether this is a relative or an absolute path, but it "feels" that it should be a relative path, otherwise this method would be named `FeatureNode::getPath()`.

This makes it again possible to execute a single feature in Behat.

Let's see if this breaks any tests and let's discuss whether or not `FeatureNode::getFile()` should return absolute or relative paths, document it properly, and enforce this when the file is being set, so this will no longer cause any confusion in the future.


pfrenssen:
OK so the tests assume that the feature paths are always absolute, will rework the PR with this premise.

pfrenssen:
So the latest fix is to make the feature filename always absolute, there were some other tests assuming that they were relative but I have updated them.

ciaranmcnulty:
Perhaps this could be encapsulated in a value object that is more explicit about the relative-ness?

pfrenssen:
That's a possibility too.

pfrenssen:
I am not very familiar with the architecture of Gherkin, but to me it seems that from an object oriented perspective the `NodeFeature` doesn't need to have knowledge about their absolute path. If this feature knows that its name is `feature/homepage.feature` then this should be enough, and it shouldn't care where exactly on the filesystem it is located.

jcornide:
Thanks @pfrenssen for working on this, It would be great to have it fixed because this makes impossible to use phpstorm with the last behat version

src/Behat/Gherkin/Node/FeatureNode.php
```diff
@@ -78,6 +78,10 @@ public function __construct(
         $file,
         $line
     ) {
+        // Verify that the feature file is an absolute path.
+        if (!empty($file) && (!is_string($file) || $file[0] !== DIRECTORY_SEPARATOR)) {
```
Taluu: On windows, the first char is not `DIRECTORY}_SEPARATOR`...

jpcaparas:
Hi guys, is there an update on this PR? 

Would really like to get `v3.4.0` integration working on PHP Storm.

everzet:
@jpcaparas I'm a bit behind due to reasons, but I'll look into fixing it this week. I sadly can't just merge this PR as it is heavily biased against Windows.

pfrenssen:
I don't know how to fix this, I have no Windows machine and I don't know how absolute paths look on it.

pfrenssen:
I had a look and the Symfony Filesystem component has a method to detect absolute paths: https://github.com/symfony/filesystem/blob/master/Filesystem.php#L606

src/Behat/Gherkin/Node/FeatureNode.php
webflo: How about using ``file_exists(realpath($file))``? Throw an exception if it fails. If not keep using the new generated absolute path.

fbourigault:
Any progress ?

pfrenssen:
@fbourigault I can progress with this when I have an answer on my question.

pfrenssen:
Actually let's go with the Filesystem component. The downside is that it adds another dependency, but then we don't need to worry about having to maintain this code if any of the OS'es change how they deal with the filesystem.

src/Behat/Gherkin/Node/FeatureNode.php
```diff
@@ -64,7 +66,7 @@ class FeatureNode implements KeywordNodeInterface, TaggedNodeInterface
      * @param ScenarioInterface[] $scenarios
      * @param string              $keyword
      * @param string              $language
-     * @param null|string         $file
+     * @param null|string         $file        The absolute path to the feature file.
```
deviantintegral: What reason do we have for this to be allowed as null? Since we've got a (minor) API break here, cleaning this up would be a nice improvement.

pfrenssen: Good suggestion, I will look into this. I know that a number of tests are calling this without a file being passed, but this might just be _because_ it is an optional parameter.

From the first look it seems like this should be required indeed. The line number is also required, and what point is it to have a line number if there is no file?

I'll have a look at the [most popular packages that require Gherkin](https://packagist.org/packages/behat/gherkin/dependents?page=1) and see if they have any cases where the class is instantiated without a file. And also, let's throw a deprecation warning for the moment, and actually drop it in a major update.

pfrenssen: I started working on this but there is actually a valid use case for not having a `$file`. You can call `Parser::parse($input)` to parse a string containing a Gherkin feature that is not saved in a file. This will end up calling `Parser::parseFeature()` which instantiates a `FeatureNode` without passing a `$file`. This also explains why the `$line` is required but the `$file` is not; even when passing a string directly, this will consist of multiple lines, so you can set a meaningful line number without having a file.

Ref. https://github.com/Behat/Gherkin/blob/master/src/Behat/Gherkin/Parser.php#L62

pfrenssen:
@warmwaterkruik you will have to give some more information on what you are trying to achieve and what is not working. Do you have code that passes in relative paths?

warmwaterkruik:
We usually run 1 feature file during development from the behat root directory where the behat.yml resides. De feature files are in different subdirectories from the behat root. From the behat root we run this command `bin/behat features/natarch-continuous/ArchiveLandingPage.feature` - With Behat 3.4 this wasn't possible anymore. With your changes added to relevant files this still is not possible. If I change the relative path to absolute path the file will still not run. E.g. : `bin/behat /var/www/natarch/project/behat/features/natarch-continuous/ArchiveLandingsPage.feature` This results in the error message: You must enter an absolute path.

pfrenssen:
I think you are mixing this up with https://github.com/Behat/Behat/issues/1076, which is about not being able to run individual tests in Behat.

This issue is about the Gherkin library mixing up relative and absolute paths. This is part of the problem why Behat is not working correctly, but let's discuss this in the correct issue.

DonCallisto:
Any progress with this?
I would like to try and test this with my project but 

```
behat/behat v3.4.3 requires behat/gherkin ^4.5.1 -> satisfiable by behat/gherkin[v4.5.1] but these conflict with your requirements or minimum-stability.
```

Of course my `repositories` is pointing to your fork but and I'm requiring `"behat/gherkin": "dev-filter-feature-path"` but can't manage to make it work due to this ... (I suppose?)

```
"dev-master": "4.4-dev"
```

Anyway, are there any feedback from maintainers? I'm not into `Gherkin` but fixes LGTM.

pfrenssen:
The PR is ready and all previous remarks were addressed. What still needs to be done is that it needs to be reviewed and committed.

JeffVeit:
This fixes Issue #1076 for me.

elkangaroo:
I can confirm this fixes Behat/Behat#1076 for us as well (using Behat + Symfony2Extension, running tests on console as well in PHPStorm).

pfrenssen:
@everzet do you have any time to take a look? This is affecting many people. Thanks!

src/Behat/Gherkin/Node/FeatureNode.php
```diff
@@ -78,6 +80,11 @@ public function __construct(
         $file,
         $line
     ) {
+        // Verify that the feature file is an absolute path.
+        $filesystem = new Filesystem();
+        if (!empty($file) && !$filesystem->isAbsolutePath($file)) {
+            throw new \InvalidArgumentException('The file should be an absolute path.');
```
fbourigault: When running from PHPStorm, I hit this exception having `features/api/bien/bienOption.feature` as `$file` value.

Here is my `behat.yml`:

```
default:
    suites:
        default:
            paths:
                - "%paths.base%/features"
            contexts:
                - PropelContext:
                - AppBundle\Utils\Behat\Context\HttpContext:
                    client: '@AppBundle\Utils\Behat\Client'
                - AppBundle\Utils\Behat\Context\ClockContext:
                    clock: '@AppBundle\Utils\Behat\MockClock'
                - ApiContext:
                    tokenStorage: '@security.token_storage'
                    userManager: '@app.user_manager'
                    clientManager: '@fos_oauth_server.client_manager.default'
                    oauth2Storage: '@fos_oauth_server.storage'
                    client: '@AppBundle\Utils\Behat\Client'
                - CoverageContext:
                - FixtureContext:
                    venteManager: '@app.vente_manager'
                    locationManager: '@app.location_manager'
                    emailManager: '@app.email_manager'
                    reportManager: '@app.report_manager'
                    userManager: '@app.user_manager'
                    clientManager: '@fos_oauth_server.client_manager'
                    mediaManager: '@app.media_manager'
                    oAuthStorage: '@app.oauth.storage'
                - PdfContext:
                    client: '@AppBundle\Utils\Behat\Client'
                - UploadContext:
                - ImageContext:
                    client: '@AppBundle\Utils\Behat\Client'
                    imagine: '@Imagine\Image\ImagineInterface'
                - PrevisiteContext:
                    mock: '@Noty\VisiteVirtuelle\Previsite\HttpClient\Plugin\Mock'
                - Noty\Vente\Behat\Context\VenteContext:
                    validator: '@validator'
                    propertyAccessor: '@property_accessor'
                - DiffusionContext:
                    normalizerFactory: '@app.passerelle.normalizer_factory'
                    passerelleManager: '@app.passerelle_manager'
    extensions:
        Caciobanu\Behat\DeprecationExtension:
        Behat\Symfony2Extension:
            kernel:
                env: test
```

the command line launched by PhpStorm:

```
/usr/local/Cellar/php/7.2.7/bin/php /private/var/folders/xy/sry34mvs6l7g7jfrjx64nh7m0000gn/T/ide-behat.php --format PhpStormBehatFormatter --no-interaction --config /Users/fabien/Projets/soqrate/noty/api/behat.yml /Users/fabien/Projets/soqrate/noty/api/features/api/bien/bienOption.feature:36
```

And the PhpStorm output:
```
Testing started at 14:47 ...
/usr/local/Cellar/php/7.2.7/bin/php /private/var/folders/xy/sry34mvs6l7g7jfrjx64nh7m0000gn/T/ide-behat.php --format PhpStormBehatFormatter --no-interaction --config /Users/fabien/Projets/soqrate/noty/api/behat.yml /Users/fabien/Projets/soqrate/noty/api/features/api/bien/bienOption.feature:36

In FeatureNode.php line 86:
                                        
  The file should be an absolute path.  
                                        

behat [-s|--suite SUITE] [-f|--format FORMAT] [-o|--out OUT] [--format-settings FORMAT-SETTINGS] [--init] [--lang LANG] [--name NAME] [--tags TAGS] [--role ROLE] [--story-syntax] [-d|--definitions DEFINITIONS] [--snippets-for [SNIPPETS-FOR]] [--snippets-type SNIPPETS-TYPE] [--append-snippets] [--no-snippets] [--strict] [--order ORDER] [--rerun] [--stop-on-failure] [--dry-run] [--] [<paths>]


Process finished with exit code 1
Empty test suite.
```

pfrenssen: Can you provide a backtrace so we can see which code is passing a relative path? Try running the test with the `-vvv` option.

jcornide:
Hi, any news on this?

src/Behat/Gherkin/Node/FeatureNode.php
idimopoulos: As an extra help for this case, when you apply this patch, you should also delete the tmp cache that behat creates. For me, applying the patch was also causing this exception to throw in PhpStorm.
Running `rm -rf /tmp/behat_*` solved it.

pfrenssen: Resolving this conversation, since no reply was given and it is probably just working as expected. The exception is thrown when a relative path is passed, which is wrong and no longer allowed thanks to this PR. The bug is probably in the calling code.

pfrenssen:
In the meantime this had diverged from master. I have solved the merge conflict and updated the tests.

AlexSkrypnyk:
@pfrenssen the patch no longer applies to 4.6.0 and your latest changes made many of our builds fail :( Could you please confirm that your patch works with `4.6.0` or I should be using a version of this patch from one of the previous commits?

c-man90:
I'm unable to install composer due to not being able to apply the patch.

AlexSkrypnyk:
@c-man90 The easiest way is to remove this patch altogether. It is not usable anymore until there is a stable release (unless you want to use `dev-master`, which is not a good thing).

c-man90:
I'm having trouble finding the source of where it's being applied.

pfrenssen:
This PR should be against master so that it can be merged.

@c-man90 @AlexSkrypnyk There are a few solutions you can use on composer level to get this PR working for you:

### Use master branch + latest version of PR

Note that I am using `cweagans/composer-patches` to apply patches.

```
"require-dev": {
    "behat/gherkin": "dev-master"
},
"patches": {
    "behat/gherkin": {
        "Correctly filter features, now that the base path is correctly set": "https://patch-diff.githubusercontent.com/raw/Behat/Gherkin/pull/129.patch"
    }
}
```

### Use latest stable + older version of PR

```
"require-dev": {
    "behat/gherkin": "^4.6.0"
},
"patches": {
    "behat/gherkin": {
        "Correctly filter features, now that the base path is correctly set": "https://github.com/Behat/Gherkin/compare/v4.6.0...a7d84eb6ba245ea268a304312d14660788bf6c29.patch"
    }
}
```

pfrenssen:
Of course it would be much easier if this was finally merged, maybe if someone can volunteer do to a review and approve the PR? It is only a handful of lines :)

AlexSkrypnyk:
@pfrenssen I appreciate the response, but do you mind re-reading my comment: the PR no longer applies to `4.6.0` for me, so I was wondering if it applies cleanly for you (your 2nd solution). 

Your 1st solution is not an option - `dev-*` branches are not stable and is a bad practice to use, so 1st solution does not work for me either.

Not sure how to get maintainers to merge it and make a new release.

Until the new version is released, I had to remove any references to this patch from all my projects :(

SunMar:
@AlexSkrypnyk the solution was already provided by @pfrenssen 

This PR does not apply to 4.6.0 because it has been rebased on master, and the master branch already has additional commits since the 4.6.0 release, so it only makes sense that patching the latests PR diff conflicts with 4.6.0.

If you want you apply it to the 4.6.0 version, use the second option where you apply the previous version of the PR that was not rebased on master and tell composer to apply it to 4.6.0.

AlexSkrypnyk:
@SunMar Thanks for pointing out. I completely missed that the patch string is now based on the commit hash - of course, this would apply cleanly to `4.6`. Thank you

And thank you @pfrenssen

danepowell:
@webflo is there anything blocking a review of this? https://github.com/Behat/Behat/issues/1076 has got to be one of the most popular Behat bugs and it would really great to get a new release of Gherkin with this fix.

func0der:
Thank you for your wok on this up until now.
To have even less impact on everything, wouldn't it be enough too change the initial problematic patch (https://github.com/Behat/Gherkin/pull/126/files) to not just use `realpath`(), so it searches in the *current working directory*, but to search in the current `%paths.base%`.
This would cause it to search the feature files in the same directory as the `behat.yml`.
This would at least fix all cases where the `feautures`-directory is a sub-folder of the `%paths.base%`.
Other cases would not be solved by this, would it?

This is just an idea. Maybe your solution already does have it all :)

danepowell:
I'm actually not sure who's maintaining Behat/Gherkin these days, would that be @stof @ciaranmcnulty  ?

Can one of ya'll please review this PR? We need to get this fixed in order to use Behat with Drupal 9.

ciaranmcnulty:
Only objection is the dependency. If the logic from Filesystem (https://github.com/symfony/filesystem/blob/master/Filesystem.php#L573-L579) can be copypasted in I'd merge this

ciaranmcnulty:
Looks good, thank you

pfrenssen:
Thanks so much for the review! Happy to see this merged :)

src/Behat/Gherkin/Node/FeatureNode.php
```diff
@@ -240,4 +244,28 @@ public function getLine()
...
+     *
+     * @param string $file A file path
+     *
+     * @return bool
+     *
+     * @see https://github.com/symfony/filesystem/blob/master/Filesystem.php
+     */
+    protected function isAbsolutePath($file)
+    {
+        if (null === $file) {
```
stof: This will never happen, as you only pass non empty values.

src/Behat/Gherkin/Node/FeatureNode.php
```diff
@@ -240,4 +244,28 @@ public function getLine()
...
+    /**
+     * Returns whether the file path is an absolute path.
+     *
+     * @param string $file A file path
+     *
+     * @return bool
+     *
+     * @see https://github.com/symfony/filesystem/blob/master/Filesystem.php
+     */
+    protected function isAbsolutePath($file)
```
stof: I would make it private to avoid any BC impact on it.

pfrenssen:
@stof thanks for the review, I opened a PR to address your remarks: #163

danepowell:
Thanks so much for merging this! Can't wait to test it. If it works, I think it will make a lot of people very happy.

JochenJung:
Hey, could it be this merge broke the ability to filter for features using the --name option?

We now get 'The file should be an absolute path' when calling:

`vendor/bin/behat -f progress --name "Some feature name"`

Works fine however when calling

`vendor/bin/behat -f progress features/suite_name/feature_name.feature`

JochenJung:
Okay.. just found the solution in another thread.

`rm -rf /tmp/behat_*` and --name works again :-)

PR #130, opened 2017-09-27
Allow `null` values in TableNode

sroze:
This PR allows tables to be constructed with `null` values within the cells.


stof:
this looks wrong to me. Parsing some Gherkin syntax would give you an empty string for empty cells, not `null`

sroze:
@stof I guess that's a good point, it might be that my usage is wrong and should force `null` values to be empty string when creating the `TableNode` objects. Wondering what is @everzet's PoV?

everzet:
Everything else in Gherkin never implicitly evaluates to null, but rather empty strings. For consistency sake I'd keep it the same in tables.

PR #138, opened 2018-03-22
[WIP] Add step to reproduce for maximum nested level error

yvoyer:
Hi, I just encountered the maximum nested level error when commenting a lot of lines in my .feature file.
I successfully reproduced it, but I'm wondering how we could resolve this error:

1. Increase the maximum nested level
2. Warn the user whith an exception when a sequence of X comment is found.
3. Issue will not be fixed 

**Option 1**: As mentioned in behat issues [727](https://github.com/Behat/Behat/issues/727) and [897](https://github.com/Behat/Behat/issues/897), the user may increase the maximum nested level in its (PHP.ini). 
But it can take some time and some research to understand that it was caused by the commenting of code.
**Option 2**: This option is probably not efficient, since the X could depend on the user's machine maximum nested level. And what would be the X value?
**Option 3**: Its just hard for a user to figure out why this happened.
**Other options**: ???

I would gladly fix this issue, but i need more input form you. thanks.

### DIFF
--- a/tests/Behat/Gherkin/ParserTest.php
+++ b/tests/Behat/Gherkin/ParserTest.php
@@ -144,4 +144,16 @@ FEATURE
             $feature->getLine()
         );
     }
+
+    public function testParsingManyCommentsShouldPass()
+    {
+        if (! extension_loaded('xdebug')) {
+            $this->markTestSkipped('xdebug extension must be enabled.');
+        }
+        $defaultPHPSetting = 256;
+        $this->iniSet('xdebug.max_nesting_level', $defaultPHPSetting);
+
+        $lineCount = 150; // 119 is the real threshold, higher just in case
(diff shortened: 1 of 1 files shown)


stof:
To me, the issue is that `Parser::parseComment` is not only responsible for parsing the comment. It is also responsible for parsing the expression following the comment and returning the node. This creates a big nesting when the next token is again a comment (which will do that loop again).

I think `Parser::parseExpression` should be refactored to use a loop over comments instead of using the recursive `Parser::parseComment` (which should then be removed):

```php
$type = $this->predictTokenType();

while ($type === 'Comment') {
    $this->expectTokenType('Comment');
    $type = $this->predictTokenType();
}

switch ($type) {
    // all the existing cases go here, except for 'Comment' which won't be possible anymore anyway.
}
```

PR #139, opened 2018-08-15
Fix "examples" keyword in feature file

antham:
Looking at [keywords](http://docs.cucumber.io/gherkin/reference/#keywords) documentation, its seems the russian word used for "examples" is "Примеры"


antham:
@everzet could you review this PR pls ?

stof:
using the same keyword in the test fixtures than the one supported by the normal usage of the library is indeed better. Thanks.

The tests are hardcoding the keywords instead of loading `i18n.php` to avoid making them fail when bringing in upstream translation updates (as that's not what's important for these tests), but syncing the test is a good thing.

PR #142, opened 2019-01-09
New translations 2019-01-9

### LINKED ISSUE #141: Support the Example keyword
as a synonym for Scenario

ciaranmcnulty:
Closes #141


ciaranmcnulty:
This is probably still a good idea, but maybe as an interim we can test that all of the files are _parseable_?

Issue #145, opened 2019-01-16
Unit tests have warning


phil-davis:
e.g. https://travis-ci.org/Behat/Gherkin/jobs/480397897

```
$ vendor/bin/phpunit -v --coverage-clover=coverage.clover
PHPUnit 5.7.27 by Sebastian Bergmann and contributors.
Runtime:       PHP 5.6.32 with Xdebug 2.5.5
Configuration: /home/travis/build/Behat/Gherkin/phpunit.xml.dist
................................................W..............  63 / 190 ( 33%)
............................................................... 126 / 190 ( 66%)
............................................................... 189 / 190 ( 99%)
.                                                               190 / 190 (100%)
Time: 20.44 seconds, Memory: 28.75MB
There was 1 warning:
1) Warning
The data provider specified for Tests\Behat\Gherkin\Keywords\CachedArrayKeywordsTest::testTranslation is invalid.
Outline should have examples table, but got none for outline "Erasing other agents' memory" on line: 35 in file: az_1.feature:
[...]

PR #146, opened 2019-01-16
Azerbijani scenario translation

phil-davis:
Issue #145 
1) make tests fail on warning so that stuff is noticed in CI
2) The ``az`` Azerbaijani translation of ``scenario`` translated to the identical word for ``examples``. That confuses the parser because when it sees ``Nümunələr`` it does not know if that keyword means ``examples`` or ``scenario``.

Fix the translation for ``scenario`` so it uses the singular ``Nümunə`` (Example)

That makes the unit tests happy, and also might help people writing scenarios an example tables in Azerbaijani.

### DIFF
--- a/i18n.php
+++ b/i18n.php
@@ -117,7 +117,7 @@ return array (
     'name' => 'Azerbaijani',
     'native' => 'Azərbaycanca',
     'rule' => 'Rule',
-    'scenario' => 'Nümunələr|Ssenari',
+    'scenario' => 'Nümunə|Ssenari',
     'scenario_outline' => 'Ssenarinin strukturu',
     'then' => 'O halda|*',
     'when' => 'Nə vaxt ki|Əgər|*',
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -6,6 +6,7 @@
          convertErrorsToExceptions="true"
          convertNoticesToExceptions="true"
          convertWarningsToExceptions="true"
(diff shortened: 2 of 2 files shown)


i18n.php
```diff
@@ -117,7 +117,7 @@
     'name' => 'Azerbaijani',
     'native' => 'Azərbaycanca',
     'rule' => 'Rule',
-    'scenario' => 'Nümunələr|Ssenari',
+    'scenario' => 'Nümunə|Ssenari',
```
ciaranmcnulty: This should not be edited directly - we get translations from cucumber upstream

phil-davis:
Example CI fail on warning with just the first commit: https://travis-ci.org/phil-davis/Gherkin/builds/480450868

i18n.php
phil-davis: Can you point me to where that is so I can make an upstream PR.

ciaranmcnulty: (short reply was because I was on mobile)

https://github.com/cucumber/cucumber/blob/master/gherkin/gherkin-languages.json

After it's merged you can run `i18n.php` in this repo and make a PR from the result 👍

phil-davis: See cucumber PR https://github.com/cucumber/cucumber/pull/559

phil-davis:
Closing here. I will make another PR to pull in the upstream change when that gets merged.

PR #147, opened 2019-01-16
Rewrite the i18n tests to parse in the test rather than the data provider

stof:
A failure in a data provider triggers a PHPUnit warning and skips all the tests. This does not allow to properly test for cases where some languages are triggering an error during parsing.
The test is now rewritten to perform the parsing in the test, allowing each language test to run separately.

Refs #145


phil-davis:
@stof PR #148 fixes the issue with the ``i18n.php`` AZ scenario translation.
After merging that, you will be able to rebase this and get CI passing.
(or cherry-pick the commit from #148 into this PR)

I made a demonstration PR #149 that shows that all is well when #148  and #147 are combined.

Issue #140, opened 2019-01-09
Support the Rule keyword


ciaranmcnulty:
This is new in Gherkin v6

https://docs.cucumber.io/gherkin/reference/#rule

stof:
If we want to support it, we should also consider refactoring Behat and Gherkin to separate the Gherkin AST from what they call pickles. This way, the Behat runner would not have to know about the extra grouping features of the Gherkin syntax.
This would be a BC break for hooks (some hooks would have to disappear, as the execution itself would not know about outlines or backgrounds anymore for instance).

In any case, adding that would require a new behat/gherkin major version, as adding rules inside features would break Behat when running them.

ciaranmcnulty:
Yes, if we went that route it might make sense to semi-generate the parser the same way that's done in other languages (and this repo might become a compat later)

Or cucumber-engine will become the future and we won't need a Gherkin parser for behat...

stof:
Well, what other languages are doing is not semi-generating the parser. They are calling the go parser (as a CLI) and then only rebuilding the output objects. I'm not sure that's what we want for now.

benztest:
I was wondering if there was any further updates regarding the Rule keyword?

As in my recent attempt to include two rules in a single feature results in an error. 
``` 
  [Behat\Gherkin\Exception\ParserException]
  Expected Step, but got text: "Rule: Authenticate as a learner" in file: C:\<path>\<file>.feature
```
.feature:
```
Feature: ...

Rule: Rule: Do no authenticate

Scenario: ...

Scenario: ...

Rule: Rule: Authenticate as a learner

Scenario: ...

Scenario: ...
```

ciaranmcnulty:
Uncommenting these lines will now produce a failing test for the Rule keyword implementation

https://github.com/Behat/Gherkin/blob/master/tests/Behat/Gherkin/Cucumber/CompatibilityTest.php#L29-L30

Harrisonbro:
Sorry if this should be obvious, but is there somewhere we can see how support for the Rule keyword moves from this issue into a version of Behat we can use as end users? Sorry, I’m not familiar enough with the project or how open source works to figure this out myself.

ciaranmcnulty:
This isn’t yet implemented here. Once it is, a new version of Behat will need to add support

kferrone:
Well that was a hard stop. So what's the status here. Years of being open doesn't look too promising. 

This looks hopeful: https://github.com/Behat/Behat/pull/1404
[...]

PR #150, opened 2019-03-09
Support Gherkin 6 rules (#140)

mdjward:
This change implements [Gherkin 6 rules](https://docs.cucumber.io/gherkin/reference/#rule) into the parser and resulting AST.

Changes to existing code (notably, `FeatureNode`) respects BC.  Tests/fixtures based on the example given on the linked Gherkin specification.


stof:
this would require a new major version. Behat would not be able to run a feature using rules (as it would not find its scenarios), without changes in the Behat runner. As such, it is a BC break.

PR #152, opened 2019-04-03
Test cucumber gherkin examples are parsable

ciaranmcnulty:
It'd be great to compare the AST we generate with the one the cucumber parser produces.

Currently we can't even parse all of the examples - this test includes a backlist and I can create issues for each.


stof:
If we add that many test fixtures, I suggest adding a `.gitattributes` file to exclude them from the downloaded archive.

phpunit.xml.dist
```diff
@@ -14,6 +14,7 @@
     <testsuites>
         <testsuite name="Behat Gherkin test suite">
             <directory>./tests/Behat/Gherkin/</directory>
+            <directory>./tests/Upstream/</directory>
```
stof: I suggest putting the upstream test in the `./tests/Behat/Gherkin/` folder with others (and later switching to PSR-4 instead of PSR-0 to remove these 2 levels of nesting)

stof:
currently, these tests don't have any assertion.

ciaranmcnulty:
@stof what would you suggest as an assertion? ATM I just want them not to throw

phpunit.xml.dist
ciaranmcnulty: Can do - they should probably have some sort of separation though (maybe `@group` is enough)

stof:
Well, probably checking whether we parsed them the right way...

ciaranmcnulty:
@stof ideally we'd compare them to the AST, but the cucumber AST serialisation is quite different :-/

phpunit.xml.dist
stof: but that's easy, as there is only one class on which the group needs to be applied.

ciaranmcnulty:
@stof Think I addressed everything except `.gitattributes` - should that be a separate PR?

tests/Behat/Gherkin/UpstreamGherkinTest.php
```diff
@@ -0,0 +1,94 @@
...
+        foreach (new \FilesystemIterator(__DIR__ . '/../../../gherkin_testdata/' . $type) as $file) {
+            if (in_array(preg_replace('/^.*\//', '', $file), $this->notSupported[$type])) {
+                continue;
+            }
+
+            if (!preg_match('/(?<stem>.*)[.]feature$/', $file, $matches)) {
+                continue;
+            }
+
+            if (in_array(preg_replace('/^.*\//', '', $matches['stem']), $this->notSupported[$type])) {
```
stof: this regex would be useless if the previous regex was applied on the filename only rather than the full path.

stof:
@ciaranmcnulty any plan to finish this PR ?

ciaranmcnulty:
I’ll get and tidy it up. I did start looking at a way to compare the cucumber AST and our YML etalons but that can be considered future scope

bin/update_testdata
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env php
+<?php
+
+chdir(__DIR__.'/../');
+shell_exec('rm -rf gherkin_testdata && svn export https://github.com/cucumber/cucumber/trunk/gherkin/testdata gherkin_testdata');
```
stof, line 5: Rather than copying the gherkin tests from upstream (and not knowing which exact version was copied), I suggest using an inline package repository in the composer.json to make composer install the `cucumber/cucumber` repository as a dev requirement.

See https://github.com/scssphp/scssphp/commit/d6a149e6bb3d54949e234a24d3bc1ed086679873#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34 for a similar approach in scssphp. There, we manage the exact reference we want to use (see https://github.com/scssphp/scssphp/pull/155 for a PR updating the version used)

ciaranmcnulty: Great technique, thankyou

stof:
@ciaranmcnulty why closing this ? this looked like a good idea.

ciaranmcnulty:
@stof i am revisiting it in a slightly different way. Will open a new PR when done

PR #163, opened 2020-02-27
Address review remarks from #129

pfrenssen:
Ref. https://github.com/Behat/Gherkin/pull/129#discussion_r385131437

### DIFF
--- a/src/Behat/Gherkin/Node/FeatureNode.php
+++ b/src/Behat/Gherkin/Node/FeatureNode.php
@@ -254,12 +254,8 @@ class FeatureNode implements KeywordNodeInterface, TaggedNodeInterface
      *
      * @see https://github.com/symfony/filesystem/blob/master/Filesystem.php
      */
-    protected function isAbsolutePath($file)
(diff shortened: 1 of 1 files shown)


ciaranmcnulty:
These are both valid review comments from @stof, I regret merging so fast since the last commits

Is making this private a BC break now? I guess the project doesn't have a stated BC policy

guilliamxavier:
@ciaranmcnulty: What about just releasing a v4.6.2 ASAP?

pfrenssen:
I don't think protected methods are usually considered to be part of a project's API, only the public methods are.

guilliamxavier:
👍 Moreover this method was only introduced in the patch-version 4.6.1, so I think there is no reason to hold this off

stof:
@pfrenssen protected methods are covered by backward compatibility, as changing them breaks child classes (and if you don't care about child classes, why using protected rather than private ?)

pfrenssen:
@stof OK I just opened this because it was your suggestion to change this to private. I don't care about this. Closing.

PR #167, opened 2020-06-06
Ignore files when exporting package

simonschaufi:
This commit is part of a campaign to reduce the amount of data transferred to save global bandwidth and reduce the amount of CO2. See https://github.com/Codeception/Codeception/pull/5527 for more info.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,7 @@
+/tests export-ignore
+/.gitattributes export-ignore
+/.gitignore export-ignore
+/.travis export-ignore
+/package.xml.tpl export-ignore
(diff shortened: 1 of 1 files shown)


simonschaufi:
@stof ping

.gitattributes
```diff
@@ -0,0 +1,7 @@
+/tests export-ignore
+/.gitattributes export-ignore
+/.gitignore export-ignore
+/.travis export-ignore
```
stof, line 4: should be `.travis.yml`, not `.travis`

stof (changes requested):
Sorry for the delay reviewing this. I missed it.

PR #172, opened 2020-07-09
Use new PHPUnit Framework TestCase and expectException in unit test code

phil-davis:
So that the unit test code is more up-to-date in preparation for phpunit6.

Note: On PHP 5.6 the Travis CI runs with PHPUnit 5.7.27 - so actually there is no CI any more that run with PHPUnit 4.5 (which is mentioned in `composer.json`. But I guess we leave that in `composer.json` because somebody might try to locally run the unit tests with some more ancient version of PHP.

Note 2: removed PHPUnit 4 and PHP 5.3 5.4 5.5 support.

### DIFF
--- a/tests/Behat/Gherkin/Cache/FileCacheTest.php
+++ b/tests/Behat/Gherkin/Cache/FileCacheTest.php
@@ -6,8 +6,9 @@ use Behat\Gherkin\Cache\FileCache;
 use Behat\Gherkin\Node\FeatureNode;
 use Behat\Gherkin\Node\ScenarioNode;
 use Behat\Gherkin\Gherkin;
+use PHPUnit\Framework\TestCase;
 
-class FileCacheTest extends \PHPUnit\Framework\TestCase
+class FileCacheTest extends TestCase
 {
     private $path;
     private $cache;
--- a/tests/Behat/Gherkin/Cache/FileCacheTest.php
+++ b/tests/Behat/Gherkin/Cache/FileCacheTest.php
@@ -7,7 +7,7 @@ use Behat\Gherkin\Node\FeatureNode;
 use Behat\Gherkin\Node\ScenarioNode;
 use Behat\Gherkin\Gherkin;
 
-class FileCacheTest extends \PHPUnit_Framework_TestCase
+class FileCacheTest extends \PHPUnit\Framework\TestCase
 {
     private $path;
     private $cache;
@@ -48,7 +48,7 @@ class FileCacheTest extends \PHPUnit_Framework_TestCase
 
     public function testBrokenCacheRead()
     {
-        $this->setExpectedException('Behat\Gherkin\Exception\CacheException');
+        $this->expectException('Behat\Gherkin\Exception\CacheException');
 
         touch($this->path . '/v' . Gherkin::VERSION . '/' . md5('broken_feature') . '.feature.cache');
         $this->cache->read('broken_feature');
@@ -56,7 +56,7 @@ class FileCacheTest extends \PHPUnit_Framework_TestCase
 
     public function testUnwriteableCacheDir()
     {
-        $this->setExpectedException('Behat\Gherkin\Exception\CacheException');
+        $this->expectException('Behat\Gherkin\Exception\CacheException');
 
         new FileCache('/dev/null/gherkin-test');
     }
--- a/composer.json
+++ b/composer.json
@@ -14,13 +14,13 @@
     ],
 
     "require": {
-        "php": ">=5.3.1"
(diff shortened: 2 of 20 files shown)


stof:
With composer.json saying that we support PHP 5.3+, not having these versions on CI is actually an issue.

phil-davis:
> With composer.json saying that we support PHP 5.3+, not having these versions on CI is actually an issue.

@stof What do you want to do about that?
Stop official support for those ancient PHP versions, or add back CI for PHP 5.3 5.4 5.5?

stof:
Well, given that the next behat/behat release will require PHP 7.2+, I think it might be fine to change things in this repo too (and then this PR is fine).

tests/Behat/Gherkin/Cache/FileCacheTest.php
```diff
@@ -48,15 +48,15 @@ public function testCacheAndRead()
...
+        $this->expectException('Behat\Gherkin\Exception\CacheException');
 
         touch($this->path . '/v' . Gherkin::VERSION . '/' . md5('broken_feature') . '.feature.cache');
         $this->cache->read('broken_feature');
     }
 
     public function testUnwriteableCacheDir()
     {
-        $this->setExpectedException('Behat\Gherkin\Exception\CacheException');
+        $this->expectException('Behat\Gherkin\Exception\CacheException');
```
stof, line 59: I think this method does not exist in 4.8

phil-davis: I adjusted `composer.json` to officially support only PHP 5.6 upwards and remove mention of PHPUnit 4.

phil-davis: And in some later PR we can easily move all the way up to just PHP 7.2+ and the much newer PHPUnit versions.

phil-davis:
@stof or anyone ping for review

Or maybe we move straight to dropping all the old PHP support and just support PHP 7.2+ to match what is happening anyway in Behat/Behat. If you want to do that, then let me know and I will make a PR and close this. That will actually be easy because there will be no need to think hard about which methods exist or not in which old phpunit major versions.

phil-davis:
Needs a decision about issue #176 - Drop older PHP versions. If that goes ahead now, then we can do a different PR with more changes and no need to worry about what PHPUnit stuff works in ancient versions.

phil-davis:
Rebased just now to get CI with master commits that were merged yesterday.
This PR is waiting on a decision about PHP version support.

tests/Behat/Gherkin/Cache/FileCacheTest.php
```diff
@@ -7,7 +7,7 @@
 use Behat\Gherkin\Node\ScenarioNode;
 use Behat\Gherkin\Gherkin;
 
-class FileCacheTest extends \PHPUnit_Framework_TestCase
+class FileCacheTest extends \PHPUnit\Framework\TestCase
```
stof, line 10: I suggest adding a use statement instead for consistency with the coding standards.

phil-davis: Done in latest commit.

phil-davis:
@stof it's up to you to decide if this is merged now, or wait for the PHP 7.2 decision and then directly do #178 or whatever.

ciaranmcnulty:
When I check out `master` and run the tests, I get the error:
```
Fatal error: Uncaught Error: Class 'PHPUnit\Framework\TestCase' not found
```

It seems related to this PR? Any ideas what's going on?

PHP7.4.13 and it's PHPUnit 5.2.7 that's installed

phil-davis:
In PHPUnit 5 the class is called `PHPUnit_Framework_TestCase`

The unit test code no longer supports that ancient PHPUnit version.

phil-davis:
https://github.com/sebastianbergmann/phpunit/releases/tag/5.7.26

5.7 is the latest PHPUnit5 - maybe the new class was added somewhere between 5.2 and 5.7

stof:
it was added in 4.8.35 and 5.7

ciaranmcnulty:
Perhaps we need a `—prefers-lowest` build

phil-davis:
In #178 we drop support for old PHP versions, and so phpunit5 will not be relevant any more anyway.

ciaranmcnulty:
Yeah but I’d like a stable release before that

phil-davis:
@ciaranmcnulty PR #184 should be enough. That will ensure that people trying to run the unit tests with an old PHP will at least get phpunit 5.7 installed, not lower.

PR #174, opened 2020-07-09
Allow PHPUnit 6 and 7 for testing and run CI with PHP 7.4

phil-davis:
PHP 5.6 uses PHPUnit 5

PHP 7.0 uses PHPunit 6

PHP 7.1 onwards uses PHPunit 7

and add a Travis CI matrix entry for PHP 7.4

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ cache:
   directories:
     - $HOME/.composer/cache/files
 
-php: [5.6, 7.0, 7.1, 7.2, 7.3]
+php: [5.6, 7.0, 7.1, 7.2, 7.3, 7.4]
 
 matrix:
   include:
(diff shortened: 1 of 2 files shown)


tests/Behat/Gherkin/ParserExceptionsTest.php
```diff
@@ -61,22 +61,19 @@ public function testTextInBackground()
...
-    Background: remove X to couse bug
-    Step is red form is not valid
-    asd
-    asd
-    as
-    da
-    sd
-    as
-    das
-    d
```
stof, line 73: this is changing what is tested (and might not prevent regressions anymore)

phil-davis: That was part of PR #173 - I have adjusted that PR now so that it does not touch the data (feature//background) to be tested.

This PR #174 will need to be rebased anyway after those other PRs are finalized and merged.

phil-davis:
@pamil @stof this is now an easy change to let the newer PHPUnit be used...

PR #177, opened 2020-07-14
Implement Gherkin parser in Parsica

mathiasverraes:


ciaranmcnulty:
Some context for this PR

@mathiasverraes and I are experimenting with Parsica as a parsing library for gherkin. This is to work out any kinks in Parsica with a real-word use-case and for me to learn how parser combinators work.

This could become a new parser for gherkin, it could become an alternative parser hidden behind a flag, it could become a replacement parser in a separate repo with a Behat extension, or it could be abandoned

composer.json
```diff
@@ -14,7 +14,8 @@
     ],
 
     "require": {
-        "php": ">=5.3.1"
+        "php": "^7.4",
```
stof, line 17: this is a blocker to me. The next minor version of Behat supports 7.2+, so we cannot be stricter in the Gherkin parser

tests/Behat/Gherkin/Parsica/EmptyFeatureTest.php
```diff
@@ -0,0 +1,168 @@
...
+
+namespace Behat\Gherkin\Parsica;
+
+use Behat\Gherkin\Node\FeatureNode;
+use Behat\Gherkin\Node\ScenarioNode;
+use Verraes\Parsica\Parser;
+use Verraes\Parsica\ParserFailure;
+use function Verraes\Parsica\string;
+
+require_once('Asserts.php');
```
stof, line 12: I'd rather configure `autoload-dev` in composer.json and use autoloading.

composer.json
mathiasverraes: As @ciaranmcnulty commented https://github.com/Behat/Gherkin/pull/177#issuecomment-658926614 this parser is very much experimental. And it will be a while before Parsica itself is stable enough to become the default parser for Behat. So I don't think we need to worry about versions now.

stof:
I suggest you to rebase your branch so that it contains the fixed Travis config which actually runs tests on PHP 7.4

stof:
@ciaranmcnulty you should really rebase so that Travis actually runs the tests on PHP 7.4 (and you might want to disable the other jobs running on older versions)

Issue #176, opened 2020-07-14
Drop older PHP versions


ciaranmcnulty:
Next Behat is going to drop PHP < 7.2

Should this parser too? It's used by codeception (which supports PHP 5.6) but they could use the older version still

(cc @stof @phil-davis)

ciaranmcnulty:
Ping @davertmik

ciaranmcnulty:
@pamil Just so you're in the loop

ciaranmcnulty:
My vote is bump it to 7.2 in a minor version

This library has been stable (/stagnant?) for a while and people on lower versions can continue to use the stable older minors.

phil-davis:
If just a minor version bump is done, then for example someone on PHP 5.6, 7.0 or 7.1 that has an explicit reference to `Behat/Gherkin` in their `composer.json`
```
    "require": {
        "behat/gherkin": "^4.0",
    }
```
Currently they will be getting 4.6 and all is well.

Then they will start getting 4.7 supporting only PHP 7.2+ and they will have to adjust their `composer.json` to pin to 4.6.*

But maybe there are not lots of "random" projects that directly require `Behat/Gherkin` and it is just `Behat/Behat` and a few others.

phil-davis:
I made PR #178 to demonstrate the unit test and CI changes to support PHP min version 7.2 and using the latest PHPUnit 8 and 9. IMO there is no run-time code that needs touching, which is "a good thing"

Taluu:
@phil-davis 
> Then they will start getting 4.7 supporting only PHP 7.2+ and they will have to adjust their composer.json to pin to 4.6.*

They won't. Comnposer won't install 4.7+ if they don't have 7.2. So they'll have nothing to do.

stof:
I vote for going PHP 7.2+ in our next minor version to match the behat/behat requirement

phil-davis:
We could also merge any other bugfix and feature PRs (e.g. #161 #164 #159 ) and make a release at that point.
Then users of the existing Behat/Behat release who are on PHP 5.6, 7.0, 7.1 will be able to get those things.

And after that, move to PHP 7.2 as min version.

phil-davis:
> and make a release at that point.

IMO a release could be done now of the current master. Then we can move forward again to drop old PHP, support newer PHP etc.

[...]

PR #161, opened 2019-12-13
Add support for multiple Example Tables with tags

dpakach:
### Description
Add support for multiple example table with tags on scenario outlines

This is a PR is addition to https://github.com/Behat/Gherkin/pull/119
Related Issue #117

Example:
```Gherkin
Scenario Outline: Test scenario outline
    When I do some thing
     And I use <data> from table
     Then I should get the <result>

     @smokeTest
     Examples:
      | data     | result |
      | simple   | data   |

     Examples:
      | data       |  result |
      | all        | kinds   |
      | of         | data    |
 ```

### Context
Having this allows us to use tags to run only small set of examples from the table on a big test suite.
for eg. in above example we can use `@smokeTest` tag filter to run only a basic example of the scenario. but If we want to run all the examples we can basically run it without any tags.

PR #178, opened 2020-07-16
Bump PHP to min version 7.2

phil-davis:
And associated bumps of phpunit etc. and refactoring of unit tests so they work with the new phpunit 8 and 9.

Fixes issue #176 (if this is agreed to go ahead)

https://phpunit.de/supported-versions.html

PHPUnit8 is for PHP 7.2+
PHPUnit9 is for PHP 7.3+


phil-davis:
IMO this is good code.

But PR #174 could be reviewed and merged first - that moves the PHPUnit versions up a bit and gets CI running on PHP 7.4 using the existing PHP 5.6 minimum.

ciaranmcnulty:
This looks good, can you add version 8 please?

phil-davis:
@ciaranmcnulty done - please review.

IMO CI will pass. We are at the time-of-day when Travis starts getting a backlog of open-source builds: https://www.traviscistatus.com/#day

The "Backlog Linux Builds for Open Source projects" graph does not look bad yet. But Travis has not started the jobs yet.

ciaranmcnulty:
I'll switch it to GitHub actions soon

ciaranmcnulty:
Maybe you can rebase from master and apply your changes to the GitHub build instead?

phil-davis:
CI just passed: https://travis-ci.org/github/Behat/Gherkin/builds/754479399

But now there is conflict because .travis.yml has been deleted.

I will try to rebase...

phil-davis:
Done. The CI is fast! @ciaranmcnulty please review

phil-davis:
Needs #184 sorted out first. Then I will rebase here...

phil-davis:
@ciaranmcnulty I have rebased this.

I guess you will make a `4.7.0` release first - that will support PHP 5.6 through 7.4 and have the new feature #161 "Add support for multiple Example Tables with tags". That will allow people on old PHP versions to still use the new feature.

Then a `4.8.0` could come soon after, with this PR merged it will advertise "official" PHP 8.0 support.

phil-davis:
@ciaranmcnulty is a release going to happen? And then merge this and release again?

phil-davis:
Needs to wait until issue #187 PR #189 is sorted out and 4.7.1 is released. Then I will rebase this and someone can decide when they want to move this forward...

phil-davis:
@ciaranmcnulty I have rebased and this is ready again, for whenever someone wants to do this.

But actually there is no rush. This PR adds PHP 8.0 to the test matrix in CI. But no actual changes to run-time code are needed. So we do know that the existing `master` run-time code does work with PHP 8.0. (It is just a bit harder if we want to routinely run the unit tests with all PHP from 5.6 through 8.0 in CI - the unit test code has to cope with a wide range of `phpunit` versions)

So maybe wait a while, in case other edge-cases or "stuff" needs to be fixed in the 4.7.* release series.

ciaranmcnulty:
Yeah it’s tested with 8.0 via behat. As you say it’s more critical if we are doing maintenance 

I do have a PR coming based on this though because PHP 5 syntax annoys me so much

phil-davis:
I personally have no problem with dropping the old PHP support. We are now always running our Behat test-runner with PHP 7.4, even if the back-end of the system-under-test happens to be running some other version of PHP (or some other programming language completely!).

I wonder how many people are actually still running their Behat test-runner scripts with "ancient" PHP?

ciaranmcnulty:
Behat has dropped older versions but the major other consumer is Codeception and they support older PHP.

To be nice to them, we'd probably need to fix major bugs on 4.7.x for a while, it would be helpful if they aided in that :)

(ping @Naktibalda)

Naktibalda:
I am afraid I've never written any Gherkin tests for Codeception.

We will release Codeception 5.0 in February which will require PHP 7.3, but Codeception 4 will receive some maintenance.

stof:
@ciaranmcnulty if there is no BC break in gherkin, I'd rather release it as 4.8.

ciaranmcnulty:
@Naktibalda Thanks, I don't feel so bad then! We can tag as 4.8 so Codeception 4 users will continue to use the older version and Codeception 5 get any newer features 👍

phil-davis:
Rebased to fix a recent conflict in `composer.json`

ciaranmcnulty:
So I guess we could merge this and call it `4.8.0`?

composer.json
```diff
@@ -14,23 +14,23 @@
     ],
 
     "require": {
-        "php": ">=5.6"
+        "php": ">=7.2"
```
ciaranmcnulty, line 17: I wonder if we should be more conservative here (i.e. `~7.2 | ~8.0`)

phil-davis: I am happy to do that, because actually we have no idea if this code will work with PHP 9.0 !
I think someone told me not to do that in a previous review :)

phil-davis: Done and rebased

ciaranmcnulty:
Thanks @phil-davis

phil-davis:
> So I guess we could merge this and call it `4.8.0`?

4.8.0 sounds good to me.

PR #180, opened 2020-07-21
Migrate to the new name for the travis jobs config

stof:
The matrix expansion of PHP version was not merged anymore with the explicitly included jobs (keeping only explicit ones). This tries to see whether this works fine when using the new name


stof:
and it does indeed fix the issue.

stof:
more importantly, this is the keyword which actually includes jobs coming from the `php` list. Their BC layer for the renaming of `matrix` to `jobs` looks broken.

PR #183, opened 2021-01-14
Translate travis config into github actions config

ciaranmcnulty:
Added PHP 8 while doing so


phil-davis:
The GitHub actions did not actually run for this PR. I guess they were somehow not enabled at the time. So that is how this managed to get merged but the PHP 8.0 test pipeline will fail.

ciaranmcnulty:
Yes this is an issue with GA - They don’t run until merged into master

phil-davis:
> Yes this is an issue with GA - They don’t run until merged into master

I guess that is a catch-22 - GitHub does not want random people making PRs to open-source repos that implement some huge CI without review and permission...

ciaranmcnulty:
Possibly I could have committed a ‘check out only’ version and then added the tests in a later PR

PR #184, opened 2021-01-15
Require at least phpunit 5.7

phil-davis:
Because the unit test code uses `PHPUnit\Framework\TestCase` which only became available from phpunit 5.7 onwards.

~And allow CI to use the newer phpunit 8 or 9 if it wants.~

~And do not test with PHP 8.0 yet. See test done in PR #185 - we are not quite ready for PHP 8.0 yet.~
That has already been done now.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,7 @@
     "require-dev": {
         "symfony/yaml": "~2.3|~3|~4",
         "symfony/phpunit-bridge": "~2.7|~3|~4",
-        "phpunit/phpunit": "~5|~6|~7"
+        "phpunit/phpunit": "~5.7|~6|~7"
     },
 
     "suggest": {


phil-davis:
Hmmm - the repo currently has a BitHub action that runs the unit tests with PHP 8.0. But `composer.json` does not mention a high-enough phpunit for that.

How did that get into `master` ?

I will push a fix.

phil-davis:
Hmmm again. We cannot use phpunit8 or phpunit9 because those require the test code to have `setUp(): void` declaration of the `setup()` return type.

But it looks like return type `void` is only available since PHP 7.1.

@ciaranmcnulty do we need to have the PHP 8.0 support at the same time as PHP 5.6 ?

ciaranmcnulty:
I don’t think we do. Let’s take it out

phil-davis:
@ciaranmcnulty ready for review. This gets green CI for the existing supported PHP 5.6 through 7.4

ciaranmcnulty:
Can you please rebase from master, I had to rejig the CI config

phil-davis:
> Can you please rebase from master, I had to rejig the CI config

@ciaranmcnulty done

PR #188, opened 2021-01-25
Add some variations of tags and comments in tests

phil-davis:
These combinations of comment lines and tag lines work OK. So maybe we can merge these extra test cases that pass.

1) I added some comments and extra examples tables to a Scenario Outline so that I demonstrates that kind of mix of stuff in a feature with tags.
2) Added a Scenario that has a comment, then tags, then the `Scenario` keyword, and put it after the complex Scenario Outline. That demonstrates that all is good when the comment and tag come in that order.
3) Added a Scenario that has a tag, then a comment. the the `Scenario` keyword, and it is not directly after a Scenario Outline. That case is also processed correctly.

I will make another PR that demonstrates the failing case (and maybe I can fix it?)

### DIFF
--- a/tests/Behat/Gherkin/Fixtures/etalons/tags_sample.yml
+++ b/tests/Behat/Gherkin/Fixtures/etalons/tags_sample.yml
@@ -59,5 +59,26 @@ feature:
         -
           tags:     [examples_tag3, examples_tag4]
           table:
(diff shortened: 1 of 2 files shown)


phil-davis:
Also see #189 where I believe that I have a fix.

PR #190, opened 2021-01-27
Test against Cucumber ASTs

ciaranmcnulty:
This is a replacement for #152 that checks we are parsing gherkin files the same way that Cucumber does. 

Tests that parse differently are marked as Not Implemented, with an indication of the issue. Hopefully we can work through them one by one.

The tests are pulled from Cucumber's test suite. We can of course add more (and maybe cucumber would appreciate edge cases upstream)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,8 @@
     "require-dev": {
         "symfony/yaml": "~2.3|~3|~4",
         "symfony/phpunit-bridge": "~2.7|~3|~4",
-        "phpunit/phpunit": "~5.7|~6|~7"
+        "phpunit/phpunit": "~5.7|~6|~7",
+        "cucumber/cucumber": "dev-gherkin-16.0.0"
     },
 
     "suggest": {
@@ -43,5 +44,20 @@
         "branch-alias": {
             "dev-master": "4.4-dev"
         }
-    }
+    },
+
+    "repositories": [
+        {
+            "type": "package",
+            "package": {
+                "name": "cucumber/cucumber",
+                "version": "dev-gherkin-16.0.0",
+                "source": {
+                    "type": "git",
+                    "url": "https://github.com/cucumber/cucumber.git",
+                    "reference": "00f128c8dc2970a564d8a706567836700a7039f0"
+                }
+            }
+        }
+    ]
 }
--- a/src/Behat/Gherkin/Loader/CucumberNDJsonAstLoader.php
+++ b/src/Behat/Gherkin/Loader/CucumberNDJsonAstLoader.php
@@ -1,0 +1,194 @@
+<?php
+
+namespace Behat\Gherkin\Loader;
+
+use Behat\Gherkin\Node\BackgroundNode;
+use Behat\Gherkin\Node\ExampleNode;
+use Behat\Gherkin\Node\ExampleTableNode;
+use Behat\Gherkin\Node\FeatureNode;
+use Behat\Gherkin\Node\OutlineNode;
+use Behat\Gherkin\Node\ScenarioInterface;
+use Behat\Gherkin\Node\ScenarioNode;
+use Behat\Gherkin\Node\StepNode;
+use Behat\Gherkin\Node\TableNode;
+
+/**
+ * Loads a feature from cucumber's protobuf JSON format
+ */
+class CucumberNDJsonAstLoader implements LoaderInterface
+{
+
+    public function supports($resource)
+    {
+        return is_string($resource);
+    }
+
+    public function load($resource)
+    {
+        $json = \json_decode(file_get_contents($resource), true);
+
+        $feature = self::getFeature($json, $resource);
+
+        return $feature ? array($feature) : array();
+    }
+
+    /**
+     * @return FeatureNode|null
+     */
+    private static function getFeature(array $json, string $filePath)
+    {
+        if (!isset($json['gherkinDocument']['feature'])) {
+            return null;
+        }
+
+        $featureJson = $json['gherkinDocument']['feature'];
+
+        $feature = new FeatureNode(
+            $featureJson['name'],
+            isset($featureJson['description']) ? trim($featureJson['description']) : null,
+            self::getTags($featureJson),
+            self::getBackground($featureJson),
+            self::getScenarios($featureJson),
+            $featureJson['keyword'],
+            $featureJson['language'],
+            preg_replace('/(?<=\\.feature).*$/', '', $filePath),
+            $featureJson['location']['line']
+        );
+
+        return $feature;
+    }
+
+    /**
+     * @return string[]
+     */
+    private static function getTags(array $json)
+    {
+        return array_map(
+            static function(array $tag) { return preg_replace('/^@/', '', $tag['name']); },
+            isset($json['tags']) ? $json['tags'] : []
+        );
+    }
+
+    /**
+     * @return ScenarioInterface[]
+     */
+    private static function getScenarios(array $json)
+    {
+
+        return array_values(
+            array_map(
+                static function ($child) {
+
+                    if (isset($child['scenario']['examples'])) {
+                        return new OutlineNode(
+                            $child['scenario']['name'],
+                            self::getTags($child['scenario']),
+                            self::getSteps(isset($child['scenario']['steps']) ? $child['scenario']['steps'] : []),
+                            self::getTables($child['scenario']['examples']),
+                            $child['scenario']['keyword'],
+                            $child['scenario']['location']['line']
+                        );
+                    }
+                    else {
+                        return new ScenarioNode(
+                            $child['scenario']['name'],
+                            self::getTags($child['scenario']),
+                            self::getSteps(isset($child['scenario']['steps']) ? $child['scenario']['steps'] : []),
+                            $child['scenario']['keyword'],
+                            $child['scenario']['location']['line']
+                        );
+                    }
+
+                },
+                array_filter(
+                    $json['children'] ? $json['children'] : [],
+                    static function ($child) {
+                        return isset($child['scenario']);
+                    }
+                )
+            )
+        );
+    }
+
+    /**
+     * @return BackgroundNode|null
+     */
+    private static function getBackground($json)
+    {
+        $backgrounds = array_values(
+            array_map(
+                static function ($child) {
+                    return new BackgroundNode(
+                        $child['background']['name'],
+                        self::getSteps(isset($child['background']['steps']) ? $child['background']['steps'] : []),
+                        $child['background']['keyword'],
+                        $child['background']['location']['line']
+                    );
+                },
+                array_filter(
+                    $json['children'] ? $json['children'] : [],
+                    static function ($child) {
+                        return isset($child['background']);
+                    }
+                )
+            )
+        );
+
+        return count($backgrounds) == 1 ? $backgrounds[0] : null;
+    }
+
+    /**
+     * @return StepNode[]
+     */
+    private static function getSteps(array $json)
+    {
+        return array_map(
+            static function(array $json) {
+                return new StepNode(
+                    trim($json['keyword']),
+                    $json['text'],
+                    [],
+                    $json['location']['line'],
+                    trim($json['keyword'])
+                );
+            },
+            $json
+        );
+    }
+
+    /**
+     * @return ExampleTableNode[]
+     */
+    private static function getTables(array $json)
+    {
+        return array_map(
+            static function($tableJson) {
+
+                $table = [];
+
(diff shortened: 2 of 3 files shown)


src/Behat/Gherkin/Loader/CucumberNDJsonAstLoader.php
```diff
@@ -0,0 +1,194 @@
+<?php
+
+namespace Behat\Gherkin\Loader;
```
ciaranmcnulty, line 3: This is in the main code rather than a test folder because

a. it follows where the Yaml loader is
b. it may be useful if we ever want to outsource the parser to the gherkin binary

ciaranmcnulty:
sigh, this worked fine on my machine :)

tests/Behat/Gherkin/Cucumber/CompatibilityTest.php
```diff
@@ -0,0 +1,129 @@
...
+    public function testFeaturesParseTheSameAsCucumber(\SplFileInfo $file)
+    {
+        if (isset((self::NOT_COMPATIBLE)[$file->getFilename()])){
+            $this->markTestIncomplete((self::NOT_COMPATIBLE)[$file->getFilename()]);
+        }
+
+        $gherkinFile = $file->getPathname();
+
+        $actual = $this->normaliseFeature($this->parser->parse(file_get_contents($gherkinFile), $gherkinFile));
+        $expected = $this->normaliseFeature($this->loader->load($gherkinFile . '.ast.ndjson')[0]);
```
stof, line 74: this `[0]` breaks for `empty.feature` which has no feature node.

src/Behat/Gherkin/Loader/CucumberNDJsonAstLoader.php
```diff
@@ -0,0 +1,194 @@
...
+                static function ($child) {
+                    return new BackgroundNode(
+                        $child['background']['name'],
+                        self::getSteps(isset($child['background']['steps']) ? $child['background']['steps'] : []),
+                        $child['background']['keyword'],
+                        $child['background']['location']['line']
+                    );
+                },
+                array_filter(
+                    $json['children'] ? $json['children'] : [],
```
stof, line 129: should be `isset($json['children']) ? $json['children'] : [],` to support cases where the key is not there (without first triggering a notice about non-existent key, which gets turned into an error on CI)

stof (changes requested):
@ciaranmcnulty could it be that your local setup ignores notices about undefined keys ?

composer.json
```diff
@@ -43,5 +44,20 @@
...
         }
-    }
+    },
+
+    "repositories": [
+        {
+            "type": "package",
+            "package": {
+                "name": "cucumber/cucumber",
+                "version": "dev-gherkin-16.0.0",
```
stof, line 54: for composer to work properly during a composer update, we need to ensure that update the reference for such a package repository is either updating the version of updating the `time` field for a dev version. Otherwise, composer will consider that your vendor folder is already uptodate (even though it does not have the same ref). Is this something corresponding to some tagged version ?

composer.json
```diff
@@ -43,5 +44,20 @@
...
+                "source": {
+                    "type": "git",
+                    "url": "https://github.com/cucumber/cucumber.git",
+                    "reference": "00f128c8dc2970a564d8a706567836700a7039f0"
```
stof, line 58: you might want to add a `dist` config as well, to support downloading the archive rather than cloning (better for caching). See https://github.com/scssphp/scssphp/blob/master/composer.json for an example.

stof:
should we add a separate test dealing with the files from the `bad` folder and expecting a gherkin exception ?

ciaranmcnulty:
@stof thanks for the feedback. Yeah, it seems like the image I was using has a different `error_reporting` - I can work through the rest.

src/Behat/Gherkin/Loader/CucumberNDJsonAstLoader.php
ciaranmcnulty: I can't wait for `??` to be available to use in this project :)

composer.json
ciaranmcnulty: This 'real' tag is `cucumber-gherkin/v16.0.0` but I couldn't seem to get that working

ciaranmcnulty:
```
Could not parse version constraint cucumber-gherkin/v16.0.0: Invalid version string "cucumber-gherkin/v16.0.0"
```

stof: that's indeed not a valid composer version

ciaranmcnulty: Not sure how to handle it when that's the actual tag, any hints? 

https://github.com/cucumber/cucumber/releases/tag/cucumber-gherkin%2Fv16.0.0

stof: Well, as that's an internal thing, we don't care if composer considers that as a dev version rather than a stable version for min stability. And as we reference a tag, we *will* update the `version` field when changing the reference (as we will replace that `16.0.0` part). So the current way is OK.

src/Behat/Gherkin/Loader/CucumberNDJsonAstLoader.php
```diff
@@ -0,0 +1,194 @@
...
+{
+
+    public function supports($resource)
+    {
+        return is_string($resource);
+    }
+
+    public function load($resource)
+    {
+        $json = \json_decode(file_get_contents($resource), true);
```
stof, line 28: this looks wrong to me. NDJson is about having one JSON object per line. you cannot json decode the whole file, otherwise it works only for single-line files.

stof (changes requested):

tests/Behat/Gherkin/Cucumber/CompatibilityTest.php
```diff
@@ -62,10 +62,11 @@ public function setUp()
     /**
      * @dataProvider cucumberFeatures
      */
-    public function testFeaturesParseTheSameAsCucumber(\SplFileInfo $file)
+    public function testFeaturesParseTheSameAsCucumber($file)
```
stof, line 65: why removing this typehint ?

tests/Behat/Gherkin/Cucumber/CompatibilityTest.php
```diff
@@ -82,11 +83,15 @@ public function testFeaturesParseTheSameAsCucumber(\SplFileInfo $file)
 
     public static function cucumberFeatures()
     {
-        foreach (new \FilesystemIterator(self::TESTDATA_PATH . '/good') as $file) {
-           if ($file->isFile() && $file->getExtension() === 'feature') {
-                yield $file->getFilename() => [$file];
+        $files = array();
```
stof, line 86: this is not an array of files but an array of test cases

stof: also, yield is perfectly fine on PHP 5.6 so you could revert this

ciaranmcnulty: The lowest version of phpunit we support doesn't allow a generator here

src/Behat/Gherkin/Loader/CucumberNDJsonAstLoader.php
ciaranmcnulty: That's true; but currently all the files are single-line! Maybe it's enough to add an assertion that there's only one line?

stof: but this file is named NDJsonLoader, not JsonLoader

tests/Behat/Gherkin/Cucumber/CompatibilityTest.php
stof: that's only a bug in 5.7.0, which was fixed in 5.7.1. Bump the min version to 5.7.1 for the 5.7 branch (or better to the latest 5.7.x patch release) and you're OK.

src/Behat/Gherkin/Loader/CucumberNDJsonAstLoader.php
stof: btw, supporting multiple lines would not be an issue: the method already reports an array, so parsing each (non-empty) line with `json_decode` and `self::getFeature` would do the work.

ciaranmcnulty: That's probably the best approach, thanks

I guess on the test side I can cast from` FeatureNode[]` to `?FeatureNode` (erroring if there's >1) ?

stof: Well, for now, yes, because our parser does not support multiple features per parsed file (btw, the description in the phpdoc says it returns a feature array, which is weird in this regard)

ciaranmcnulty: That's because it's implemented ultimately by things like DirectoryLoader

ciaranmcnulty: OK now using `file()`

ciaranmcnulty:
I think I addressed the feedback now

src/Behat/Gherkin/Loader/CucumberNDJsonAstLoader.php
```diff
@@ -0,0 +1,193 @@
...
+{
+
+    public function supports($resource)
+    {
+        return is_string($resource);
+    }
+
+    public function load($resource)
+    {
+        return array_map(
```
stof, line 26: you should either filter out `null` values or forbid `getFeature` from returning it

src/Behat/Gherkin/Loader/CucumberNDJsonAstLoader.php
```diff
@@ -0,0 +1,193 @@
...
+{
+
+    public function supports($resource)
+    {
+        return is_string($resource);
+    }
+
+    public function load($resource)
+    {
+        return array_values(array_map(
```
stof, line 26: isn't this missing a `array_filter` ?

src/Behat/Gherkin/Loader/CucumberNDJsonAstLoader.php
```diff
@@ -0,0 +1,193 @@
...
+                    }
+                )
+            )
+        );
+    }
+
+    /**
+     * @return BackgroundNode|null
+     */
+    private static function getBackground($json)
```
stof, line 115: missing `array` typehint

src/Behat/Gherkin/Loader/CucumberNDJsonAstLoader.php
ciaranmcnulty: Yep

ciaranmcnulty:
@stof changes addressed I think

src/Behat/Gherkin/Loader/CucumberNDJsonAstLoader.php
```diff
@@ -23,7 +23,7 @@ public function supports($resource)
 
     public function load($resource)
     {
-        return array_values(array_map(
+        return array_filter(array_map(
```
stof, line 26: I would actually still keep the `array_values` call so that we return a list rather than a array with holes.

stof:
@ciaranmcnulty I suggest creating issues for the various parity issues (if they don't exist yet)

ciaranmcnulty:
@stof yeah after it's merged I'll raise issues (so I can hyperlink to the exclusions in the issue descriptions)

tests/Behat/Gherkin/Cucumber/CompatibilityTest.php
```diff
@@ -0,0 +1,165 @@
...
+     * @var Parser
+     */
+    private $parser;
+
+    /**
+     * @var LoaderInterface
+     */
+    private $loader;
+
+    public function setUp()
```
stof, line 61: should be `protected` to match the visibility in PHPUnit

stof (commented):
I suggest cleaning the history of this PR before the merge by squashing commits which are about small review fixes (or partial reverts of changes done in a previous commit of the PR)

ciaranmcnulty:
@stof default in this repo is to squash and merge at the end, so I wasn't keeping the history too clean. I'm happy to tidy up if you think

stof:
well, if you do a squash at the merging time, then this already counts as keeping the history clean (I don't care about the history inside the PR view, but in the merged history). I was not aware of the fact you always squash at merging time.

ciaranmcnulty:
Only because someone set the repo to that being the default, at some point :)

ciaranmcnulty:
Thanks for all the review as always, @stof. It is much appreciated!

PR #192, opened 2021-02-01
Updated translations from cucumber

ciaranmcnulty:
there is something wrong with the newly imported `ne` translations, but I can't work out what it is

Opening this PR to illustrate the issue


stof:
The issue in the `ne` translation is that `अनी` is the translation for `and`, but is also an alternate translation for the `then` case. When trying to match it, one of them matches, which reports the wrong `keywordType` when the other one was expected.
`र` is probably affected because it is a translation for `and` (which inherits the previous keyword type, which is wrong due to being the one from matching `अनी`).

I would say that this is a shortcoming of our test infrastructure, where the auto-generated test does not take into account the fact that translations might not be unique.

ciaranmcnulty:
Looking upstream, it seems `ne` has been in the translations list forever but only just got moved before `nl` instead of after, so I suspect there's a bug in the import script

ciaranmcnulty:
Thanks @stof I've excluded the `ne` for now, maybe I will open an issue upstream

stof:
@ciaranmcnulty AFAICT, upstream will always work fine with duplicate translations. But they will always prefer one over the other to determine the step type (in the order `given`, `when`, `then`, `and`, `but`).
We do the same here but the difference is that we auto-generate some assertions which don't take that case into account.

stof:
> Looking upstream, it seems `ne` has been in the translations list forever but only just got moved before `nl` instead of after, so I suspect there's a bug in the import script

not true. It was added in October 2019 in https://github.com/cucumber/cucumber/commit/7d9327a69a5fb8363b8091f42e60782f835a128d#diff-77e95157e09332a4528c98c545b6a28972c83dee2e2643785c14c2da8e0db7c6 but our last sync is older than that...

ciaranmcnulty:
@stof I can't believe our last sync is older than than 🤦 

When #190 is merged we can source the data from the local copy of cucumber/cucumber, and I'll do a GitHub action that PRs an update when a new cucumber tag is released

PR #194, opened 2021-02-02
Update cucumber dependency automatically and use it for i18n

ciaranmcnulty:
Since #190 we have a local copy of cucumber, so we can use that for the i18n script. 

This also adds a GitHub Actions workflow that will create a PR when cucumber do a new gherkin release


.github/workflows/update.yml
```diff
@@ -0,0 +1,43 @@
+name: Update Cucumber
+
+on:
+  schedule:
+    -   cron: '0 7 * * *'
+  pull_request:
```
ciaranmcnulty, line 6: This is just to trigger on this PR to check it works in the main repo

composer.json
```diff
@@ -21,7 +21,7 @@
         "symfony/yaml": "~2.3|~3|~4",
         "symfony/phpunit-bridge": "~2.7|~3|~4",
         "phpunit/phpunit": "^5.7.1|~6|~7",
-        "cucumber/cucumber": "dev-gherkin-16.0.0"
+        "cucumber/cucumber": "dev-gherkin-15.0.1"
```
ciaranmcnulty, line 24: Setting it to an older version so that there is something for the workflow to do, for testing

ciaranmcnulty:
There is some permissions problem - it worked locally on my fork so I need to figure out why it's failing and whether it'll 'just work' when merged

PR #195, opened 2021-02-03
Automatic cucumber updates + use them for translations

ciaranmcnulty:
Since #190 we have a local copy of cucumber, so we can use that for the i18n script.

This also adds a GitHub Actions workflow that will create a PR when cucumber do a new gherkin release

Note this replaces #194 by pushing the branch to the main repo rather than my fork, to debug a permissions issue


ciaranmcnulty:
See #196 for a generated PR!

.github/workflows/update.yml
```diff
@@ -0,0 +1,43 @@
+name: Update Cucumber
+
+on:
+  schedule:
+    -   cron: '0 7 * * *'
+  pull_request:
```
ciaranmcnulty, line 6: Just for testing, will remove before merge

composer.json
```diff
@@ -21,7 +21,7 @@
         "symfony/yaml": "~2.3|~3|~4",
         "symfony/phpunit-bridge": "~2.7|~3|~4",
         "phpunit/phpunit": "^5.7.1|~6|~7",
-        "cucumber/cucumber": "dev-gherkin-16.0.0"
+        "cucumber/cucumber": "dev-gherkin-15.0.1"
```
ciaranmcnulty, line 24: This was to make sure the script did something, but I could actually commit it to master and wait a day, then we'd see a new PR the next morning if it's all working ok?

bin/update_i18n
```diff
@@ -60,8 +60,8 @@ file_put_contents(__DIR__.'/../i18n.php', <<<EOD
  * DO NOT TOUCH THIS FILE!
  *
  * This file is automatically generated by `bin/update_i18n`.
- * Actual Gherkin translations live in cucumber/gherkin repo:
- * {$gherkinUrl}
+ * Actual Gherkin translations live in the cucumber gherkin translation file:
```
stof, line 63: I would keep the existing message about the fact that the actual translations live in the gherkin repo. I would only replace the URL to point to the github view rather than the raw file. This will help when someone wants to contribute a change.

They will **not** send their translation changes to `__DIR__ . '/../vendor/cucumber/cucumber/gherkin/gherkin-languages.json'`

bin/update_cucumber
```diff
@@ -0,0 +1,53 @@
...
+if (PHP_VERSION_ID < 70400) {
+    echo "Updater requires PHP 7.4";
+    exit(1);
+}
+
+$composerFile = __DIR__ . '/../composer.json';
+$composerConfig = file_get_contents($composerFile);
+foreach (json_decode($composerConfig, true, 512, JSON_THROW_ON_ERROR)['repositories'] as $repository)
+{
+    if ($repository['package']['name'] == 'cucumber/cucumber') {
```
stof, line 13: I would first do a `if ($repository['type'] !== 'package') { continue; }` so that this script does not break if we add more repositories in the config with other types.

stof (changes requested):

bin/update_i18n
ciaranmcnulty: Good point, thanks

ciaranmcnulty:
For some reason this is opening PRs with i18n changes in, even though `master` is using 16.0.0 as far as I can tell

I'll debug on my fork rather than keep spamming PRs in this repo

ciaranmcnulty:
I've stabilised the sort order and now there's no diff against `master`, hence no PR

bin/update_i18n
```diff
@@ -34,7 +34,14 @@ foreach (json_decode($json, true) as $lang => $keywords) {
         }
 
         usort($words, function($type1, $type2) {
-            return mb_strlen($type2, 'utf8') - mb_strlen($type1, 'utf8');
+            $len1 = mb_strlen($type2, 'utf8');
+            $len2 = mb_strlen($type1, 'utf8');
+
+            if ($len1 == $len2) {
+                return $type1 > $type2 ? 1 : -0;
```
stof, line 41: Why `-0` ? Shouldn't it be `-1` ?

bin/update_i18n
```diff
@@ -34,7 +34,14 @@ foreach (json_decode($json, true) as $lang => $keywords) {
...
+            }
+
+            return $len1 - $len2;
```
stof, line 44: the whole comparison could be simplified by using the spaceship operator to return the expected `-1`, `0` or `1` value: `return [mb_strlen($type2, 'utf8'), $type1] <=> [mb_strlen($type1, 'utf8'), $type2];`
note that the `$type1` and `$type2` are swapped in the first array item compared to the second in this comparison. that's on purpose because we want a descending order on the length criteria but an ascending order on the secondary text criteria

bin/update_i18n
ciaranmcnulty: Yep

bin/update_i18n
ciaranmcnulty: That's better, I forgot this script was requiring 7.4 so it was available :+1:

bin/update_i18n
```diff
@@ -61,7 +68,7 @@ file_put_contents(__DIR__.'/../i18n.php', <<<EOD
  *
  * This file is automatically generated by `bin/update_i18n`.
  * Actual Gherkin translations live in cucumber/gherkin repo:
- * {$gherkinUrl}
+ * https://raw.githubusercontent.com/cucumber/cucumber/master/gherkin/gherkin-languages.json
```
stof, line 71: I suggest changing the URL to https://github.com/cucumber/cucumber/blob/master/gherkin/gherkin-languages.json instead, so that it links to the place where you can actually send the change (as that's what this comment is about).

bin/update_i18n
ciaranmcnulty: Yep this works much better thanks

ciaranmcnulty:
I've made the suggested changes, removed the 'on push' config, and added some conditionals so the translations update only runs if there was a new cucumber tag

So it's probably safe to merge and we'd find out if it works for real the next time cucumber tags a new gherkin

stof (approved):

ciaranmcnulty:
Thanks @stof!

PR #204, opened 2021-02-04
Refactor the parsing of comments to avoid recursion

stof:
Closes https://github.com/Behat/Gherkin/pull/138 (the first commit adding a test is the one from this PR)


stof:
I tested locally and this indeed makes the test pass with xdebug enabling the limit.

stof:
@ciaranmcnulty I don't know what is the state of #178 yet, but it might make sense to include that fix in a 4.7.3 release.

ciaranmcnulty:
Nice work @stof, thank you

PR #214, opened 2021-02-04
Support alternative docstrings format (```)

### LINKED ISSUE #155: Backtick docstrings not yet supported
Gherkin 6 supports backticks as a format for multiline strings:

```
   Given this is a thing:
      ```
      OMG
      ```
```

ciaranmcnulty:
Closes #155

### DIFF
--- a/src/Behat/Gherkin/Lexer.php
+++ b/src/Behat/Gherkin/Lexer.php
@@ -38,6 +38,7 @@ class Lexer
     private $featureStarted = false;
     private $allowMultilineArguments = false;
     private $allowSteps = false;
+    private $pyStringDelimiter = false;
 
     /**
      * Initializes lexer.
@@ -435,13 +436,24 @@ class Lexer
             return null;
         }
 
-        if (false === ($pos = mb_strpos($this->line, '"""', 0, 'utf8'))) {
+        if ((false === ($indent = mb_strpos($this->line, $delimiter = '"""', 0, 'utf8')))
+        && (false === ($indent = mb_strpos($this->line, $delimiter ='```', 0, 'utf8')))) {
             return null;
+        }
+
+        if ($this->inPyString) {
+            if ($this->pyStringDelimiter !== $delimiter) {
+                return null;
+            }
+            $this->pyStringDelimiter = false;
+        }
+        else {
+            $this->pyStringDelimiter= $delimiter;
         }
 
         $this->inPyString = !$this->inPyString;
         $token = $this->takeToken('PyStringOp');
-        $this->pyStringSwallow = $pos;
+        $this->pyStringSwallow = $indent;
 
         $this->consumeLine();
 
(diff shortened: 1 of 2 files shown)


ciaranmcnulty:
Should probably be accompanied with some local tests too

src/Behat/Gherkin/Lexer.php
```diff
@@ -38,6 +38,7 @@ class Lexer
     private $featureStarted = false;
     private $allowMultilineArguments = false;
     private $allowSteps = false;
+    private $pyStringDelimiter = false;
```
stof, line 41: please use `null` rather than `false`. A `string|false` type is a crappy API.

src/Behat/Gherkin/Lexer.php
```diff
@@ -435,13 +436,24 @@ protected function scanPyStringOp()
             return null;
         }
 
-        if (false === ($pos = mb_strpos($this->line, '"""', 0, 'utf8'))) {
+        if ((false === ($indent = mb_strpos($this->line, $delimiter = '"""', 0, 'utf8')))
+        && (false === ($indent = mb_strpos($this->line, $delimiter ='```', 0, 'utf8')))) {
```
stof, line 440: what if you have a ` ``` ` followed by `"""` later in the line ? I fear this would parse it wrong

stof (changes requested):

src/Behat/Gherkin/Lexer.php
ciaranmcnulty: I'm not sure what should happen in that case actually

Things like this are ok:
```
"""xml
<content/>
"""
```

I don't know if this should be:

```
""" this is some text ```
foo
"""
```

I guess I can see what the gherkin parser does

ciaranmcnulty: I guess there's an assumption that the only things before the `"""` should be whitespace, that isn't captured here? Maybe changing this to a regex is the cleanest approach?

ciaranmcnulty:
I can't particularly see a test to add aside from adding some fixtures (which would be duplicating the cucumber test)

stof:
if the cucumber tests cover that already, that's already enough to cover it IMO

src/Behat/Gherkin/Lexer.php
stof: maybe such case of `"""` not being the first non-whitespace content of the line should be contributed as additional tests in the cucumber tests.

ciaranmcnulty:
I definitely felt the need for more lower-level tests here. Perhaps the various Protected methods inside the Lexer could be promoted into objects and tested more thoroughly?

stof:
I don't think these methods should be promoted (I don't even think they should be protected btw). That would couple tests to the exact implementation of the parsing.
To me, this should be covered by feeding more valid and invalid inputs to the parser instead (which would then outlive any refactoring of the parsing implementation). And such inputs should ideally be contributed upstream rather than be kept in our repo (if they are not there already) so that official gherkin parsers also benefit from these tests (and we are sure that we have the same behavior).

ciaranmcnulty:
Cool. Contributing to the upstream tests is slightly harder than it could be, but the documentation has recently improved so I'll be adding some in

Issue #213, opened 2021-02-04
Whitespace is allowed in tags


ciaranmcnulty:
We currently allow `@foo bar baz` as a single tag, but cucumber does not

Removing this line will produce a failing test
https://github.com/Behat/Gherkin/blob/master/tests/Behat/Gherkin/Cucumber/CompatibilityTest.php#L48

PR #215, opened 2021-02-05
Tag parsing

### LINKED ISSUE #210: Comments after tags are captured
If we have something like:

```gherkin
@tag#comment
Scenario: It does something
```

Then the tag will be something like `tag#comment`

Removing this line will produce a failing test
https://github.com/Behat/Gherkin/blob/master/tests/Behat/Gherkin/Cucumber/CompatibilityTest.php#L42

ciaranmcnulty:
Closes #210 and #213 

Not too happy with the implementation yet

For the tag whitespace we could optionally show a warning rather than an exception for backwards compatibility

### DIFF
--- a/src/Behat/Gherkin/Lexer.php
+++ b/src/Behat/Gherkin/Lexer.php
@@ -519,8 +519,7 @@ class Lexer
         if(preg_match('/^(?<line>.*)\s+#.*$/', $line, $matches)) {
             ['line' => $line] = $matches;
             $this->consumeLineUntil(mb_strlen($line, 'utf-8'));
-        }
-        else {
+        } else {
             $this->consumeLine();
         }
 
--- a/src/Behat/Gherkin/Lexer.php
+++ b/src/Behat/Gherkin/Lexer.php
@@ -502,6 +502,7 @@ class Lexer
     protected function scanTags()
     {
         $line = $this->getTrimmedLine();
+        $line = preg_replace('/\s+#.*$/', '', $line);
         if (!isset($line[0]) || '@' !== $line[0]) {
             return null;
         }
--- a/src/Behat/Gherkin/Lexer.php
+++ b/src/Behat/Gherkin/Lexer.php
@@ -180,6 +180,15 @@ class Lexer
     }
 
     /**
+     * Consumes first part of line from input without incrementing the line number
+     */
+    protected function consumeLineUntil(int $trimmedOffset)
+    {
+        $this->line = mb_substr(ltrim($this->line), $trimmedOffset, null, 'utf-8');
+        $this->trimmedLine = null;
+    }
+
+    /**
      * Returns trimmed version of line.
      *
      * @return string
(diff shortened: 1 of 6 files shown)


src/Behat/Gherkin/Lexer.php
```diff
@@ -502,6 +502,7 @@ protected function scanTableRow()
     protected function scanTags()
     {
         $line = $this->getTrimmedLine();
+        $line = preg_replace('/\s+#.*$/', '', $line);
```
ciaranmcnulty, line 505: Just ignoring comments feels wrong in the Lexer, but we don't have a consumeLineUpTo() - maybe that's a better approach so it'd be followed by a Comment with the same line number?

src/Behat/Gherkin/Parser.php
```diff
@@ -635,6 +638,26 @@ protected function popTags()
...
+    /**
+     * Checks the tags fit the required format
+     *
+     * @param string[] $tags
+     */
+    protected function guardTags(array $tags)
+    {
+        foreach ($tags as $tag) {
+            if (preg_match('/\s/', $tag)) {
+                throw new ParserException(
```
ciaranmcnulty, line 650: This could be `E_USER_DEPRECATED` if we think whitespace tags are a valid current usage rather than an error

stof:
I would vote for adding such API to consume the line only partially instead of the hack about `#`

ciaranmcnulty:
@stof Yeah I will add something like `consumeLineUntil($char)` or similar

stof:
@ciaranmcnulty I suggest confirming upstream whether comments are meant to be allowed at the end of any line. If the answer is yes, it might be a matter of changing `consumeLine` instead.

ciaranmcnulty:
It looks like they're only allowed at the end of tags lines, added in https://github.com/cucumber/cucumber/pull/880

src/Behat/Gherkin/Parser.php
ciaranmcnulty: What do you think about this @stof?

src/Behat/Gherkin/Lexer.php
```diff
@@ -502,16 +511,24 @@ protected function scanTableRow()
...
+
         if (!isset($line[0]) || '@' !== $line[0]) {
             return null;
         }
 
+        if(preg_match('/^(?<line>.*)\s+#.*$/', $line, $matches)) {
+            ['line' => $line] = $matches;
+            $this->consumeLineUntil(mb_strlen($line, 'utf-8'));
+        }
+        else {
```
stof, line 523: why don't you follow PSR-12 here ?

src/Behat/Gherkin/Parser.php
stof: do tag expressions in TagFilter support tags with whitespace in them or no ? If such tags are not usable in TagFilter, then I would go for a ParserException

src/Behat/Gherkin/Lexer.php
ciaranmcnulty: I don't find it very valuable, but I'll change

ciaranmcnulty: Hm maybe that sounded too rude. It's PSR-12 I dislike, your reviews are invaluable

stof: I mean, the existing code already follows PSR-12 (or maybe the older PSR-2, we probably don't have code relying on new PHP features that were encoded in PSR-12). This is about being consistent.

ciaranmcnulty:
TagFilter is quite happy to take / split by tags with whitespace so there's a chance there are people out there depending on this functionality.

trigger_error with a deprecation warning?

stof:
@ciaranmcnulty then a deprecation is indeed better (and probably also in TagFilter if the filter uses a tag with whitespace)

ciaranmcnulty:
I think #220 is necessary to test the deprecations properly

ciaranmcnulty:
@stof would appreciate feedback on how the deprecation is implemented

funkyproject:
Hello , 

That trigger a deprecation notice when we use a filter likes "@tag1 && @tag2". 
Even if the tags haven't spaces.

PR #216, opened 2021-02-05
Cast table node exceptions into ParserExceptions when throwing

### LINKED ISSUE #212: Inconsistent cells count throws low-level error
If we have a table that isn't rectangular:

```
| foo |
| bar | baz |
```

we get a specific exception that is not a ParserException - it should probably be caught and a ParserException thrown

Removing this line will produce a failing test
https://github.com/Behat/Gherkin/blob/master/tests/Behat/Gherkin/Cucumber/CompatibilityTest.php#L46

ciaranmcnulty:
Rather than leaking this, we should make a ParserException

Closes #212

### DIFF
--- a/src/Behat/Gherkin/Parser.php
+++ b/src/Behat/Gherkin/Parser.php
@@ -566,11 +566,13 @@ class Parser
      */
     protected function parseTable()
     {
+        $table = $this->parseTableRows();
+
         try {
-            return new TableNode($this->parseTableRows());
+            return new TableNode($table);
         } catch(NodeException $e) {
             throw new ParserException(
-                $e->getMessage() . $this->file ? ' in file '.$this->file : '',
+                $e->getMessage() . ($this->file ? ' in file '.$this->file : ''),
                 0,
                 $e
             );
--- a/src/Behat/Gherkin/Parser.php
+++ b/src/Behat/Gherkin/Parser.php
@@ -550,13 +550,15 @@ class Parser
      */
     protected function parseExamples()
     {
-        $token = $this->expectTokenType('Examples');
-
-        $keyword = $token['keyword'];
-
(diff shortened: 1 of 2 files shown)


src/Behat/Gherkin/Parser.php
```diff
@@ -565,7 +566,16 @@ protected function parseExamples()
      */
     protected function parseTable()
     {
-        return new TableNode($this->parseTableRows());
+        try {
+            return new TableNode($this->parseTableRows());
+        }
+        catch(NodeException $e) {
```
stof, line 572: should be on the previous line to follow PSR-2/PSR-12

src/Behat/Gherkin/Parser.php
```diff
@@ -565,7 +566,16 @@ protected function parseExamples()
...
+            throw new ParserException(
+                $e->getMessage() . $this->file ? ' in file '.$this->file : '',
```
stof, line 574: you are missing braces around the ternary. The current code is executed as `($e->getMessage() . $this->file) ? ' in file '.$this->file : ''` which is not what you want.

src/Behat/Gherkin/Parser.php
```diff
@@ -565,7 +566,16 @@ protected function parseExamples()
      */
     protected function parseTable()
     {
-        return new TableNode($this->parseTableRows());
+        try {
+            return new TableNode($this->parseTableRows());
```
stof, line 570: I would probably keep the `parseTableRows` outside the `try/catch` so that we catch only the exceptions from `TableNode`.

stof (changes requested):
We probably need the same try/catch in `parseExamples` when instantiating the `ExampleTableNode`

stof (approved):

PR #219, opened 2021-02-09
Simplify the boolean condition for the tag matching

stof:
The new condition is totally equivalent in term of boolean logic, and easier to understand.


ciaranmcnulty:
Thanks @stof !

It really makes me wonder how it ended up that way

PR #226, opened 2021-02-17
Ignore the bin folder in archives

stof:
This folder contains maintenance scripts, not files meant to be part of the package.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,4 @@
+/bin/ export-ignore
 /tests export-ignore
 /.gitattributes export-ignore
 /.gitignore export-ignore


.gitattributes
```diff
@@ -1,3 +1,4 @@
+/bin/ export-ignore
```
ciaranmcnulty, line 1: There's not a trailing slash on `tests/` below, which is more correct?

stof: the trailing slash enforce matching only a directory. Without it, it could also match a file named `bin` at the root.

As we cannot have both a folder and a file with the same name at the root anyway (and that we match only at the root due to the leading slash), I would say that both are equally correct.

ciaranmcnulty (approved):

PR #229, opened 2021-03-09
Run behat fixture features against Cucumber parser

ciaranmcnulty:
To further document areas of difference with Cucumber, this uses the go binary releases to test that the AST produced by the real cucumber is the same as the AST behat produces


src/Behat/Gherkin/Loader/CucumberNDJsonAstLoader.php
```diff
@@ -118,7 +118,7 @@ private static function getBackground(array $json)
             array_map(
                 static function ($child) {
                     return new BackgroundNode(
-                        $child['background']['name'],
+                        isset($child['background']['name']) ? $child['background']['name'] : null,
```
stof, line 121: why not using `??`

tests/Behat/Gherkin/Cucumber/CompatibilityTest.php
```diff
@@ -86,13 +121,45 @@ public function testFeaturesParseTheSameAsCucumber(\SplFileInfo $file)
...
+     */
+    public function testBehatFeaturesParseTheSame(\SplFileInfo $ymlFile)
+    {
+        if (isset($this->behatFeaturesNotParsingCorrectly[$ymlFile->getFilename()])){
+            $this->markTestIncomplete($this->behatFeaturesNotParsingCorrectly[$ymlFile->getFilename()]);
+        }
+
+        if (isset($this->behatFeaturesCucumberCannotParseCorrectly[$ymlFile->getFilename()])){
+            $this->markTestIncomplete($this->behatFeaturesCucumberCannotParseCorrectly[$ymlFile->getFilename()]);
+            return;
```
stof, line 135: this `return` is dead code as `markTestIncomplete` never returns but always throws

bin/update_cucumber
```diff
@@ -55,6 +55,12 @@ file_put_contents($composerFile, $newJson);
 
 echo "Updated composer config:\n$newJson";
 
+$githubFile = __DIR__ . '/../.github/workflows/build.yml';
+$githubConfig = file_get_contents($githubFile);
+$newConfig = str_replace($oldTag, $newTag, $githubConfig);
+file_put_contents($githubFile, $newConfig);
+
+echo "Updated github build action:$newConfig\n";
```
stof, line 63: you should put a `\n` before the new config, to be more readable (otherwise, the first line of the config is stuck with the message)

src/Behat/Gherkin/Loader/CucumberNDJsonAstLoader.php
ciaranmcnulty: I just forgot what versions we support

stof: Well, you used `??` in the same diff a few lines below :smile:

PR #237, opened 2021-07-02
Fix DocBlocks (Boolean -> bool)

simonhammes:
This change helps static analyzers such as Psalm or PHPStan to understand the DocBlocks. 
Previously, Psalm tried to find a class called `Boolean`, which it obviously could not find.

Linked to https://github.com/Behat/Behat/issues/1307


stof (approved):

simonhammes:
@stof Thank you very much!

PR #244, opened 2021-10-01
Fix main branch alias version

mvorisek:


composer.json
```diff
@@ -41,7 +41,7 @@
 
     "extra": {
         "branch-alias": {
-            "dev-master": "4.4-dev"
+            "dev-master": "4.8-dev"
```
stof, line 44: I suggest using `4.x-dev`, which is less precise about which minor version will be the next one, but avoids forgetting to update it at each release.

mvorisek: updated, feel free to squash the changes before merge

mvorisek: updated

Issue #245, opened 2021-10-18
Migrate parser to `cucumber/common`?


ciaranmcnulty:
I'm going to do a spike at porting the cucumber parser over to PHP within the cucumber repository. This would contain a complete AST of value objects we could use.

For BC it should be possible to write a thin translation layer to map that to the existing Gherkin objects in this repo via a Loader, but a future version of Behat could rely on its AST directly, or even at some point rely on its 'pickles'.

Pros:

* Help from Cucumber team in maintenance, as internal structure would closely mirror the Java etc implementations
* Faster support for Cucumber language features
* Would use cucumber versioning so users could pin specific translations (see #193)

Cons:

* Relies on Cucumber team caring about PHP
* Introduces a layer of complexity in mapping

Any thoughts/feedback?

ciaranmcnulty:
Yeah it would basically be the same mapping as NdJsoNLoader does in the tests, except from cucumber objects instead of JSON

unkind:
It requires to implement parser generator for [gasparnagy/berp](https://github.com/gasparnagy/berp) first (template file).

ciaranmcnulty:
@unkind Just to update, I've started work on this and have a cucumber/messages implementation published on packagist

unkind:
How it's going? Need some help?

unkind:
Not sure about English language, but "Scenario" instead of "Example" in Russian sounds literally like "Script" for a movie or even a theatrical play. It's harder to represent to a newbies.

ciaranmcnulty:
See #253

Issue #153, opened 2019-04-03
Scenario Outlines inconsistencies


ciaranmcnulty:
Checking the parsing examples from Gherkin, the following fail due to issues with scenario outlines:

 * ~[descriptions.feature](https://github.com/cucumber/cucumber/blob/master/gherkin/testdata/good/descriptions.feature) - the description on the final Example causes a parse error~ [moved to #154]
 * [incomplete_scenario_outline](https://github.com/cucumber/gherkin/blob/main/testdata/good/incomplete_scenario_outline.feature) (and `readme_example`, `scenario_outlines_with_tags`, `several_examples`, `tags`) - scenario outlines without an Example are not allowed by our parser but are just ignored by cucumber
* [scenario_outline](https://github.com/cucumber/gherkin/blob/main/testdata/good/scenario_outline.feature) - Looks like Scenario keyword can be used for outlines now?

ciaranmcnulty:
Checked with @aslakhellesoy and in Gherkin 6 Scenario Outline and Scenario are synonyms - they only differ based on whether they are followed by an Example, so the following are valid:

```
Scenario Outline: Something
   Given I don't have any examples
```

```
Scenario: Something else
    Given <foo>

Examples:
    | foo |
    | bar |
```

stof:
This would require a BC break here, and in Behat though, as outlines are currently treated differently from scenarios in Behat.

Btw, I'm wondering how cucumber deals with the output formatting for these cases. Does it still tries to output something looking like a colored gherkin file ?

ciaranmcnulty:
I don't think it's necessarily a BC break - it allows some new syntaxes (above) that previously would have been a parse error

(our AST would need to still differentiate ScenarioOutline nodes from Scenario nodes, but we could detect which it is based on whether it is followed by Examples)

stof:
ah indeed, if we keep our 2 different nodes based on the presence of Examples, there is no BC break.

ciaranmcnulty:
@dpakach As you've been looking at how Examples and Outlines are treated, what do you think about implementing this?

aslakhellesoy:
Just my 2c - the official Gherkin parser does the same - the AST node that gets created depends on the absence/presence of examples. And we managed to release this without any users complaining about BC issues.

ciaranmcnulty:
Removing these lines will now produce failing tests 

[...]

PR #246, opened 2021-11-05
Add Rule keyword support

unkind:
Hi,

PR provides the `Rule` keyword support (https://cucumber.io/docs/gherkin/reference/#rule).


ciaranmcnulty:
I really want this feature, looking forward to reviewing it

unkind:
This one still doesn't work: `'complex_background.feature' => 'Rule keyword not supported',`.

It turns out `Scenario` and `Scenario Outline` have the same node type in Cucumber:

```gherkin
Feature: Complex background
...

  Rule: My Rule
    Scenario: with examples
      Given the <value> minimalism

      Examples:
      | value |
      | 1     |
      | 2     |
```

src/Behat/Gherkin/Keywords/KeywordsInterface.php
```diff
@@ -100,4 +100,12 @@ public function getButKeywords();
...
      */
     public function getStepKeywords();
+
+
+    /**
+     * Returns all rule keywords (splitted by "|").
+     *
+     * @return string
+     */
+    public function getRuleKeywords();
```
ciaranmcnulty, line 110: [BC] We should define a new interface that extends KeywordsInterface and adds this method

src/Behat/Gherkin/Node/FeatureNode.php
```diff
@@ -72,6 +77,7 @@ public function __construct(
         $description,
         array $tags,
         BackgroundNode $background = null,
+        array $rules,
```
ciaranmcnulty, line 80: [BC] we should add this as the last parameter and make it optional (with a default)

ciaranmcnulty, line 80: I have a slight concern that we lose the ordering of scenarios+rules here by having them as separate arguments. (The main concern is the impact on Behat's formatters).

Perhaps we should be passing an array of Scenarios+Rules? Needs some thought

src/Behat/Gherkin/Node/FeatureNode.php
```diff
@@ -185,6 +192,14 @@ public function getBackground()
         return $this->background;
     }
 
+    /**
+     * @return RuleNode[]
+     */
+    public function getRules(): array
```
ciaranmcnulty, line 198: [BC] We need to consider how a consumer will know this method is available. Maybe we need an interface to hint it? I slightly prefer

```php
if ($featureNode instanceof FeatureRulesNode) {
```

vs

```php
if(method_exists($featureNode, 'getRules')){
```

tests/Behat/Gherkin/Keywords/KeywordsTest.php
```diff
@@ -100,6 +100,7 @@ public function translationTestDataProvider()
                     ,
                     array(),
                     $background,
+                    [], // TODO: rules
```
ciaranmcnulty, line 103: I guess this should be resolved before merge

ciaranmcnulty (commented):
It's really good that you've used the cucumber tests to guide things - your comment about the Scenario Outline difference is captured in #153 

I've left a few comments, the ones marked [BC] would allow us to do this as a minor release, I'm not sure if that's the right approach or not yet

ciaranmcnulty:
@pamil @stof we need to consider how Behat will adapt to this change. It'll impact tagging, running and a lot more.

  1. We could do it in a minor version of behat/gherkin with BC layers, then Behat can do some conditional logic to see if the parser supports rules
     * Quite a 'safe' option
     * Could enable a situation where a user has a Rule-aware Gherkin but not a Rule-aware Behat and therefore rules aren't a parse error, but are ignored by the runner
     
  2. We could do it in a major, expand Behat's supported versions, and have similar conditional logic
     * Guarantees Behat understands rules if they're parsed
     * Probably messier conditional logic in Behat if we didn't put in a BC layer
  
  3. We could do it in a major and bump Behat's dependency overnight
     * Guarantees Behat understands rules if they're parsed
     * No BC issues
     * Could conflict on codebases that use e.g. Codeception+Behat (does that happen?)

src/Behat/Gherkin/Node/FeatureNode.php
stof: Well, if a consumer does not want to have to check for existence, the easiest way is to bump their min version of behat/gherkin.

To me, there is no need for an interface (which would not be allowed to be removed without a BC break, and there is no right way to trigger deprecations for instanceof checks relying on deprecated interfaces, only static analysis can do it)

unkind:
I found inconsistency with https://github.com/cucumber/common:

```gherkin
Feature: Some rules

  Background:
    Given fb

  Rule: A
    The rule A description

    Background:
      Given ab

    Example: Example A
      Given a
```

That parser doesn't remove indentation of description:

```json
"description":"    The rule A description"
```

unkind:
#245 is just better way for this.

Issue #250, opened 2022-04-26
TagFilter not checking for empty string


magikid:
When I run `behat --tags=""` then I expect no exceptions to be thrown.  In the version of behat that I'm running (3.8.1), it throws the following exception:

```
$ behat --tags="" --list-scenarios

In TagFilter.php line 146:
                                         
  Warning: Uninitialized string offset 0  
                                          
```

It looks like `TagFilter` doesn't check for the empty string before exploding the string and attempting to use array access on it.

PR #251, opened 2022-04-26
Check for empty Tag filter string

magikid:
If the `filterString` in `TagFilter` is an empty string, an exception
was being thrown because the code was expecting a non-empty string.
This checks if the string is non-empty and returns early if it is.

Issue: #250


src/Behat/Gherkin/Filter/TagFilter.php
```diff
@@ -137,6 +138,10 @@ protected function isTagsMatchCondition($tags)
     {
         $satisfies = true;
 
+        if (mb_strlen($this->filterString) === 0) {
```
stof, line 141: no need to use `mb_strlen`. Using `\strlen(...)` will be faster (PHP stores the length of strings so this would be a `O(1)` operation while `mb_strlen` is `O(n)`).

magikid: Done

stof (approved):

PR #253, opened 2022-07-10
Compatibility layer for `cucumber/gherkin`

ciaranmcnulty:
This adds a compatibility layer that can be used when `cucumber/gherkin` is also installed, that uses that parser instead and then maps the result back to our 'regular' AST for Behat (and maybe Codeception?) to consume.

This means that when parsing edge cases we will conform much more closely with the cucumber core implementations. It also means newer syntax items will be ignored / handled more cleanly rather than triggering odd parsing errors

To avoid bumping the minimum PHP version we support, this does not depend directly on `cucumber/gherkin`; instead there is a static `CucumberGherkinLoader::isAvailable()` method that can be called to see if the CucumberGherkinLoader can be used instead of the regular one, e.g. 

```php
if (
  class_exists(CucumberGherkinLoader::class) // if supporting older behat/gherkin
  && CucumberGherkinLoader::isAvailable()
) {
   $parser = new CucumberGherkinLoader(...)
} 
else {
   $parser = new GherkinFileLoader(...)
}
```

An alternative option to this would be to merge this into cucumber/gherkin 5.0.0 and let that require cucumber/gherkin (and PHP 8.1), then the version detection can move to Behat - I'm not sure that's warranted yet

## Motivation

Moving to the cucumber parser entirely will reduce the maintenance effort on Behat, and unlock newer features we didn't support yet. 

That's a big refactor though, and this goes someway towards it by enabling Behat to introduce the cucumber parser partially as an experimental feature (behind a config flag) to detect feature files in the wild that are unparsable, or parsed in unexpected ways.

## Compatibility

 - Passes the behat/gherkin etalon tests (tests in PR) with one exception (see Known incompatibilities)
 - Passes the entire behat/behat test suite (manually tested by me)

## Mapping notes

The majority of the mapping is simple from one tree to another, aside from a few places where the Behat AST doesn't support a feature:

### Multi-line scenario/background/outline names

```Gherkin
Scenario: Something awesome
   Something amazing
```

Cucumber sees this as Scenario with `name: 'Something awesome'` and `description: 'Something amazing'`. For compatibility we map this as a multi-line scenario name (which cucumber does not support but Behat does) so `title: "Something awesome\nSomething amazing"`

### Descriptions trimming

Cucumber preserves whitespaces in descriptions, Behat right and left-trims them based on how indented the previous keyword is. This logic is retained to be as like Behat as possible.

### Rule support

Cucumber supports scenarios inside tagged Rules:

```Gherkin
Feature:

    @foo
    Scenario:

    @bar
    Rule:

        @baz
        Scenario:
```

Because the Behat AST doesn't support Rule, this will be flattened into the same AST as this feature:

```Gherkin
Feature:

    @foo
    Scenario:


   

        @bar @baz
        Scenario:
```

This will enable implementations to parse files with Rule, and to use tags from Rule keywords when filtering examples

##  Known incompatibilities

There will doubtless be incompatibilities in the wild, but most will be around 'odd' Gherkin. The following are known:

### Comments at the end of Feature description no longer work

[Cucumber has a bug about this](https://github.com/cucumber/common/issues/1413) so one of the test files has had to be excluded.

### Backgrounds inside Rules do not work

There's no sensible way to map Background up into the Feature level and only have it apply to some scenarios, therefore we throw a parser error if it happens.

This syntax is discouraged by Cucumber anyhow

### DIFF
--- a/src/Behat/Gherkin/Loader/CucumberGherkinLoader.php
+++ b/src/Behat/Gherkin/Loader/CucumberGherkinLoader.php
@@ -2,6 +2,7 @@
 
 namespace Behat\Gherkin\Loader;
 
+use Behat\Gherkin\Cache\CacheInterface;
 use Behat\Gherkin\Cucumber\BackgroundNodeMapper;
 use Behat\Gherkin\Cucumber\ExampleTableNodeMapper;
 use Behat\Gherkin\Cucumber\FeatureNodeMapper;
@@ -26,6 +27,11 @@ final class CucumberGherkinLoader extends AbstractFileLoader
      * @var GherkinParser
      */
     private $parser;
+
+    /**
+     * @var ?CacheInterface
+     */
+    protected $cache;
 
     public function __construct()
     {
@@ -75,6 +81,16 @@ final class CucumberGherkinLoader extends AbstractFileLoader
     }
 
     /**
+     * Sets cache layer.
+     *
+     * @param CacheInterface $cache Cache layer
+     */
+    public function setCache(CacheInterface $cache)
+    {
+        $this->cache = $cache;
+    }
+
+    /**
      * Loads features from provided resource.
      *
      * @param string $path Resource to load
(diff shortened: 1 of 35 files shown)


ciaranmcnulty:
@Naktibalda any comments from a codeception POV?

src/Behat/Gherkin/Loader/CucumberGherkinLoader.php
```diff
@@ -27,6 +28,11 @@ final class CucumberGherkinLoader extends AbstractFileLoader
      */
     private $parser;
 
+    /**
+     * @var ?CacheInterface
+     */
+    protected $cache;
```
stof, line 34: should be private

test.feature
```diff
@@ -0,0 +1,7 @@
+Feature:
```
stof, line 1: This feature file at the root looks like a debugging file to me, that should not be committed.

ciaranmcnulty:
Another option for support would be to:

 1. Add the CucumberGherkinLoader with its `::isAvailable` method returning false, and make everything else a no-op
 2. Release the real implementation as behat/gherkin 5.0 that explicitly requires cucumber/gherkin and PHP 8.1
 
 Is that cleaner? Behat/Codeception could rely on `^4.9 || ^5.0` I think without breaking support for anyone?

ciaranmcnulty:
I want to redo this on top of #254 when that's merged

PR #259, opened 2024-08-20
Update CI to test against recent versions of PHP also

heiglandreas:


jrfnl (approved):

jrfnl:
Would be good to see this merged.

FYI: I've tested this PR in combination with PR #263 and the build would still pass, so all good.

PR #269, opened 2024-11-04
Update 4.10.0 changelog to highlight behaviour change

acoulton:
As this was not clear to end-users from the original release notes.

Fixes https://github.com/Behat/Behat/issues/1514


acoulton:
@nickvergessen is this clear enough? I will update the release notes on the github release as well when this merges.

carlos-granados (approved):
Looks good

PR #271, opened 2024-11-11
Improve title handling, especially for example scenarios 

### LINKED ISSUE #270: Scenario (outline) title and placeholders
In [this PR](https://github.com/Behat/Behat/pull/1459#discussion_r1831473219), we toyed with the idea of having placeholders in scenario outline titles.

## Why did this come up?

Currently, we distinguish between scenarios (aka "examples") of a scenario outline by appending a number to the title. This causes a problem when specific examples are rerun - if you rerun all examples except the first, the first example that is ran will be suffixed with `" #1"` even though it is actually the 2nd example in the feature file.
Conceptually it is actually the first, but in practice, that's not useful to users. See: https://github.com/Behat/Behat/issues/1448

## Options

1. keeping it as is :)
    ```text
    Passing number results in another number #1
    Passing number results in another number #2
    Passing number results in another number #3
    ```
2. appending the example values instead (was the main scope of that PR, btw):
    ```text
    Passing number results in another number (input: 2, output: 2)
    Passing number results in another number (input: 2.4, output: 2)
    Passing number results in another number (input: 2.5, output: 3)
    ```
3. replacing placeholders (assume the original title was `Passing <input> results in <output>`:
    ```text
    Passing 2 results in 2
    Passing 2.4 results in 2
    Passing 2.5 results in 3
    ```
    Concise, readable and simple. :)
4. like 3, but adding numbers to duplicate titles (see also question 3 below), with title `Results in <result>`:
    ```text
    Results in 2 #1
    Results in 2 #2
    Results in 3
    ```
    This would/may provide a backward-compatible upgrade path.

Apparently, option 3 is already [implemented in cucumber](https://github.com/cucumber/common/issues/2149), and has been [requested in the past](https://github.com/Behat/Behat/issues/1024).

## Open Questions (for Option 3)

1. Should we expose placeholder replacement in `ExampleNode`? As a behat user, I would have found that useful in the past, as a behat contributor, it would have been useful in https://github.com/Behat/Behat/pull/1459.
2. We have `getTitle` and `getOutlineTitle`. What would be the best way to implement placeholder replacement? My current opinion:
   - `getTitle` - this feels like it should return the definitive title, e.g. to be shown in reports and so on. In the ideal world, I think this should be the one replacing placeholders and adding numbers for deduplication.
   - `getOutlineTitle` - Internally (e.g. even in gherkin), it seems that there shouldn't be a concept of a scenario outline - instead they should be flattened into regular scenarios. With that in mind, I think the "outline title" should be the original (parent?) title - since that parent is the "scenario outline".
   - `getOutlineTitleWithPlaceholders` (additional method) - I would go for such a method if we want to be absolutely BC safe. That said, I do feel that the majority of end users would rather have the BC break than the current approach, but that's my hypothesis.

   It's important to note that `getTitle` is already used extensively and changing it might break things.
3. Not a question per se, but we're considering numbering repeated titles, this would preserve the past behaviour and avoid duplication given titles with missing placeholders.
   1. That does beg the question of how an ExampleNode would find that its title has been duplicated elsewhere.

----

@acoulton On point 2, this approach could work, but I'm not too fond of it:
```php
public function getTitle(?bool $raw = null): string {
    if ($raw === null) {
        @trigger_error('You must specify if you want the raw title or not, in the future this will change to FALSE', E_USER_DEPRECATED);
        $raw = true;
    }
    
    return $raw
        ? $this->title
        : $this->normaliseTitle($this->title); // normalise replaces placeholders and adds numbering
}
```

uuf6429:
Fixes #270

----

Remaining points
- [x] Cover changes with tests
- [x] What about interfaces? Changing them constitutes a BC break. (`getTitle` is defined in `KeywordNodeInterface` - there's also `ScenarioLikeInterface` and `ScenarioInterface`).
      We could add a new interface on top, though - no BC break there.
- [x] Still a bit uncomfortable requiring a migration from `getTitle()` to ~~`getScenarioTitle()`~~ `getName()`.
- [ ] Could it make sense to make `\Behat\Gherkin\Node\ExampleNode::replaceTextTokens` public? It feels less important to do so with this PR, plus it's a BC break (since `ExampleNode` is not final)

### DIFF
--- a/src/Behat/Gherkin/Node/ExampleNode.php
+++ b/src/Behat/Gherkin/Node/ExampleNode.php
@@ -97,12 +97,11 @@ class ExampleNode implements ScenarioInterface
      *
      * @return string
      *
-     * @deprecated You should normally not depend on the original row text, but if you really do, just switch to {@see self::getExampleText()}.
+     * @deprecated You should normally not depend on the original row text, but if you really do, please switch
+     *             to {@see self::getExampleText()} as this method will be removed in the next major version.
      */
     public function getTitle()
     {
-        @trigger_error('`ExampleNode::getTitle()` calls should be migrated to `ExampleNode::getScenarioTitle()` or `ExampleNode::getExampleText()`.', E_USER_DEPRECATED);
-
         return $this->text;
     }
 
@@ -189,11 +188,11 @@ class ExampleNode implements ScenarioInterface
     }
 
     /**
-     * Returns title for this scenario outline example.
+     * Returns the name of the scenario example.
      *
      * @return string
      */
-    public function getScenarioTitle()
+    public function getName()
     {
         return "{$this->replaceTextTokens($this->outlineTitle)} #{$this->index}";
     }
--- a/src/Behat/Gherkin/Node/ExampleNode.php
+++ b/src/Behat/Gherkin/Node/ExampleNode.php
@@ -197,11 +197,9 @@ class ExampleNode implements ScenarioInterface, NamedScenarioInterface
      *
      * You should normally not need this, since it is an implementation detail.
      * If you need the individual example values, use {@see self::getTokens()}.
-     * To get the fully-normalised/expanded title, use {@see self::getScenarioTitle()}.
-     *
-     * @return string
+     * To get the fully-normalised/expanded title, use {@see self::getName()}.
      */
-    public function getExampleText()
+    public function getExampleText(): string
     {
         return $this->text;
     }
--- a/src/Behat/Gherkin/Node/ExampleNode.php
+++ b/src/Behat/Gherkin/Node/ExampleNode.php
@@ -15,7 +15,7 @@ namespace Behat\Gherkin\Node;
  *
  * @author Konstantin Kudryashov <ever.zet@gmail.com>
  */
-class ExampleNode implements ScenarioInterface
+class ExampleNode implements ScenarioInterface, NamedScenarioInterface
 {
     /**
      * @var string
@@ -187,12 +187,7 @@ class ExampleNode implements ScenarioInterface
         return $this->outlineTitle;
     }
 
-    /**
-     * Returns the name of the scenario example.
-     *
-     * @return string
-     */
-    public function getName()
+    public function getName(): ?string
     {
         return "{$this->replaceTextTokens($this->outlineTitle)} #{$this->index}";
     }
--- a/src/Behat/Gherkin/Node/ExampleNode.php
+++ b/src/Behat/Gherkin/Node/ExampleNode.php
@@ -59,7 +59,7 @@ class ExampleNode implements ScenarioInterface, NamedScenarioInterface
      * @param string[]    $tokens
      * @param integer     $line         Line number within the feature file.
      * @param string|null $outlineTitle Original title of the scenario outline.
-     * @param null|int    $index        The 0-based index of the row/example within the scenario outline.
+     * @param null|int    $index        The 1-based index of the row/example within the scenario outline.
      */
     public function __construct($text, array $tags, $outlineSteps, array $tokens, $line, $outlineTitle = null, $index = null)
     {
--- a/src/Behat/Gherkin/Node/ExampleNode.php
+++ b/src/Behat/Gherkin/Node/ExampleNode.php
@@ -20,7 +20,7 @@ class ExampleNode implements ScenarioInterface
     /**
      * @var string
      */
-    private $title;
+    private $text;
     /**
      * @var string[]
      */
@@ -45,25 +45,31 @@ class ExampleNode implements ScenarioInterface
      * @var string
      */
     private $outlineTitle;
+    /**
+     * @var null|int
+     */
+    private $index;
 
     /**
      * Initializes outline.
      *
-     * @param string      $title
+     * @param string      $text         The entire row as a string, e.g. "| 1 | 2 | 3 |"
      * @param string[]    $tags
      * @param StepNode[]  $outlineSteps
      * @param string[]    $tokens
-     * @param integer     $line
-     * @param string|null $outlineTitle
+     * @param integer     $line         Line number within the feature file.
+     * @param string|null $outlineTitle Original title of the scenario outline.
+     * @param null|int    $index        The 0-based index of the row/example within the scenario outline.
      */
-    public function __construct($title, array $tags, $outlineSteps, array $tokens, $line, $outlineTitle = null)
+    public function __construct($text, array $tags, $outlineSteps, array $tokens, $line, $outlineTitle = null, $index = null)
     {
-        $this->title = $title;
+        $this->text = $text;
         $this->tags = $tags;
         $this->outlineSteps = $outlineSteps;
         $this->tokens = $tokens;
         $this->line = $line;
         $this->outlineTitle = $outlineTitle;
+        $this->index = $index;
     }
 
     /**
@@ -87,13 +93,17 @@ class ExampleNode implements ScenarioInterface
     }
 
     /**
-     * Returns example title.
+     * Returns the example row as a single string.
      *
      * @return string
+     *
+     * @deprecated You should normally not depend on the original row text, but if you really do, just switch to {@see self::getExampleText()}.
      */
     public function getTitle()
     {
-        return $this->title;
+        @trigger_error('`ExampleNode::getTitle()` calls should be migrated to `ExampleNode::getScenarioTitle()` or `ExampleNode::getExampleText()`.', E_USER_DEPRECATED);
+
+        return $this->text;
     }
 
     /**
@@ -176,6 +186,30 @@ class ExampleNode implements ScenarioInterface
     public function getOutlineTitle()
     {
         return $this->outlineTitle;
+    }
+
+    /**
+     * Returns title for this scenario outline example.
+     *
+     * @return string
+     */
+    public function getScenarioTitle()
+    {
+        return "{$this->replaceTextTokens($this->outlineTitle)} #{$this->index}";
+    }
+
+    /**
+     * Returns the example row as a single string.
+     *
+     * You should normally not need this, since it is an implementation detail.
+     * If you need the individual example values, use {@see self::getTokens()}.
+     * To get the fully-normalised/expanded title, use {@see self::getScenarioTitle()}.
+     *
+     * @return string
+     */
+    public function getExampleText()
+    {
+        return $this->text;
     }
 
     /**
--- a/src/Behat/Gherkin/Node/ScenarioNode.php
+++ b/src/Behat/Gherkin/Node/ScenarioNode.php
@@ -15,7 +15,7 @@ namespace Behat\Gherkin\Node;
  *
  * @author Konstantin Kudryashov <ever.zet@gmail.com>
  */
-class ScenarioNode implements ScenarioInterface
+class ScenarioNode implements ScenarioInterface, NamedScenarioInterface
 {
     /**
      * @var string
@@ -84,7 +84,7 @@ class ScenarioNode implements ScenarioInterface
      *
      * @return null|string
      */
-    public function getName()
+    public function getName(): ?string
     {
         return $this->title;
     }
--- a/src/Behat/Gherkin/Node/ScenarioNode.php
+++ b/src/Behat/Gherkin/Node/ScenarioNode.php
@@ -71,21 +71,20 @@ class ScenarioNode implements ScenarioInterface
      *
      * @return null|string
      *
-     * @deprecated You should use {@see self::getScenarioTitle()} instead.
+     * @deprecated You should use {@see self::getScenarioTitle()} instead as this method will be removed in the next
+     *             major version.
      */
     public function getTitle()
     {
-        @trigger_error('`ScenarioNode::getTitle()` calls should be migrated to `ScenarioNode::getScenarioTitle()`.', E_USER_DEPRECATED);
-
         return $this->title;
     }
 
     /**
-     * Returns scenario title.
+     * Returns the name of the scenario.
      *
      * @return null|string
      */
-    public function getScenarioTitle()
+    public function getName()
     {
         return $this->title;
     }
--- a/src/Behat/Gherkin/Node/ScenarioNode.php
+++ b/src/Behat/Gherkin/Node/ScenarioNode.php
@@ -70,8 +70,22 @@ class ScenarioNode implements ScenarioInterface
      * Returns scenario title.
      *
      * @return null|string
+     *
+     * @deprecated You should use {@see self::getScenarioTitle()} instead.
      */
     public function getTitle()
+    {
+        @trigger_error('`ScenarioNode::getTitle()` calls should be migrated to `ScenarioNode::getScenarioTitle()`.', E_USER_DEPRECATED);
+
+        return $this->title;
+    }
+
+    /**
+     * Returns scenario title.
+     *
+     * @return null|string
+     */
+    public function getScenarioTitle()
     {
         return $this->title;
     }
--- a/src/Behat/Gherkin/Node/ScenarioNode.php
+++ b/src/Behat/Gherkin/Node/ScenarioNode.php
@@ -71,7 +71,7 @@ class ScenarioNode implements ScenarioInterface, NamedScenarioInterface
      *
      * @return null|string
      *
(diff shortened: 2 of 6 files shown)


src/Behat/Gherkin/Node/ExampleNode.php
```diff
@@ -178,6 +188,30 @@ public function getOutlineTitle()
...
+    /**
+     * Returns the example row as a single string.
+     *
+     * You should normally not need this, since it is an implementation detail.
+     * If you need the individual example values, use {@see self::getTokens()}.
+     * To get the fully-normalised/expanded title, use {@see self::getScenarioTitle()}.
+     *
+     * @return string
+     */
+    public function getExampleText()
```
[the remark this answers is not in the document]
acoulton: There is still at least one valid usecase - for a formatter that wants to print the table row with the exact padding it needs to align with rows above and below.

That said I have done some more digging and it seems Behat currently does this two different ways:
* [PrettyExamplePrinter](https://github.com/Behat/Behat/blob/master/src/Behat/Behat/Output/Node/Printer/Pretty/PrettyExamplePrinter.php#L72) just calls `$example->getTitle()`.
* [PrettyExampleRowPrinter](https://github.com/Behat/Behat/blob/bfccc6549596d048e646754d5b1ef36fd71a3f13/src/Behat/Behat/Output/Node/Printer/Pretty/PrettyExampleRowPrinter.php#L83-L85) uses a more complex implementation that searches for the example in the `OutlineNode` to find the row number, then uses the deprecated `$outline->getExampleTable()` to get to the row content. That would be even more complex if rewritten not to use the deprecated method.

It took a while to understand what these classes are, but it turns out `PrettyExamplePrinter` is used when the `expand: true` option is set, while `PrettyExampleRowPrinter` is used if `expand = false`. 

The `PrettyExampleRowPrinter` has to use a more complex approach because it colours individual cells in the row depending on whether they are referenced in a step that failed (see below). So it needs the original padding for vertical alignment, but it also needs the individual `key => value` pairs. However it could potentially be simplified a bit, particularly with the new `index`/`exampleNumber` property.
image

However, I think that it would be a shame (and obviously a BC break) to force the `PrettyExamplePrinter` (or any similar third-party formatter) to implement complex logic like that in `PrettyExampleRowPrinter` just to get the cell padding that it can currently use from `getTitle()`.

src/Behat/Gherkin/Node/ExampleNode.php
```diff
@@ -87,13 +93,17 @@ public function getKeyword()
...
+     * Returns the example row as a single string.
      *
      * @return string
+     *
+     * @deprecated You should normally not depend on the original row text, but if you really do, just switch to {@see self::getExampleText()}.
      */
     public function getTitle()
     {
-        return $this->title;
+        @trigger_error('`ExampleNode::getTitle()` calls should be migrated to `ExampleNode::getScenarioTitle()` or `ExampleNode::getExampleText()`.', E_USER_DEPRECATED);
```
acoulton, line 104: *if* we stick with deprecating this, I wouldn't trigger a runtime deprecation for this, just mark it for IDE's to highlight.

acoulton (commented):
Sorry for the slow review, been a busy week.

Urgh, I hadn't appreciated how complex & nested the interfaces for these classes are. Considering that, I'd be nervous about adding yet another.

As I see it, the problem we're fighting is:

* Arguably, `getTitle()` has always returned the wrong thing for an ExampleNode - the idea implied by the interface is it looks and behaves like a Scenario, in which case the `title` should be the user-supplied title of the Scenario Outline, not the example values in text form.
* However, we can't change the result of `getTitle()` because there are existing valid uses - in Behat and probably elsewhere - that need the text example and rely on getting it from that method.
* And these are implementations of interfaces, not simple DTOs, so as you say, there's no easy way to provide a BC path to deprecate ambiguous method and add a new one.

So I think my suggestion for now is:
* Add the index to `ExampleNode` as planned.
* Add `getName()` to `ExampleNode` (instead of calling it `getScenarioTitle()`) - this would align with the naming in cucumber's AST.
* Add the `getExampleText()` as planned.
* Deprecate `getTitle()` *only* on `ExampleNode` but **only** with a docblock tag pointing to `getName()` or `getExampleText()`, not a runtime deprecation.
* Open an issue for a future major version to consider changing the interfaces so that anything "scenario-like" has a `getName()` instead of a `getTitle()`. I would probably do this alongside bumping Behat's major, so that we can be sure any extensions etc only get the Gherkin objects they were expecting even if they don't specify a `behat/gherkin` constraint of their own.

This means that:

* The change is BC for all existing users
* Only users who are writing formatters or similar and explicitly considering ExampleNode will see the deprecation (in their IDE or e.g. phpstan etc). But these are really the only cases at the moment where it's necessary to encourage people to be explicit about which one they want.
* If / when we release a new major renaming `getTitle()` to `getName()` then code that has already been updated for `ExampleNode` will be fine, anything that is only using the scenario interfaces will by default get the new (/expected) value instead of the table text, and we can highlight the difference in the upgrade notes.

This feels a bit messy, and I'm not sure it's right, but I think it probably is the least-worst compromise solution to adding this feature in a BC way.

Thoughts welcome!

src/Behat/Gherkin/Node/ExampleNode.php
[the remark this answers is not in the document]
uuf6429: Oh, wow.. quite unexpected.

OK then, I'll leave it as is.

src/Behat/Gherkin/Node/ExampleNode.php
uuf6429: That would make it harder for the end users to realise that (they would need to be actively looking at the usages). Are you still fine with that?

acoulton: Yeah, because most of the time I think it isn't / should not be a concern for an end user. 

If I am running behat with a third-party formatter, and my third-party formatter is calling a deprecated behat method, that shouldn't concern me as long as the formatter is still giving the output I expect. Assuming the formatter specifies its behat dependencies correctly, either I will eventually upgrade to a newer formatter release that solves it without me noticing, or I will be prevented from upgrading to a newer behat until the formatter is compatible.

That situation (and others like it) feel a lot more common than ones where the user themselves is calling the methods - and even if they are, it's not important that they do anything about it if they're happy with the titles their current code is displaying. Until / unless we remove it in a future BC break, but at that point they will find and fix it when they try to upgrade.

uuf6429: It shouldn't affect the end user, however (unless they're capturing all deprecations indiscriminately). I thought this was handled automatically in the symfony deprecations helper for example; that deprecations would be shown in the source project and usages by its immediate dependants (but not further).

Anyway I don't have a strong opinion - I'm also fine without this.

uuf6429:
> - Deprecate getTitle() only on ExampleNode but only with a docblock tag pointing to getName() or getExampleText(), not a runtime deprecation.
> - Open an issue for a future major version to consider changing the interfaces so that anything "scenario-like" has a getName() instead of a getTitle().

Why wait? If we do that later, we won't have a chance to deprecate it - it will suddenly be renamed from getTitle to getName (unless we have to wait for 2 major versions).

Additionally, I would consider creating a new `NamedNodeInterface` or similar that defines getName and apply it to the current two classes - since both classes would have a getName, even if someone extends those classes, nothing would break. Later (in the major version bumpp) we could merge `NamedNodeInterface` into some existing interface or whatever.

The point is that it would nice (for users) to code against `$scenarioOrExample->getName()` instead of a lot of conditions - after the major version bump that part would remain unchanged.

acoulton:
@uuf6429 my reluctance to introduce new interfaces/deprecate things at interface level at this stage is exactly because I think it needs more thought about what that structure should be.

ISTR there are nodes currently defining `getTitle` where this is probably correct.

I think part of the problem is methods have been defined at the wrong level of the hierarchy because they appear to be similar. For example `KeywordNode` - most lines in the gherkin file have a keyword and some following text, but it doesn't follow that the *meaning* of that text is the same for all types of node - as we're finding now with ExampleNode where the text is actually not the `title` at all.

You can see from the existing Behat formatter implementations that actually it very rarely makes sense to call `getTitle()` on a `KeywordNode` without knowing exactly what type of node it is. That probably means the method is in the wrong place, or called the wrong thing.

But to deprecate it on the interface now, we need to find a well-designed and BC way to replace it across all the nodes that implement it : even though some nodes might actually be fine with `getTitle`, just through a different (or no) interface.

Really for this usecase what we want to capture is that `Example` is just a special kind of `Scenario` (possibly the only one). And they both have a `name` which is some text with a specific meaning & purpose.

So if we had to introduce an interface now I would call it `NamedScenarioInterface` and apply it only to ExampleNode and ScenarioNode. 

I wouldn't use `NamedNodeInterface` because as with the existing interfaces, that implies nodes are similar because they happen to have a `getName()` method when we can see that isn't really true.

acoulton:
Oh and on this specific:

> Why wait? If we do that later, we won't have a chance to deprecate it - it will suddenly be renamed from getTitle to getName (unless we have to wait for 2 major versions). 

I don't see us going immediately from this to a new major. Indeed with current resources a new major could still be a way off, especially if we're going to look at switching parser/moving to Pickles etc.

And even when we release a major there will be a period where we are maintaining both version series.

So there will still be an opportunity to deprecate methods between now and then - I just think we should minimise that until we can give users a clear overall explanation of what will be changing.

Deprecating the method on ExampleNode where it is currently ambiguous feels fine to do right now as there is an identified problem with that specific object and we have an easy solution.

acoulton:
I *think* Outline is a higher level grouping of scenarios, like a feature or a suite (or a Rule if/when we support them).

It's not the same as a Scenario/Example which represents a specific set of executable steps.

So I would leave Outline alone, for the moment at least.

uuf6429:
@acoulton side-note: there is a `\Behat\Gherkin\Filter\NameFilter` class which filter by _title_. 😅

uuf6429:
@acoulton while adjusting tests, I found out that outline represents the "scenario outline" node and that, for example, iterating over scenarios in a feature file will actually yield instance of that object. Apparently it is up to caller to ignore iterate over the "example nodes" of that object while ignore the rest of the object.
In short, a naive `foreach ($scenarios as $scenario) echo $scenario->getName()` would not work. Because of that, I think I would implement `getName()`+`NamedScenarioInterface` on that class too.

uuf6429:
Hi @carlos-granados!

I guess @acoulton might be a bit occupied at the moment, so maybe you could give the task a look? A second (well, third) set of eyes wouldn't hurt. 😄

src/Behat/Gherkin/Node/ExampleNode.php
```diff
@@ -178,6 +187,25 @@ public function getOutlineTitle()
...
+    {
+        return "{$this->replaceTextTokens($this->outlineTitle)} #{$this->index}";
+    }
+
+    /**
+     * Returns the example row as a single string.
+     *
+     * You should normally not need this, since it is an implementation detail.
+     * If you need the individual example values, use {@see self::getTokens()}.
+     * To get the fully-normalised/expanded title, use {@see self::getScenarioTitle()}.
```
carlos-granados, line 200: `getScenarioTitle()` does not seem to exist

src/Behat/Gherkin/Node/ScenarioNode.php
```diff
@@ -70,12 +70,20 @@ public function getNodeType()
      * Returns scenario title.
      *
      * @return null|string
+     *
+     * @deprecated You should use {@see self::getScenarioTitle()} instead as this method will be removed in the next
```
carlos-granados, line 74: `getScenarioTitle()` does not exist here either

carlos-granados (changes requested):

carlos-granados:
Apart from my comments about a couple of comments that need to be updated, everything else looks OK to me

src/Behat/Gherkin/Node/ExampleNode.php
```diff
@@ -178,6 +187,25 @@ public function getOutlineTitle()
...
+     *
+     * @return string
+     */
+    public function getExampleText()
```
acoulton, line 204: I *think* we can strictly type this, since it's new & there is always text?

```suggestion
     */
    public function getExampleText(): string
```

tests/Behat/Gherkin/Node/OutlineNodeTest.php
```diff
@@ -122,11 +123,37 @@ public function testPopulatesExampleWithOutlineTitle()
...
+            12 => array('John', 'john@example.com'),
         ), 'Examples');
 
-        $outline = new OutlineNode('An outline title', array(), $steps, $table, null, null);
-
-        $this->assertCount(1, $examples = $outline->getExamples());
-        $this->assertEquals('An outline title', current($examples)->getOutlineTitle());
+        $outline = new OutlineNode('An outline title for <name>', array(), $steps, $table, null, null);
+
+        $this->assertEquals(
```
acoulton, line 131: I don't think it makes any difference here, but I always try to use `assertSame` unless there's a reason not to - just avoids any risk of false positives if someone adds another case (e.g. maybe with empty strings) in the future and doesn't notice the loose assertion.

```suggestion
        $this->assertSame(
```

acoulton (changes requested):
@uuf6429 I'm so sorry for the slow review - I hadn't originally realised you'd done more on it (it seems github doesn't send a notification when a draft PR is marked ready for review, only if you specifically request a review or add a comment). And then I was off sick and then trying to catch up.

This looks good to me with a couple of very small comments - @carlos-granados the `getScenarioTitle()` references in phpdocs are just leftover from a previous iteration, they can be updated to `getName()`

tests/Behat/Gherkin/Node/OutlineNodeTest.php
uuf6429: Agreed, that's also an old habit of mine - in the past (PHPUnit 5 or 4 maybe?) `assertSame` didn't show the level of detail (e.g. a diff) like `assertEquals` :)

acoulton: Ah yeah, I remember that...

acoulton (approved):
Great, thanks :)

src/Behat/Gherkin/Node/ExampleNode.php
```diff
@@ -45,25 +45,31 @@ class ExampleNode implements ScenarioInterface
...
-     * @param string      $title
+     * @param string      $text         The entire row as a string, e.g. "| 1 | 2 | 3 |"
      * @param string[]    $tags
      * @param StepNode[]  $outlineSteps
      * @param string[]    $tokens
-     * @param integer     $line
-     * @param string|null $outlineTitle
+     * @param integer     $line         Line number within the feature file.
+     * @param string|null $outlineTitle Original title of the scenario outline.
+     * @param null|int    $index        The 0-based index of the row/example within the scenario outline.
```
carlos-granados, line 62: It says that the index is 0-based but it looks to be 1-based

carlos-granados (changes requested):
Just found a tiny thing that needs changing

src/Behat/Gherkin/Node/ExampleNode.php
uuf6429: Ah, you're totally right.. the argument passed is `row + 1`..

carlos-granados (approved):
All good now, thanks @uuf6429

carlos-granados:
@acoulton now that his is finally merged, should we create a release so that this can be used on the Behat side?

acoulton:
@uuf6429 released now, thank you!

uuf6429:
First of all, thanks again for the effort.

One quick point, can we somehow make sure to squash PRs? It's just that I'm not typically working on branches with the understanding that the commits (and more importantly, their messages) will show up verbatim in the release branch.

acoulton:
@uuf6429 we can certainly try, the challenge is if there are unrelated changes - for example the fixes for the file cache tests on windows in this PR - in the same branch. I don't think they should be just squashed into a single merge commit for the overall feature, and there's no way to partially squash on GitHub.

I think it's probably better if PR authors squash and force-push as appropriate before the final code review (apart from simple changes / ones that can easily be squash-merged).

But you're right that we should probably be looking at the commits & messages as well as the diff when reviewing PRs.

PR #272, opened 2024-11-11
change: Drop support for PHP < 8.1, Symfony < 5.4 and Symfony 6.0 - 6.3

acoulton:
In line with the new behat version support policy adopted in Behat/Behat#1503.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -14,18 +14,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php: [7.2, 7.3, 7.4, 8.0, 8.1, 8.2, 8.3, 8.4]
+        php: [8.1, 8.2, 8.3, 8.4]
         composer-flags: [ "" ]
         symfony-version: [ "" ]
         include:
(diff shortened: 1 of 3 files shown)


carlos-granados (approved):
Looks good

.github/workflows/build.yml
```diff
@@ -14,18 +14,14 @@ jobs:
...
-          -   php: 7.3
-              symfony-version: '4.*'
-          -   php: 7.4
-              symfony-version: '5.*'
-          -   php: 8.0
-              symfony-version: '5.*'
+          -   php: 8.1
+              symfony-version: '5.4'
+          -   php: 8.1
+              symfony-version: '6.4'
```
stof, line 24: the step below expects `symfony-version` to be a version constraint. Using `6.4` here will make it install 6.4.0, which is not what we want

acoulton: Thanks @stof, good catch. I'd not registered the install here is different to the one in behat.

carlos-granados (approved):

PR #275, opened 2024-12-07
Upgrade to phpunit 10

uuf6429:


.phpunit.cache/test-results
```diff
@@ -0,0 +1 @@
...hat\\Gherkin\\ParserTest::testParser#outline_with_multiple_examples":0.002,"Tests\\Behat\\Gherkin\\ParserTest::testParser#outline_with_spaces":0.002,"Tests\\Behat\\Gherkin\\ParserTest::testParser#outline_with_step_table":0.002,"Tests\\Behat\\Gherkin\\ParserTest::testParser#pystring":0.001,"Tests\\Behat\\Gherkin\\ParserTest::testParser#ru_addition":0.002,"Tests\\Behat\\Gherkin\\ParserTest::testParser#ru_commented":0.001,"Tests\\Behat\\Gherkin\\ParserTest::testParser#ru_consecutive_calculations":0.002,"Tests\\Behat\\Gherkin\\ParserTest::testParser#ru_division":0.002,"Tests\\Behat\\Gherkin\\ParserTest::testParser#start_comments":0.001,"Tests\\Behat\\Gherkin\\ParserTest::testParser#tables":0.002,"Tests\\Behat\\Gherkin\\ParserTest::testParser#tags_sample":0.004,"Tests\\Behat\\Gherkin\\ParserTest::testParser#test_unit":0.001,"Tests\\Behat\\Gherkin\\ParserTest::testParser#trimpystring":0.001,"Tests\\Behat\\Gherkin\\ParserTest::testParser#undefined_multiline_args":0.001,"Tests\\Behat\\Gherkin\\ParserTest::testParserResetsTagsBetweenFeatures":0,"Tests\\Behat\\Gherkin\\ParserTest::testSingleCharacterStepSupport":0,"Tests\\Behat\\Gherkin\\ParserTest::testParsingManyCommentsShouldPass":0.001}}
```
acoulton, line 1: Should this have been committed? I think it should be gitignored?

package.xml.tpl
```diff
@@ -38,7 +38,7 @@
             <file role="php" baseinstalldir="gherkin" name="vendor/.composer/autoload.php" />
             <file role="php" baseinstalldir="gherkin" name="vendor/.composer/autoload_namespaces.php" />
             <file role="php" baseinstalldir="gherkin" name="phpdoc.ini.dist" />
-            <file role="php" baseinstalldir="gherkin" name="phpunit.xml.dist" />
+            <file role="php" baseinstalldir="gherkin" name="phpunit.dist.xml" />
```
acoulton, line 41: Is this package.xml.tpl actually used for anything? It looks quite outdated (minimum php = 5.3.1, for example).

phpunit.dist.xml
```diff
@@ -1,25 +1,21 @@
...
-<phpunit backupGlobals="false"
-         backupStaticAttributes="false"
+<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/10.5/phpunit.xsd"
+         backupGlobals="false"
          colors="true"
-         convertDeprecationsToExceptions="true"
-         convertErrorsToExceptions="true"
-         convertNoticesToExceptions="true"
-         convertWarningsToExceptions="true"
```
acoulton, line 9: I think we need to add the equivalent phpunit 10 failure modes e.g.
```
         failOnDeprecation="true"
         failOnNotice="true"
         failOnWarning="true"
         failOnRisky="true"
         displayDetailsOnIncompleteTests="true"
         displayDetailsOnSkippedTests="true"
         displayDetailsOnTestsThatTriggerDeprecations="true"
         displayDetailsOnTestsThatTriggerErrors="true"
         displayDetailsOnTestsThatTriggerNotices="true"
         displayDetailsOnTestsThatTriggerWarnings="true"
```

tests/Behat/Gherkin/Filter/AbstractFilterTestBase.php
```diff
@@ -7,7 +7,7 @@
 use Behat\Gherkin\Parser;
 use PHPUnit\Framework\TestCase;
 
-abstract class FilterTest extends TestCase
+abstract class AbstractFilterTestBase extends TestCase
```
acoulton, line 10: In our own projects we've gone with `xxTestCase` to follow the `TestCase` base class naming of phpunit, but I think `AbstractXXXTestBase` is also fine.

acoulton (commented):
Thanks @uuf6429 this looks generally good, a couple of comments

.phpunit.cache/test-results
uuf6429: oops, nice catch 😆

tests/Behat/Gherkin/Filter/AbstractFilterTestBase.php
uuf6429: I think I like -TestCase better, I'll go for that.

package.xml.tpl
uuf6429: I have absolutely no idea. I understood it might relate to PEAR/PECL somehow? I couldn't find reference to it anywhere though.

Same situation with `phpdoc.ini`.

.gitignore
```diff
@@ -3,3 +3,4 @@ vendor
 composer.phar
 composer.lock
 .phpunit.result.cache
+phpunit.xml
```
stof, line 6: you need to also ignore the `.phpunit.cache` folder (as PHPUnit 10 changed the location of the cache).

Note that you will also need to amend your commit to avoid committing the cache (which is currently part of the PR)

package.xml.tpl
stof: This is indeed a left-over from the time this package was available through a PEAR channel years ago.

acoulton: I think we can delete it, also phpdoc.ini.dist, and I can't find any usages of the `libpath.php` either so I'm guessing that might also be PEAR/something-pre-composer-related. Probably best to remove those in a separate PR, I can do that at some point unless you have a chance @uuf6429

uuf6429: I'll do all the changes later on today, I can also do that.

acoulton (approved):
Thanks 😀

stof (approved):

PR #276, opened 2024-12-07
PHPStan setup & level1 fixes

uuf6429:
~~Based on #277.~~

This PR sets up PHPStan and applies fixes until level 1.

~~The plan is to fix the problems separately later on.~~


uuf6429:
@acoulton one motivation for adding PHPStan is that during the recent title change, I realised that there is a big discrepancy in types. For example:
```php
class SomeRecord
{
    /** @var string */
    private $keyword;                                         // 1️⃣

    /** @var int */
    private $line;                                            // 1️⃣

    /**
     * @param string $keyword
     * @param int $line
     */
    public function __construct($keyword, $line) {            // 2️⃣
        $this->keyword = $keyword;
        $this->line = $line;
    }

    /** @return string */                                     // 3️⃣
    public function getKeyword(){ return $this->keyword; }    // 4️⃣

    /** @return int */                                        // 3️⃣
    public function getLine(){ return $this->line; }          // 4️⃣
}

class SomeRecordTest extends TestCase
{
    public function testSomething(): void
    {
        $r = new SomeRecord('word', null);                    // 5️⃣
        ...
    }
}
```

The main problem: `$line` is defined as `int`, but many tests pass a `null`. This happens a lot, for various other properties.

I would like to fix that and declare types in a backword compatible way:
- 1️⃣ type can be safely defined (for private props), but not sure if it should be nullable because of the test (see point 5)
- 2️⃣ type can probably be defined - it will affect users passing invalid types (question from 1️⃣ about null also applies)
- 3️⃣ if tests are to be kept as is, then null would be a valid value and the PHPDoc should be updated.
- 4️⃣ probably we cannot declare the return type without a BC break.
- 5️⃣ from my understanding, null should not be allowed and a valid value - such as -1 or 0 should be used instead.

What do you think?

acoulton:
@uuf6429 I haven't looked at this or your PHP-CS pulls in detail as I realised they're marked draft and overlap a bit with the PHPUnit / docs PR so I assume we should review & merge those first.

On the type-safety, I think (if we're certain it is **only** tests that are passing nulls):

* I agree, those tests are invalid and should be passing a value that mirrors real-world usage (so e.g. `1`).
* It is definitely fine to put a typehint on the private properties
* It's also fine to put a typehint on the constructor params - if the phpdoc typed them as `int` then IMO it is not a BC break to start enforcing that (and it will be enforced by adding a typehint to the property anyway, so might as well be done on the constructor)
* As you say, we can't add concrete return types without breaking BC - unless the class is `final`.

carlos-granados:
Regarding this draft pull request:

In the main Behat repository we use Psalm, while this PR introduces PHPStan in this one. I believe that it would be better if we unified this and used a single static analysis tool. Otherwise this can be confusing when you are working in one code base or the other and you need to follow different processes depending on which repo you are on. So I believe that we have two options:
- Update the main Behat repo to use PHPStan instead of Psalm. I believe that right now there are many more projects using PHPStan than using PSalm, so this may also help with developers adding code to the project as it is more probable that they are familiar with PHPStan than with PSalm. Our current PSalm level is not too high and we almost don't use any PSalm specific annotations so I believe that a transition to PHPStan should not be too difficult
- Or use PSalm in this repository as well

Also, I am not a big fan of baseline files. They have a number of drawbacks:
- Accumulation of Ignored Errors: Over time, the baseline file can grow as it accumulates ignored errors, making it challenging to manage and potentially leading to overlooked issues. 
- Stale Error Records: As the code evolves, some errors in the baseline may become obsolete. PHPStan doesn't automatically remove these outdated entries, requiring manual intervention to keep the baseline current. 
- Potential for Complacency: Relying heavily on the baseline might lead to a false sense of code quality, as existing issues remain unaddressed. 
- Complexity in Management: In projects with multiple developers or extensive codebases, managing the baseline can become complex
I prefer to start with a low level of rules and slowly update the code until we reach a level which gives us enough confidence in it

uuf6429:
@carlos-granados as you know, this is just a draft, so I wouldn't focus too much on it yet. But to answer your points:

1. We should just switch to PHPStan in my opinion, especially since we have already implemented it in a few other behat/mink project (e.g. some mink drivers).
2. I agree about the baseline - the point here is that it is a workaround so that this PR could be merged separately (or at least reviewed so). Otherwise it will end up containing a lot of fixes and/or a broken or incomplete pipeline.

The next point would be if we merge this PR and the fixes PR into one (to avoid polluting github) or if we keep them separate (and e.g. the baseline would be stored somewhere in the git history). This can be discussed later though.

PS: A small comment on one of your points: if a baselined error becomes obsolete, PHPStan will fail automatically (unless disabled by a specific setting). Just an FYI.

acoulton:
I'd agree with just using PHPStan - I noticed the doctrine project [have just decided to drop psalm](https://www.doctrine-project.org/2024/10/14/doctrine-core-team-meetup-2024.html#:~:text=Psalm%20and%20PHPStan%20going%20forward,repositories%20over%20the%20next%20weeks.) and use PHPStan-only in future. 

Also agree with @carlos-granados that my preference would be to introduce it at the lowest-compatible level (without a baseline) and then gradually increase the level in parallel with fixing the issues. This also means it's easy to see from the config which level we have *actually* achieved.

src/Behat/Gherkin/Parser.php
```diff
@@ -482,7 +482,7 @@ protected function parseOutline()
             }
         }
 
-        if (empty($examples)) {
+        if (!count($examples)) {
```
carlos-granados, line 485: I prefer to use `if (count($examples) === 0) {`, I think it is easier to read and clearer to understand. Just an opinion, feel free to ignore it

uuf6429: I don't have a strong opinion against it. I'll replace it later.

uuf6429: done

acoulton (commented):
Thanks for this @uuf6429 - I spotted a couple of things / questions.

src/Behat/Gherkin/Loader/CucumberNDJsonAstLoader.php
```diff
@@ -112,56 +112,47 @@ static function ($child) {
...
-                    static function ($child) {
-                        return isset($child['background']);
-                    }
+                    static fn ($child) => isset($child['background']),
                 )
             )
         );
 
-        return count($backgrounds) == 1 ? $backgrounds[0] : null;
+        return count($backgrounds) === 1 ? $backgrounds[0] : null;
```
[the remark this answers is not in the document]
uuf6429: I almost fell for it too.. then I saw that the method's code is also weird - looping over each item and then returning the first one (or null).

But here's the thing... **it specifically checks for one item** (not *more than 1*).
`array_shift` is unsuitable because it also (naturally) handles multiple items.

In any case, I think the function could be simplified - we still don't need to loop over all elements and build objects.

src/Behat/Gherkin/Loader/CucumberNDJsonAstLoader.php
```diff
@@ -173,6 +164,7 @@ static function ($cell) {
                     },
                     $tableJson['tableHeader']['cells']
                 );
+                $table[$tableJson['tableHeader']['location']['line']] = array_column($tableJson['tableHeader']['cells'], 'value');
```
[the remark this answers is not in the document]
uuf6429: Good catch... that was supposed to be a replacement of the previous lines.

src/Behat/Gherkin/Loader/CucumberNDJsonAstLoader.php
```diff
@@ -181,6 +173,7 @@ static function ($cell) {
                         },
                         $bodyRow['cells']
                     );
+                    $table[$bodyRow['location']['line']] = array_column($bodyRow['cells'], 'value');
```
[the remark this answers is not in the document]
uuf6429: Same as the case above.

src/Behat/Gherkin/Loader/DirectoryLoader.php
```diff
@@ -12,8 +12,6 @@
 
 use Behat\Gherkin\Gherkin;
 use Behat\Gherkin\Node\FeatureNode;
-use RecursiveDirectoryIterator;
-use RecursiveIteratorIterator;
```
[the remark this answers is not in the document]
uuf6429: That was unintentional php-cs-fixer leftovers. I'll check the entire PR for that and the other points you mentioned.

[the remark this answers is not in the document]
uuf6429: Done; I checked it earlier of course, but unfortunately I missed those points you found.

This came up because I had some major conflicts with master and took a lot of careful conflict solving...turns out it wasn't as perfect as I'd have hoped.

phpstan.dist.neon
```diff
@@ -0,0 +1,6 @@
+parameters:
+    level: 1
+    paths:
+        - bin
```
stof, line 4: will phpstan actually check the bin files that have no extension ?

src/Behat/Gherkin/Loader/ArrayLoader.php
```diff
@@ -289,22 +289,22 @@ private function examplesAreInArray($exHash)
      * Processes cases when examples are in the form of array of arrays
      * OR in the form of array of objects.
      *
-     * @param $exHash array hash
+     * @param $examplesHash array hash
      * @param $examplesKeyword string
      * @param $examples array
```
stof, line 294: those tags are invalid. The type must be before the name, not after

uuf6429: Hmm, I see that in other cases, we simply used `$hash`. I think I'll do the same here.

stof: what I mean is that we should use `@param string $examplesKeyword`, not `@param $examplesKeyword string`. currently, `string` is not the type but the human-readable description.

uuf6429: oooh, I totally missed that

stof: To be exact, they are not invalid as far as a phpdoc parser is concerned (the type is optional). But they are not the intended behavior (and would report a missing type in phpstan level 6 when phpstan requires knowing the type)

uuf6429: Done; I also simplified/cleaned-up the code a bit as well.

phpstan.dist.neon
uuf6429: Hmm, it does have a side-effect that it doesn't work on Windows... `vendor/bin/phpstan analyze bin/*` - the `*` is expanded by the shell, not PHP (and on Windows PHPStan simply complains that there is no "%PWD%/bin/*" file). :/

src/Behat/Gherkin/Loader/CucumberNDJsonAstLoader.php
```diff
@@ -112,75 +110,57 @@ static function ($child) {
...
+        }
+
+        $background = array_shift($backgrounds);
+
+        return new BackgroundNode(
+            $background['background']['name'],
+            self::getSteps($background['background']['steps'] ?? []),
+            $background['background']['keyword'],
+            $background['background']['location']['line']
+        );
```
acoulton, line 131: This is a much nicer solution, thanks

acoulton (approved):
Great, thanks @uuf6429

src/Behat/Gherkin/Parser.php
```diff
@@ -731,25 +731,28 @@ private function parseTableRows()
...
+        return new StepNode(
+            $node->getKeyword(),
+            $node->getText(),
+            $node->getArguments(),
+            $node->getLine(),
+            $keywordType
+        );
     }
 
+    /**
```
carlos-granados, line 753: Wouldn't it be better to just change the return type to `never`?

carlos-granados (commented):
Looking good, thanks @uuf6429 , just a small comment

carlos-granados (approved):
All good now, thanks @uuf6429

PR #278, opened 2024-12-09
Remove redundant files

uuf6429:


acoulton (changes requested):
Looks like the .phpunit.cache/test-results file has crept in again - otherwise this looks good, thanks

acoulton (approved):
Thanks!

carlos-granados (approved):
Thanks @uuf6429

PR #279, opened 2024-12-24
change: Deprecate Gherkin::VERSION constant and change file cache namespacing

acoulton:


carlos-granados (approved):
Looks good, thanks @acoulton

PR #280, opened 2024-12-26
PHP-CS-Fixer enable risky and list rules

uuf6429:


uuf6429:
Should I git-ignore-blame 19bd978d4e807127a41af9fb62dfd0ecc8f010bd?

PR #281, opened 2025-01-31
Update to PHPStan level 3; various fixes

uuf6429:


tests/Behat/Gherkin/Filter/LineFilterTest.php
```diff
@@ -71,19 +71,20 @@ public function testFilterFeatureScenario()
...
 
-    public function testFilterFeatureOutline()
+    public function testFilterFeatureOutline(): void
     {
         $filter = new LineFilter(13);
         $feature = $filter->filterFeature($this->getParsedFeature());
-        /* @var OutlineNode[] $scenarios */
         $this->assertCount(1, $scenarios = $feature->getScenarios());
         $this->assertSame('Scenario#3', $scenarios[0]->getTitle());
+        $this->assertInstanceOf(OutlineNode::class, $scenarios[0]);
```
uuf6429, line 80: These `assertInstanceOf()` are a convenient way to (phpunit-)test the code as well as tell phpstan that the `ScenarioInterface` object is in fact an `OutlineNode`.

src/Behat/Gherkin/Loader/DirectoryLoader.php
```diff
@@ -57,6 +58,9 @@ public function supports($resource)
     public function load($resource)
     {
         $path = $this->findAbsolutePath($resource);
+        if ($path === false) {
+            throw new \LogicException("Unable to locate absolute path of supported resource: $resource");
```
uuf6429, line 62: I hope that what is happening here is clear - it should not be possible for this exception to be thrown unless something happens between calling `supports()` and `load()`.
In any case, feel free to come up with a better message.. I'm a bit unsure with the current one.

src/Behat/Gherkin/Loader/AbstractFileLoader.php
```diff
@@ -50,7 +50,7 @@ protected function findRelativePath($path)
      *
      * @param string $path Relative path
      *
-     * @return string
+     * @return false|string
```
uuf6429, line 53: I couldn't see any better way to solve this - downstream users with PHPStan that extended this class will now get warnings, unless they fix it in a similar way to (for example), the `DirectoryLoader` changes in this PR.

acoulton: With tools like phpstan becoming more prevalent, it's an interesting question whether changing the phpdoc / annotation types of a method counts as a BC break or not...

In this case, however, it's clearly a bugfix - the previous value is obviously incorrect - so I think you're right to fix it like this.

src/Behat/Gherkin/Loader/DirectoryLoader.php
acoulton: Or I guess if you call it yourself from somewhere without calling `supports` first - we don't know it's definitely been checked.

I might introduce a parent method `requireAbsolutePath($resource): string` to centralise / simplify this logic in the implementations.

And for a message maybe more like: "Cannot find absolute path for '$resource': filesystem may have changed during execution or `supports()` may not have been called"?

acoulton (commented):
Looking good, thanks @uuf6429 - just the discussion on the absolute path checks.

src/Behat/Gherkin/Loader/DirectoryLoader.php
uuf6429: I rewrote this part with the other suggestion (centralizing the check), so that message is no longer fully relevant. Feel free to comment on the change instead.

acoulton (approved):
Great, thanks :-)

carlos-granados (approved):
looking good, thanks @uuf6429

PR #282, opened 2025-02-06
Fixes for PHPStan level 5

uuf6429:


bin/update_cucumber
```diff
@@ -1,11 +1,6 @@
 #!/usr/bin/env php
 <?php
 
-if (PHP_VERSION_ID < 70400) {
```
uuf6429, line 4: This check is no longer needed since this package can only be installed on PHP 8.

Additionally, [PHPStan understands this and shows a warning](https://github.com/phpstan/phpstan/issues/12528).

composer.json
```diff
@@ -41,7 +41,7 @@
 
     "autoload-dev": {
         "psr-4": {
-            "Tests\\Behat\\": "tests/Behat/"
+            "Tests\\": "tests/"
```
uuf6429, line 44: This helps:
1. IDEs (e.g. PhpStorm or IDEA) correctly detect the tests directory as the "tests root"
2. It decrease the wtf factor if someone adds test classes elsewhere which wouldn't load otherwise
3. It may not be the best, performance-wise, but I think it's not so critical seeing as it only impacts tests (i.e. `composer --dev`).

src/Behat/Gherkin/Filter/LineFilter.php
```diff
@@ -22,16 +22,19 @@
...
+    /**
+     * @var int
+     */
     protected $filterLine;
 
     /**
      * Initializes filter.
      *
-     * @param string $filterLine Line of the scenario to filter on
+     * @param int|numeric-string $filterLine Line of the scenario to filter on
```
uuf6429, line 33: Passing a non-numeric-string is Doing It Wrong and deserves a PHPStan slap. ;)

src/Behat/Gherkin/Parser.php
```diff
@@ -83,7 +85,7 @@ public function parse($input, $file = null)
         while ('EOS' !== ($predicted = $this->predictTokenType())) {
             $node = $this->parseExpression();
 
-            if ($node === null || $node === "\n") {
```
uuf6429, line 86: I checked and couldn't find evidence of any parser method returning a null (in addition, the PHPDocs never mentioned nulls either, hence the PHPStan warning).

src/Behat/Gherkin/Parser.php
```diff
@@ -262,14 +264,12 @@ protected function parseFeature()
                 ));
             }
 
-            if (!$node instanceof ScenarioNode) {
```
uuf6429, line 265: This condition and the other cases below were flagged as impossible by PHPStan because:
1. there is already an earlier branch unconditionally handling such instance
2. `$node` is never overwritten

src/Behat/Gherkin/Filter/LineRangeFilter.php
```diff
@@ -22,14 +22,20 @@
...
+     */
     protected $filterMaxLine;
 
     /**
      * Initializes filter.
      *
-     * @param string $filterMinLine Minimum line of a scenario to filter on
-     * @param string $filterMaxLine Maximum line of a scenario to filter on
+     * @param int|numeric-string $filterMinLine Minimum line of a scenario to filter on
+     * @param int|numeric-string|'*' $filterMaxLine Maximum line of a scenario to filter on
```
uuf6429, line 38: I verified that PHPStan handles this well: https://phpstan.org/r/bd2527e3-519d-491e-9eb6-7208761748dd

tests/Behat/Gherkin/ParserTest.php
```diff
@@ -171,13 +155,13 @@ protected function parseEtalon($etalon)
...
 
-    public function testParsingManyCommentsShouldPass()
+    public function testParsingManyCommentsShouldPass(): void
     {
         if (!extension_loaded('xdebug')) {
             $this->markTestSkipped('xdebug extension must be enabled.');
         }
         $defaultPHPSetting = 256;
-        $this->iniSet('xdebug.max_nesting_level', $defaultPHPSetting);
+        ini_set('xdebug.max_nesting_level', $defaultPHPSetting);
```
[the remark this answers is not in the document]
uuf6429: Hmm, I see that the [GH workflow doesn't seem to run with xdebug](https://github.com/Behat/Gherkin/actions/runs/13174001644/job/36769284227#step:3:16), I'll try it later on at home and see what happens (since this seems to relate exclusively to xdebug).

PS: We should probably integrate codecov into this project. 🤔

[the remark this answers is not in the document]
uuf6429: Ah, I see what you mean now, something like:
```php
try {
    ini_set(<new value>);
    <the test>
} finally {
    ini_set(<old value>);
}
```

I'll fix it later on today, thanks for pointing it out.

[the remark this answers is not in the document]
uuf6429: Done, I tried two approaches: 4f02b77285d371147f16828422d4b51b20afa1ec and e23890489e5ae5a31522f41d011c43ec4fb805e7.
Let me know if you prefer the first one, and I'll revert the last commit.

src/Behat/Gherkin/Parser.php
```diff
@@ -31,6 +31,8 @@
  * $featuresArray = $parser->parse('/path/to/feature.feature');
  *
  * @author Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * @phpstan-type TParsedExpressionResult FeatureNode|BackgroundNode|ScenarioNode|OutlineNode|ExampleTableNode|TableNode|PyStringNode|StepNode|string|"\n"|''
```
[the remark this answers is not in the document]
acoulton: I'm not sure I follow why they need to be mentioned separately?

tests/Behat/Gherkin/ParserTest.php
```diff
@@ -171,15 +155,22 @@ protected function parseEtalon($etalon)
...
+        if ($oldMaxNestingLevel === false) {
+            throw new \RuntimeException('Could not set INI setting value');
+        }
+
+        try {
+            $lineCount = 150; // 119 is the real threshold, higher just in case
+            $this->assertNull($this->getGherkinParser()->parse(str_repeat("# \n", $lineCount)));
+        } finally {
+            ini_set('xdebug.max_nesting_level', $oldMaxNestingLevel);
+        }
```
acoulton, line 174: if this is the only place we're customising ini values, I prefer the inline syntax you have here, I think it's clearer than using the helper function in your first commit.

acoulton (approved):

src/Behat/Gherkin/Parser.php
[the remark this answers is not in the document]
stof: `string|"\n"|''` will be simplified to `string` by phpstan anyway, so I don't see how this would be helpful.

uuf6429:
@stof all points that have been raised are solved, anything else to move forward on this PR?

tests/Behat/Gherkin/Cucumber/CompatibilityTest.php
```diff
@@ -117,73 +106,77 @@ public function testBadFeaturesDoNotParse(SplFileInfo $file)
...
     }
 
-    public static function badCucumberFeatures()
+    /**
+     * @return iterable<string, array{file: SplFileInfo}>
+     */
+    public static function badCucumberFeatures(): iterable
     {
-        return self::getCucumberFeatures('/bad');
+        yield from self::getCucumberFeatures('/bad');
```
stof, line 122: why using `yield from` instead of returning the original iterable directly ?

stof (approved):

tests/Behat/Gherkin/Cucumber/CompatibilityTest.php
uuf6429: Hmm, in this context it's not so helpful. I'll change it back.

composer.json
uuf6429: I wouldn't be against it, but then it does become a bit inconsistent with src:
image
(although I suppose we can do it for src too - is a change in filesystem structure considered a BC break?)

stof: it is not a BC break (assuming you update the autoload config to be valid for the new path), unless your project requires users to hardcode a path to a file inside your package (which is not the case for Gherkin)

PR #287, opened 2025-02-12
Handle null path better

### LINKED ISSUE #284: Symfony deprecation message (copy&paste mistake?)
I just found the following code:

https://github.com/Behat/Gherkin/blob/e016902ab9dcca51d72d22c64d1424a2d03e7b14/src/Behat/Gherkin/Node/FeatureNode.php#L263

Considering this project and that line has absolutely nothing to do with Symfony, I think it makes sense to update it.

On a different perspective, PHPStan right complains because that line is not terminal - the code will continue running, in which case it tries executing `strspn(null, ...` with god-knows-what consequences. 😅 What should I do?

uuf6429:
Closes #284.

### DIFF
--- a/src/Behat/Gherkin/Node/FeatureNode.php
+++ b/src/Behat/Gherkin/Node/FeatureNode.php
@@ -254,13 +254,11 @@ class FeatureNode implements KeywordNodeInterface, TaggedNodeInterface
      * @param string $file A file path
      *
      * @return bool
-     *
-     * @see https://github.com/symfony/filesystem/blob/master/Filesystem.php
      */
     protected function isAbsolutePath($file)
     {
         if ($file === null) {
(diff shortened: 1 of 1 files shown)


src/Behat/Gherkin/Node/FeatureNode.php
```diff
@@ -254,13 +254,11 @@ public function getLine()
      * @param string $file A file path
      *
      * @return bool
-     *
-     * @see https://github.com/symfony/filesystem/blob/master/Filesystem.php
```
acoulton, line 258: I think it might be better to keep this comment a) to give at least a bit of context (it's hard looking at the code to understand what it's doing, at least knowing where it came from provides a pointer to go and find history/tests/etc if ever needed) and b) to provide some credit to the Symfony project given this appears to be copy/pasted?

acoulton (changes requested):
The exception LGTM, just the docblock comment. Thanks

acoulton (approved):
Great, thanks

carlos-granados (approved):
Thanks @uuf6429

PR #288, opened 2025-02-19
Relocate files to remove redundant directories

uuf6429:
This PR does what was [suggested here](https://github.com/Behat/Gherkin/pull/282#discussion_r1961299591). I thought it would make sense / be cleaner to make a separate PR.

Changes:
1. `./src` is now loaded as PSR-4
2. `./src` represents `Behat\Gherkin\`
3. `./tests` represents `Tests\Behat\Gherkin\`
4. fixed namespace of existing test

~~⚠️ I've added a git-ignore-blame entry, so this PR **should not be squashed when accepted.**~~


uuf6429:
@acoulton, @stof: this should be finished now. Only concern is that existing PRs will probably be faced with some conflicts after merging, but I guess that's unavoidable.

acoulton (approved):
@uuf6429 I rebased and force pushed this to resolve the conflict with #282 (as that PR still had the previous proposal for changing the composer.json autoload config).

The commits are still showing as yours but now showing a warning that they're unverified (as I obviously can't sign as you). Feel free to rebase and force push again to resolve that.

I have dropped the commit with the git-blame-ignore (the commit ref was outdated anyway) - AFAIK this is not actually required/appropriate for straight file moves, if you look at the blame on https://github.com/uuf6429/Gherkin/blame/chore/relocate-files/tests/ParserTest.php you'll see that it is correctly showing the history despite the file move.

I don't think that conflicts with other PRs will be an issue - the majority are old and have conflicts already anyway, and I've generally found git is usually quite reliable about resolving file renames so it should be quick enough to update anything that needs it.

uuf6429:
@acoulton done, thanks for taking the time to fix things.

acoulton (approved):
Great, thanks :-)

mvorisek:
Hi, behat/gherkin v4.13.0 release has broken our CI - https://github.com/atk4/ui/actions/runs/14864378786/job/41737354734.

I have verified this is the only changed dependency. If I downgrade to v4.12.0, the issue is gone.

Our Behat configuration is https://github.com/atk4/ui/blob/6.0.0/behat.yml.dist. Quite minimal config.

Is there a bug or do we need to change something?

acoulton:
@mvorisek apologies for the inconvenience. This is tracked at #317 - basically, you need to either update behat or pin to behat/gherkin 4.12.

In your case:

* I can see from the GitHub Actions logs that your CI is installing behat/behat v3.15.0 (this release is no longer supported)
* That is because your composer.json requires `behat/mink-extension` which was officially abandoned 2 years ago (but has not been updated since 2018 and was forked to friends-of-behat/mink-extension in Jan 2020). That is holding you back to symfony/config v4.0, which is unsupported since Jan 2019, and that in turn is holding back several other packages including the more recent Behat releases.

If you switch from `behat/mink-extension` to `friends-of-behat/mink-extension` you will get the drop-in replacement with updated dependency constraints, which will resolve this as well as generally unblocking several of your dependencies. I've sent you a PR in https://github.com/atk4/ui/pull/2285

mvorisek:
Love you perfect guidance, thank you a lot! ❤

uuf6429:
So to be clear, the issue is not related to this PR (it was only changing the location of test files), right?

mvorisek:
https://github.com/atk4/ui/pull/2285 fixed the issue for us. This PR is probably fine.

acoulton:
@uuf6429 it is ultimately caused by this - this PR moved the src files as well as the test files, and other packages depended on finding the i18n.php by looking relative to one of the class files. 

We discussed it in #290 when it was first merged, and thought we'd found a solution, but we hadn't properly anticipated the impact on non-behat users or that there will often be cases where people can get a more recent gherkin (as it has no deps) but can't get the latest Behat.

uuf6429:
@acoulton Just checked the PR and your right. I missed that whole discussion and unfortunately the PR description is not really up-to-date. :/ I'll fix that at least.

PR #289, opened 2025-02-23
Improve tag handling; update test

uuf6429:
This isn't a new feature, but rather a small improvement on tag handling.

Here's what changed:
1. `hasTag`/`hasTags` code is always the same, so I moved it into a trait (that depends on abstract method `getTags`).
2. I removed the somewhat redundant PHPDoc comments, in favour of the one provided by the interface
3. `ExampleTableNode` now implements `TaggedNodeInterface` and the new trait - I couldn't see why it wouldn't have that in the first place
4. I've updated a test to cover point 3.

It would have been nice to move `getTags` to the trait too, but at this PHP level we can't define abstract properties or property requirements, so I think this is the safest and cleanest approach at the moment.

### DIFF
--- a/src/Node/TaggedNodeTrait.php
+++ b/src/Node/TaggedNodeTrait.php
@@ -12,6 +12,8 @@ namespace Behat\Gherkin\Node;
 
 /**
  * This trait partially implements {@see TaggedNodeInterface}.
+ *
+ * @internal
  */
 trait TaggedNodeTrait
 {
--- a/src/Node/TaggedNodeTrait.php
+++ b/src/Node/TaggedNodeTrait.php
@@ -1,0 +1,40 @@
+<?php
+
+/*
+ * This file is part of the Behat Gherkin Parser.
+ * (c) Konstantin Kudryashov <ever.zet@gmail.com>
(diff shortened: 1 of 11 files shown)


src/Node/TaggedNodeTrait.php
```diff
@@ -0,0 +1,40 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Behat\Gherkin\Node;
+
+/**
+ * This trait partially implements {@see TaggedNodeInterface}.
+ */
+trait TaggedNodeTrait
```
stof, line 16: I suggest tagging this trait as `@internal`. I don't think we should consider the trait itself as part of the public API covered by BC. It should stay an implementation detail of our classes

uuf6429: I was totally wondering about that - good point.

acoulton (approved):
Thanks @uuf6429

carlos-granados (approved):
Looks good, thanks @uuf6429

PR #290, opened 2025-02-25
Use hardcoded path to i18n file if not provided

carlos-granados:
Alternate solution to the problem described in https://github.com/Behat/Behat/pull/1602 and https://github.com/Behat/Behat/pull/1603. Since the file that `CachedArrayKeywords` needs to load is in the same package, it does not need to be told its location by another package. We keep the option of passing a file name in case someone is using this class to load an i18n file from some other location. After this is merged in this package we should update Behat so that it does not attempt to provide the path to the file

### DIFF
--- a/tests/Behat/Gherkin/Keywords/CachedArrayKeywordsTest.php
+++ b/tests/Behat/Gherkin/Keywords/CachedArrayKeywordsTest.php
@@ -17,12 +17,13 @@ class CachedArrayKeywordsTest extends KeywordsTestCase
 {
     protected function getKeywords()
     {
-        return new CachedArrayKeywords(__DIR__ . '/../../../../i18n.php');
+        // Test with the default i18n file provided in this repository
+        return new CachedArrayKeywords();
     }
 
     protected function getKeywordsArray()
     {
-        return include __DIR__ . '/../../../../i18n.php';
+        return require __DIR__ . '/../../../../i18n.php';
     }
 
     protected function getSteps($keywords, $text, &$line, $keywordType)
--- a/tests/Behat/Gherkin/Keywords/CachedArrayKeywordsTest.php
+++ b/tests/Behat/Gherkin/Keywords/CachedArrayKeywordsTest.php
@@ -18,7 +18,7 @@ class CachedArrayKeywordsTest extends KeywordsTestCase
     protected function getKeywords()
     {
         // Test with the default i18n file provided in this repository
-        return new CachedArrayKeywords();
+        return CachedArrayKeywords::withDefaultKeywords();
     }
 
     protected function getKeywordsArray()
--- a/src/Behat/Gherkin/Keywords/CachedArrayKeywords.php
+++ b/src/Behat/Gherkin/Keywords/CachedArrayKeywords.php
@@ -19,13 +19,15 @@ namespace Behat\Gherkin\Keywords;
  */
 class CachedArrayKeywords extends ArrayKeywords
 {
+    private const I18N_FILE_LOCATION = __DIR__ . '/../../../../i18n.php';
+
     /**
      * Initializes holder with file.
      *
      * @param string $file Cached array path
      */
-    public function __construct($file)
+    public function __construct($file = null)
     {
-        parent::__construct(include $file);
+        parent::__construct(require $file ?? self::I18N_FILE_LOCATION);
     }
 }
--- a/src/Behat/Gherkin/Keywords/CachedArrayKeywords.php
+++ b/src/Behat/Gherkin/Keywords/CachedArrayKeywords.php
@@ -19,15 +19,18 @@ namespace Behat\Gherkin\Keywords;
  */
 class CachedArrayKeywords extends ArrayKeywords
 {
-    private const I18N_FILE_LOCATION = __DIR__ . '/../../../../i18n.php';
+    public static function withDefaultKeywords(): self
+    {
+        return new self(__DIR__ . '/../../../../i18n.php');
+    }
 
     /**
      * Initializes holder with file.
      *
      * @param string $file Cached array path
      */
-    public function __construct($file = null)
+    public function __construct($file)
     {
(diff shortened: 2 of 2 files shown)


src/Keywords/CachedArrayKeywords.php
```diff
@@ -19,13 +19,15 @@
  */
 class CachedArrayKeywords extends ArrayKeywords
 {
+    private const I18N_FILE_LOCATION = __DIR__ . '/../../i18n.php';
+
     /**
      * Initializes holder with file.
      *
      * @param string $file Cached array path
```
stof, line 27:
```suggestion
     * @param string|null $file Cached array path
```

src/Keywords/CachedArrayKeywords.php
```diff
@@ -19,13 +19,15 @@
...
      * Initializes holder with file.
      *
-     * @param string $file Cached array path
+     * @param string|null $file Cached array path
      */
-    public function __construct($file)
+    public function __construct($file = null)
     {
-        parent::__construct(include $file);
+        parent::__construct(include $file ?? self::I18N_FILE_LOCATION);
```
acoulton, line 31: should we change this to `require` while we're here? include would still fail if the file didn't exist, as the `false` wouldn't match the `array` typehint on the parent constructor - but `require` would give a better error message?

acoulton (approved):
I think this is a good solution, but maybe we should also revert moving the files in the `src` directory, at least for now?

Otherwise it will only work if we release Gherkin and Behat at the same time (and bump the minimum version requirement in Behat). 

Also Behat may not be the only thing using this class with a path to the Gherkin i18n file - I'm not sure the limited benefit we get from flattening the directory structure is worth even a small risk of causing problems for other people?

src/Keywords/CachedArrayKeywords.php
carlos-granados: Yes, that makes sense, updated

carlos-granados:
> I think this is a good solution, but maybe we should also revert moving the files in the `src` directory, at least for now?
> 
> Otherwise it will only work if we release Gherkin and Behat at the same time (and bump the minimum version requirement in Behat).
> 
> Also Behat may not be the only thing using this class with a path to the Gherkin i18n file - I'm not sure the limited benefit we get from flattening the directory structure is worth even a small risk of causing problems for other people?

@acoulton I think that moving the files is a good move that we should keep. No one should have been depending on the *location* of files, and if they did they would need to work to fix that unnecessary dependency. But it is true that currently this makes the update cumbersome, having to release Gherkin and Behat at the same time. What I think we could do:

- Create here a new v4.12.0 release branch not from the current master but from a previous point in the code before the files were moved.
- Move this PR to point to this new branch (adjusting the file location to the old location)
- Once this PR is merged, release the v4.12.0 version. This would provide a version that anyone could use to break the link with the location of the file. It would work both with code which provided the file location and for code that did not
- Merge this release branch into master, adjusting the file location to the new location
- Update Behat to require version v4.12.0 and adjust the code to stop sending the file location parameter. We would then have a Behat version that we could release whenever convenient

What do you think?

acoulton:
> No one should have been depending on the location of files, and if they did they would need to work to fix that unnecessary dependency.

Normally I would agree (and did when I merged #288). 

However, discovering now that we ship a class that it seems would always/usually need to be used with a path in this repo it seems a little more reasonable that people might have being doing that...

But I think your solution is good, I think the majority of users will be using this with Behat so as long as that upgrade path is smooth that's probably ok.

carlos-granados:
@acoulton I created the new release branch, modified this PR to point to the new branch and modified the code to point to the old location

tests/Behat/Gherkin/Keywords/CachedArrayKeywordsTest.php
```diff
@@ -17,12 +17,13 @@ class CachedArrayKeywordsTest extends KeywordsTestCase
 {
     protected function getKeywords()
     {
-        return new CachedArrayKeywords(__DIR__ . '/../../../../i18n.php');
+        // Test with the default i18n file provided in this repository
```
carlos-granados, line 20: 👍

acoulton:
@carlos-granados thanks - I noticed there's a test that was also specifying the path (both in the setup and the assertions). I've pushed a commit to your branch that removes the constructor argument, so that should test should now prove that the default file location is correct (it will fail when the files are moved).

While doing that, I wondered if it would be clearer to use a named constructor rather than an optional argument - something like:

```php
public static function withGherkinI18n(): self {
    // or maybe withDefaultKeywords // withDefaultTranslations?
    return new self(__DIR__ . '/../../../../i18n.php');
}

// and then leave the constructor parameter required.
```

That way it's explicit to end users what this class does with a null / missing constructor argument - `CachedArrayKeywords` alone isn't that obvious what it's actually loading / caching...

What do you think?

carlos-granados:
> @carlos-granados thanks - I noticed there's a test that was also specifying the path (both in the setup and the assertions). I've pushed a commit to your branch that removes the constructor argument, so that should test should now prove that the default file location is correct (it will fail when the files are moved).
> 
> While doing that, I wondered if it would be clearer to use a named constructor rather than an optional argument - something like:
> 
> ```
> public static function withGherkinI18n(): self {
>     // or maybe withDefaultKeywords // withDefaultTranslations?
>     return new self(__DIR__ . '/../../../../i18n.php');
> }
> 
> // and then leave the constructor parameter required.
> ```
> 
> That way it's explicit to end users what this class does with a null / missing constructor argument - `CachedArrayKeywords` alone isn't that obvious what it's actually loading / caching...
> 
> What do you think?

Good idea, I think that `withDefaultKeywords` would be the best option

carlos-granados:
> > Good idea, I think that withDefaultKeywords would be the best option
> 
> I can update that in this PR, unless you want to?

Yes, please, go ahead I am busy with something else at the moment

PR #294, opened 2025-03-01
test: Handle optional tableHeader when loading NDJson examples

### LINKED ISSUE #286: CucumberND table headers might be missing
The following line: https://github.com/Behat/Gherkin/blob/e016902ab9dcca51d72d22c64d1424a2d03e7b14/src/Behat/Gherkin/Loader/CucumberNDJsonAstLoader.php#L160

assumes that `tableHeader` is always set, but according to the spec schema, [tableHeader](https://github.com/cucumber/messages/blob/e1537b07e511feb6405ed9aa00261ff79d8a9710/jsonschema/GherkinDocument.json#L138) is [not required](https://github.com/cucumber/messages/blob/e1537b07e511feb6405ed9aa00261ff79d8a9710/jsonschema/GherkinDocument.json#L108), and therefore optional.

I discovered this while implementing PHPStan fixes, using hand-written schema in PHPStan types DSL:
```php
/**
 * @phpstan-type TExamples THasTags&array{
 *     location: TLocation,
 *     keyword: string,
 *     name: string,
 *     description: string,
 *     tableHeader?: TTableRow,              <-----
 *     tableBody: list<TTableRow>,
 *     id: string,
 * }
 */
```
What shall we do about this?

1. Make the PHPStan type mandatory (by removing the `?`) and generally ignore the code.
2. Make PHPStan ignore that line (`@phpstan-ignore-next-line`)
3. Skip that entire line (the top line shown above) when `$tableJson['tableHeader']` is not set`
4. Empty table header row?
5. Something else...?

I think option 3 sounds reasonable, but we need to sure it doesn't change behaviour.

uuf6429:
Closes #286 (refer there for details).


src/Loader/CucumberNDJsonAstLoader.php
```diff
@@ -51,7 +52,7 @@ private static function getFeature(array $json, $filePath)
 
         return new FeatureNode(
             $featureJson['name'] ?? null,
-            $featureJson['description'] ? trim($featureJson['description']) : null,
+            trim($featureJson['description'] ?? '') ?: null,
```
uuf6429, line 55: note: it seems odd to me that 'name'/title of the feature can be optional, but the description is mandatory. This change relaxes the restriction and slightly improves code readability.

src/Loader/CucumberNDJsonAstLoader.php
```diff
@@ -157,6 +158,15 @@ private static function getTables(array $items): array
             static function ($tableJson) {
                 $table = [];
 
+                if (!isset($tableJson['tableHeader'])) {
```
uuf6429, line 161: In the end I went for this approach - if this situation happens currently, we would have a PHP warning (or potentially an error, if they are captured by the library user), so while this change is not 100% backward compatible, I think it's a good compromise for an ambiguous situation.

tests/Loader/CucumberNDJsonAstLoaderTest.php
```diff
@@ -0,0 +1,175 @@
...
+    {
+        $this->loader = new CucumberNDJsonAstLoader();
+    }
+
+    public function testStringResourcesAreSupported(): void
+    {
+        $this->assertTrue($this->loader->supports('a string'));
+    }
+
+    public function testValidLoading(): void
```
uuf6429, line 36: I think most of the points here are already covered by the external cucumber parity tests, but I don't think it's a bad idea to have this here. It could serve as a blueprint/inspiration for testing other situations.

uuf6429:
@acoulton @carlos-granados could you give this a look when you have time please?

tests/Loader/CucumberNDJsonAstLoaderTest.php
```diff
@@ -0,0 +1,175 @@
...
+                ),
+            ],
+            $features,
+        );
+    }
+
+    public function testOutlineTableHeaderIsRequired(): void
+    {
+        $this->expectException(NodeException::class);
+        $this->expectExceptionMessage('Table header is required, but none was specified for the example on line 3.');
```
acoulton, line 132: I would always prefer to put the `expectException` at the last possible point before the code that is expected to throw. So in this case I'd generate the cucumber as a variable first, then put the expectations immediately before `$this->loader->load($resource)`

This reduces risk of false positives (which I appreciate are limited here) but also makes the failure output clearer if there's an unexpected exception during the setup of the test.

acoulton (commented):
Sorry for the delay on this @uuf6429 

Code-wise it looks good bar a small tweak to the test, and I agree that it's better to throw an exception than the current PHP warning / errorhandler call.

However I'm a bit confused if this is valid / if it will cause us to lose parity with upstream cucumber (or if it's just a theoretical case that can't actually ever be triggered by a valid Gherkin file)?

I can see from the links you posted in #286 that it does seem the table header is officially not required - and I share your confusion about how that can be the case, since as you say a Gherkin table header is always just the first row, there's no syntax to have a table that only has a body.

The docs at https://cucumber.io/docs/gherkin/reference/#scenario-outline also suggest the header is required:

> the Scenario Outline is run once for each row in the Examples section beneath it (not counting the first header row).

I can't find anything in the cucumber/messages issues or PRs to explain this. Notably, it looks like the `tableBody` was originally [also not required and was added later](https://github.com/cucumber/messages/commit/ce9f28944b8c19b32e04c22af156f2b52345ab1d#diff-69252728078b60bf59750707428b38924591264e92497fef90006dd33402e70aR112) - I wonder if the header has just been overlooked?

Were you able to get a better understanding of what's going on here? Is it maybe worth opening an issue on cucumber/messages to check if this is something we should actually be handling rather than rejecting?

tests/Loader/CucumberNDJsonAstLoaderTest.php
uuf6429: Typically I'm doing that sort of thing as well, but somehow I never thought about it for exception.. good catch.

uuf6429:
@acoulton

> Were you able to get a better understanding of what's going on here? Is it maybe worth opening an issue on cucumber/messages to check if this is something we should actually be handling rather than rejecting?

Not really, but @ciaranmcnulty [seems to have an idea](https://github.com/Behat/Gherkin/issues/286#issuecomment-2732822866) that to me sounds very plausible (it does mean that this PR would need to be reworked though).

Do you still think we should go for a more conclusive result e.g. by asking in cucumber/messages as you suggested?

acoulton:
@uuf6429 I have to say I'm still not sure what's going on here - it does look like `tableHeader` is still always present in the cucumber .ndjson files right up to 32.1.1 - and the table is definitely split between header & body.

I think we should at least hold off on this till #306 is merged with the latest cucumber/gherkin examples.

I would then be tempted to just throw if the key is not present so that we can review it if/when it arises, rather than assuming what the structure might look like without an example of that.

src/Node/TableNode.php
```diff
@@ -362,6 +362,14 @@ public function mergeRowsFromTable(TableNode $node)
...
     }
 
+    /**
+     * Returns true if table is empty (has no rows, but it might have a header).
+     */
+    public function isEmpty(): bool
+    {
+        return count($this->table) <= 1;
+    }
+
```
acoulton, line 372: Is this method required? I don't see it used anywhere.

I think it could cause confusion - for a table attached to a step node it's entirely valid to have a 1-row table which is not considered "empty" e.g.

```gherkin
When I fill in the following:
  | username | acoulton |
```

It's only `Examples` tables that are expected to have a header and subsequent data rows?

tests/Loader/CucumberNDJsonAstLoaderTest.php
```diff
@@ -0,0 +1,186 @@
...
+                    'en',
+                    $file,
+                    111,
+                ),
+            ],
+            $features,
+        );
+    }
+
+    public function testOutlineTableBodyIgnoredWhenTableHeaderIsMissing(): void
```
acoulton, line 131: Think this test method name is out of date?

src/Loader/CucumberNDJsonAstLoader.php
```diff
@@ -157,6 +158,15 @@ private static function getTables(array $items): array
...
+                if (empty($tableJson['tableHeader']) && !empty($tableJson['tableBody'])) {
+                    throw new NodeException(
+                        sprintf(
+                            'Table header is required when a table body is provided for the example on line %s.',
+                            $tableJson['location']['line'],
+                        )
+                    );
+                }
+
                 $table[$tableJson['tableHeader']['location']['line']] = array_column($tableJson['tableHeader']['cells'], 'value');
```
acoulton, line 170: I might be missing something, but it looks like now we're not testing/handling the case that originally prompted this whole thing, where `tableHeader` isn't present and `tableBody` is empty so there is no ExampleTableNode to load?

Don't we need to array_filter the `$items` - either before or after the array_map - to skip over any that are like that?

acoulton (changes requested):
@uuf6429 thanks, this looks very close to me - a couple of things I spotted (sorry if I'm missing something)

src/Node/TableNode.php
uuf6429: Not anymore 👍

tests/Loader/CucumberNDJsonAstLoaderTest.php
uuf6429: Correct, I'll update it.

src/Loader/CucumberNDJsonAstLoader.php
uuf6429: But `$items` represents multiple tables. But now that I think about it, the following condition should be clearer than the current one:
```php
                if (!isset($tableJson['tableHeader']) && $tableJson['tableBody'] === []) {
```

PS: There's a test that checks that the exception is thrown.

Let me know if anything is still unclear.

acoulton: Yeah, but I'm talking about this case https://github.com/Behat/Gherkin/issues/286#issuecomment-2849191554 where $items contains a table that has no `tableHeader` *and* no `tableBody` (the case that explains why `tableHeader` is not required in the schema).

As it stands, that will pass the check, then fail on the `tableHeader` key being undefined.

We just need to ignore examples like that, which come from gherkin like [this](https://github.com/cucumber/gherkin/blob/37f1871eb1a1350c40fe19beda5d28b083eeb976/testdata/good/incomplete_scenario_outline.feature#L13)

uuf6429: I see your point, as I understand the code is also wrong and we're missing a test case.

It should be something like:
```text
if (body is empty) {       // header is not important at this point
  return
}

if (header is not set) {   // here it is implied that body is not empty
  throw exception
}

load header and body
```
correct?

acoulton: Yep, perfect thanks, and then just wrap that in an array_filter to strip out any where it returned early.

It should pass with the NDJSON in that comment I linked.

acoulton (approved):
Thanks @uuf6429 this is great now

carlos-granados (approved):
Thanks @uuf6429 and @acoulton for all your work around this

PR #295, opened 2025-03-01
Improve test coverage

uuf6429:


src/Filter/NarrativeFilter.php
```diff
@@ -52,7 +52,9 @@ public function isFeatureMatch(FeatureNode $feature)
      *
      * @param ScenarioInterface $scenario Scenario or Outline node instance
      *
-     * @return bool
+     * @return false This filter is designed to work only with features
+     *
+     * @codeCoverageIgnore
```
uuf6429, line 57: I guess it doesn't make much sense to have a test that checks that these methods return false. PHPStan in fact would even complain that the test assertion will always be false.

src/Keywords/CucumberKeywords.php
```diff
@@ -50,6 +50,10 @@ public function __construct($yaml)
             throw $e;
         }
 
+        if (!is_array($content)) {
+            throw new ParseException(sprintf('Root element must be an array, but %s found.', get_debug_type($content)));
```
uuf6429, line 54: This a bit better than getting a `TypeError` from calling the parent constructor which requires an array value.

src/Parser.php
```diff
@@ -95,10 +95,12 @@ public function parse($input, $file = null)
             }
 
             if ($feature && $node instanceof FeatureNode) {
+                // @codeCoverageIgnoreStart
```
uuf6429, line 98: I could not replicate this and the other cases below. In particular, some seem to be impossible e.g. given that we only processing up to 4 different types of tokens, we should not get any other "unexpected" token unless the lexer is broken and lets such a token through.

tests/ParserTest.php
```diff
@@ -82,6 +84,83 @@ public function testSingleCharacterStepSupport(): void
         $this->assertCount(1, $scenario->getSteps());
     }
 
+    public function testParsingManyCommentsShouldPass(): void
```
uuf6429, line 87: This is just code formatting - I moved it from after protected methods.

tests/Keywords/CucumberKeywordsTest.php
```diff
@@ -39,4 +40,53 @@ protected function getSteps($keywords, $text, &$line, $keywordType)
...
+    }
+
+    public function testYamlFileMustBeReadable(): void
+    {
+        $tempFile = tempnam(sys_get_temp_dir(), 'test');
+
+        try {
+            file_put_contents($tempFile, "aaa:\n  bbb: cccc");
+            if (PHP_OS_FAMILY === 'Windows') {
+                exec('icacls ' . escapeshellarg($tempFile) . ' /deny Everyone:(R)');
```
uuf6429, line 79: I can confirm this works on Windows. I'm guessing the unix variation below should also work.

stof:
I fond it weird to have a PR saying it improves code coverage, and then seeing that half of the changes are not about adding new tests to increase coverage but adding `@codeCoverageIgnore` annotations to ignore uncovered code from the metric. This looks like gaming the metric to me.

uuf6429:
> I fond it weird to have a PR saying it improves code coverage, and then seeing that half of the changes are not about adding new tests to increase coverage but adding `@codeCoverageIgnore` annotations to ignore uncovered code from the metric. This looks like gaming the metric to me.

Probably I should have marked it as a "draft pr", since I'd like to discuss some of those points.

I think it makes sense to mark truly uncoverable code - to avoid people wasting time trying to cover it. But in some cases, it could actually be logically impossible - in which case some code might be removed. Some other case could be guaranteed with phpstan (and again some code guards could be removed). 

I'll get into more detail later on today when I get home.

src/Filter/NarrativeFilter.php
uuf6429: @stof for example in this case, if I add a test for it phpstan complains that `assertFalse(false)` is always true (`false` => because of the phpdoc).

If we want to make tests for this, we can change the PHPDoc to `bool` and then phpstan would not complain. I'm fine with both approaches.

uuf6429:
@stof @acoulton do we want to guarantee backward compatibility for test classes?

I'd like to change some protected methods of `ParserTest`.
- If we want to maintain compatibility, I can make new methods instead of altering existing ones.
- If not, then we could mark them as `final` and `@internal` where applicable (maybe in another PR to keep this one on topic).

stof:
Anything in the `tests` folder is entirely internal, as it is not even included in the dist archives of the composer package.

src/Parser.php
uuf6429: I tried testing it again, this time by mocking the lexer to return my own set of tokens:
1. `['type' => 'Feature'], ['type' => 'Feature'], ['type' => 'EOS']` - "Scenario, ... expected but Feature found"
2. ['type' => 'Feature'], ['type' => 'EOS'], ['type' => 'Feature'], ['type' => 'EOS'] - no error, because it stops on first `EOS`.

In short, even if the lexer misbehaves, we won't handle more than one feature because:
1. `parse()` calls `parseExpression()` until `EOS` is encountered
2. `parseExpression()` calls `parseFeature()`
3. `parseFeature()` goes into a loop until `EOS` is encountered
4. therefore `parse()` is never called again (and when it receives back control, we're in EOS so it terminates).

uuf6429:
@stof @acoulton would be great to get some direction/decisions with regards to `@codeCoverageIgnore`. I don't think that they are the same situation:
- in some cases code could be changed and then tested
- others cases it can't be changed (so I believe we should ignore coverage)
- and some others the code might actually be redundant and should be removed

acoulton (commented):
Thanks @uuf6429 and sorry I didn't get to this sooner.

I've tried to understand the ignores in Parser, I think it might be possible to solve them without excluding coverage? But I may have misunderstood. My hunch is that we don't want too much defensive code, but we probably do want something to identify and throw even if "the lexer is broken and lets such a token through" rather than that behaviour being undefined.

If we do add any `@codeCoverageIgnore` tags then I think each one should be documented inline to explain why that code isn't covered.

src/Filter/NarrativeFilter.php
uuf6429: If we keep `@return` boolean, PHPStan will not complain in the test.

To be honest, I don't really like that we redefined the result with the PHPDoc - it could screw us over in later on.

> IMO if there's a method that theoretically could be tested, but isn't, then that shouldn't be excluded from code coverage. 

Agreed.

tests/Keywords/CucumberKeywordsTest.php
uuf6429: True, I just checked - the functions (`is_file`, `is_readable`) we're using should support stream wrappers, so that should also work. That would also get rid of necessitating a random file in the system temp dir. 🤔

src/Parser.php
```diff
@@ -376,14 +380,16 @@ protected function parseScenario()
...
+            if ($steps === [] && is_string($node)) {
                 $text = preg_replace('/^\s{0,' . ($token['indent'] + 2) . '}|\s*$/', '', $node);
                 $title .= "\n" . $text;
                 continue;
             }
 
             if ($node === "\n") {
+                // @codeCoverageIgnoreStart
                 continue;
+                // @codeCoverageIgnoreEnd
```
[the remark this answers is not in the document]
uuf6429: I was not able to replicate it with a purpose built example feature text.

Through some debugging, I found out why:
- right after scanning a "Scenario: .." keyword we might end up with NewLine nodes, but they are handled by the previous if condition block (`if ($steps === [] && is_string($node)) {`).
- after a step has been parsed, subsequent newsline are caught as part of `scanStep()` because:
  1. `scanStep()` is of higher priority in `getNextToken()` (than `scanNewline()`)
  2. `scanStep()` uses a regex that captures all left-padded whitespace (`'/^\s*(' . $keywords . ')([^\s].*)/u'`), which includes newlines

On the other hand, I should be able to replicate it with a "misbehaving lexer".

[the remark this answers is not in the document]
uuf6429: A misbehaving lexer would not replicate it either - I also found why.
After parsing the first step, it "predicts" that it will only parse more steps (or otherwise quit the loop). The step parsing method (in the parser) silently consumes comment and newline nodes, so that "continue" line will neve be hit.

See here: https://github.com/uuf6429/Gherkin/blob/9dac12e38b7f252a9b8c2b4144af987bc3b0eb6b/src/Parser.php#L519

Again, `parseStep()` takes precedence over other parse methods and starts discarding newline nodes.

src/Filter/NarrativeFilter.php
acoulton: > To be honest, I don't really like that we redefined the result with the PHPDoc - it could screw us over in later on.

I think it's OK - TBH I'd probably have `isScenarioMatch() : false` if this was a `final` class (which is fine even if the parent interface is `:bool`. These classes are all defined as being filters that apply to features, so I don't see how we'd ever need them to match a scenario.

src/Parser.php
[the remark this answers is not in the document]
acoulton: In that case I think I'd add a test with a newline (if there isn't one already) to prove the behaviour, and then remove the `if ($node === "\n")` block?

src/Filter/NarrativeFilter.php
uuf6429: I went for the approach of keep bool return type (I even removed the docblocks since they seem redundant) and added some tests.

tests/Keywords/CucumberKeywordsTest.php
uuf6429: I'll keep it as is, since it works. If we see any issues, we can go through the additional effort of switching to VFS.

src/Parser.php
```diff
@@ -337,12 +339,14 @@ protected function parseBackground()
                 ));
             }
 
+            // @codeCoverageIgnoreStart
```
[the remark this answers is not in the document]
uuf6429: I went one step further and added a new exception class (extending the previous one, to also not break BC).

It feels a bit hacky, since half of the logic would not apply, but this might be the lesser evil anyway.

src/Parser.php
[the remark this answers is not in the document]
uuf6429: I don't really get your point, what's the purpose of the test?
- If it is temporary, then I don't see how it improves any trust.
- If we want to keep it in case the issue suddenly reappears, then I worry that the test would still not be able to catch it.
- If we want it kept for confidence reasons, then I worry that the next person would just see it as redundant (plus the previous point applies).

In any case, I kinda gave up on this. I removed the `codeCoverageIgnore` line, so that this PR can be merged uncontroversially. Feel free to suggestion something else though.

src/Parser.php
```diff
@@ -319,14 +319,16 @@ protected function parseBackground()
...
 
-            if (count($steps) === 0 && is_string($node)) {
+            if ($steps === [] && is_string($node)) {
                 $text = preg_replace('/^\s{0,' . ($token['indent'] + 2) . '}|\s*$/', '', $node);
                 $title .= "\n" . $text;
                 continue;
             }
 
             if ($node === "\n") {
+                // @codeCoverageIgnoreStart
```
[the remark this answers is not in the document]
uuf6429: Same as the other comment, let's keep the conversation there.

src/Parser.php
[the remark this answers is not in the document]
acoulton: It's very possible I'm misunderstanding - you have gone much more into the detail than I have.

However I think the theory of those `if ($node === "\n")` blocks is that it's valid to have extra empty lines between steps / comments etc in a Background or Scenario e.g. like this?

```gherkin
Scenario: Do something
   Given whatever

   Then this is fine
```

But it turns out that the extra newlines are already handled elsewhere in the parser, meaning those blocks are redundant - which is why it's impossible to get code coverage on them.

Therefore, it should be fine to delete that dead code.

However, we still want confidence (now and in future) that we can correctly parse input like the above. If that's not already covered by our tests (or the examples we pull in from cucumber) then we should add that before removing these "fallback" branches.

For as long as this is valid Gherkin input, the only way the test itself would become redundant would be if another test was added that also covered behaviour with extra newlines.

I think?

But also happy to leave that code in and un-covered for now if you want to move on.

src/Parser.php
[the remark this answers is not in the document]
acoulton: I think that makes it much clearer, thanks.

acoulton (approved):
@uuf6429 I'm happy with this as-is. Thanks for all your work on it!

If you'd like to spend any more time on the newlines thing then great, if not then we can merge this and someone can look at that in future if they feel like it.

src/Parser.php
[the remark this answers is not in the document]
uuf6429: Yep, that's just about it.

There seems to be at least 3 places where newslines are being handled:
1. that if condition that we're seeing
2. in the parser where it is expecting and handling steps
3. in the lexer, where the step parsing is swallowing all previous whitespace

When I found about the 3rd, I tried testing it with a mocked lexer that returns newlines (after a step token), but that didn't cover because of point 2.

I tried different combinations as well, e.g. step-newline-newline-comment-newline-newline-step.

I'll leave this out for now, it would probably be better to be its own issue/PR anyway.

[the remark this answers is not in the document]
acoulton: Fair enough, a separate PR makes sense.

stof (approved):

PR #296, opened 2025-04-12
Fix/improve bin scripts

uuf6429:
This PR fixes a few issues found by PHPStan in the `bin/` scripts. Additionally, it extends the php-cs-fixer scanning to those files as well.

### DIFF
--- a/bin/cucumber_changelog
+++ b/bin/cucumber_changelog
@@ -1,14 +1,22 @@
 #!/usr/bin/env php
 <?php
 
-if ($argc != 2) {
+/*
+ * This file is part of the Behat Gherkin Parser.
+ * (c) Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+if ($argc !== 2) {
     echo "Usage: {$argv[0]} <version>\n";
     exit(1);
 }
 
 $version = $argv[1];
 
-if (!preg_match('/[0-9]+.[0-9]+.[0-9]+/', $version)) {
+if (!preg_match('/\d+.\d+.\d+/', $version)) {
     echo "Error: Version format not recognised\n";
     exit(1);
 }
@@ -18,7 +26,7 @@ if (!$changelog = file_get_contents(__DIR__ . '/../vendor/cucumber/cucumber/gher
     exit(1);
 }
 
-if (!preg_match('/(?<changes>##\s+\['.preg_quote($version, '/').'\].*?)^##\s/ms', $changelog, $matches)) {
+if (!preg_match('/(?<changes>##\s+\[' . preg_quote($version, '/') . '\].*?)^##\s/ms', $changelog, $matches)) {
     echo "Error: Could not find version $version in changelog\n";
     exit(1);
 }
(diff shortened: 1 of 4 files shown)


bin/cucumber_changelog
```diff
@@ -28,6 +36,6 @@ if (!preg_match('/(?<changes>##\s+\['.preg_quote($version, '/').'\].*?)^##\s/ms'
 echo "Found changelog:\n$changes";
 
 if (getenv('GITHUB_ACTIONS')) {
-    $changes = strtr($changes, ["%" => "%25", "\r" => '%0D', "\n" => '%0A', ':' => '%3A', ',' => '%2C']);
+    $changes = strtr($changes, ['%' => '%25', "\r" => '%0D', "\n" => '%0A', ':' => '%3A', ',' => '%2C']);
```
uuf6429, line 39: Could it make sense to replace this with `rawurlencode()`?

acoulton: It might, I'm not actually sure what encoding is required here for github actions output (and then the input to the create-pull-request action). I need to get back to all this as these jobs aren't actually running / working anyway so I think probably leave it for now.

bin/update_cucumber
```diff
@@ -1,15 +1,23 @@
...
+
 $composerFile = __DIR__ . '/../composer.json';
-$composerConfig = file_get_contents($composerFile);
-foreach (json_decode($composerConfig, true, 512, JSON_THROW_ON_ERROR)['repositories'] as $repository)
-{
-    if ($repository['type'] !== 'package') {
+$composerConfigStr = file_get_contents($composerFile) ?: '';
+$composerConfig = (array) json_decode($composerConfigStr, true, 512, JSON_THROW_ON_ERROR);
+foreach ($composerConfig['repositories'] ?? [] as $repository) {
+    if (!isset($repository['type']) || $repository['type'] !== 'package') {
```
acoulton, line 16: Possibly slightly clearer and avoids having to keep the array keys in sync in future?
```suggestion
    if (($repository['type'] ?? null) !== 'package') {
```

bin/update_cucumber
```diff
@@ -1,15 +1,23 @@
...
         continue;
     }
 
-    if ($repository['package']['name'] == 'cucumber/cucumber') {
+    if (isset($repository['package']['name']) && $repository['package']['name'] === 'cucumber/cucumber') {
```
acoulton, line 20:
```suggestion
    if (($repository['package']['name']) ?? null) === 'cucumber/cucumber') {
```

bin/update_i18n
```diff
@@ -1,30 +1,39 @@
...
+$json = file_get_contents($gherkinFile) ?: '';
+$array = [];
 
-foreach (json_decode($json, true) as $lang => $keywords) {
-    $langMessages = array();
+foreach ((array) json_decode($json, true) as $lang => $keywords) {
+    $langMessages = [];
 
-    foreach ($keywords as $type => $words) {
+    foreach ((array) $keywords as $type => $words) {
```
acoulton, line 19: Might be safer to assert that the json decodes to an array / the keywords are an array rather than casting - will give a clearer failure if the upstream file format changes?

acoulton (changes requested):
Thanks @uuf6429 just a few suggestions

acoulton (approved):
Thanks @uuf6429, this looks good to me

carlos-granados (approved):
Thanks @uuf6429 looks good

PR #297, opened 2025-04-12
PHPStan level 9 fixes (tests only)

uuf6429:
Since the jump to PHPStan level 9 introduces a lot of changes, I decided to split that in different tasks, with the final task actually changing the PHPStan level in the config file.

This PR in particular only changes things related to tests, so that reviewing should be relatively easy and relaxed.

Note: There's only one non-test file that was changed, but it was just to add some alias types.


tests/Cache/FileCacheTest.php
```diff
@@ -73,7 +73,7 @@ public function testBrokenCacheRead(): void
             'broken_feature',
             new FeatureNode(null, null, [], null, [], '', '', null, 1),
         );
-        $files = glob($this->path . '/**/*.feature.cache');
+        $files = glob($this->path . '/**/*.feature.cache') ?: [];
```
acoulton, line 76: Might be easier to debug & more consistent with the `file_get_contents` fixes to extract the glob calls (here & ParserTest) to a method that throws if the glob returns false?

tests/FileReaderTrait.php
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Tests\Behat\Gherkin;
+
+use RuntimeException;
+
+trait FileReaderTrait
```
acoulton, line 15: I'd probably have made this a helper class with public static methods - curious about why you chose a trait here (I'm not saying it's wrong)?

src/Keywords/KeywordsInterface.php
```diff
@@ -14,6 +14,14 @@
...
  *
  * @author Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * @phpstan-type TLanguage string
+ * @phpstan-type TKeywordsArray array{feature: string, background: string, scenario: string, scenario_outline: string, examples: string, given: string, when: string, then: string, and: string, but: string}
+ * @phpstan-type TMultiLanguageKeywords array<TLanguage, TKeywordsArray>
+ * @phpstan-type TKeywordsString string
+ * @phpstan-type TStepKeywordsType 'Given'|'When'|'Then'|'And'|'But'
+ * @phpstan-type TGeneralKeywordsType 'Feature'|'Background'|'Scenario'|'Outline'|'Examples'|'Step'
+ * @phpstan-type TKeywordsType TGeneralKeywordsType|TStepKeywordsType
```
acoulton, line 24: I'm not certain what these types are for - I didn't see them referenced anywhere in the PR. I might have missed them, but I don't think so?

acoulton (commented):
Thanks @uuf6429 just a couple of comments

src/Keywords/KeywordsInterface.php
uuf6429: You're right, only the first 3 are needed.

tests/FileReaderTrait.php
uuf6429: No particular reason.. even though traits theoretically make it more manageable.

I've also wondering introducing [thecodingmachine/safe](https://github.com/thecodingmachine/safe), at least for dev/testing... wdyt?

tests/Cucumber/CompatibilityTest.php
```diff
@@ -127,7 +138,11 @@ public static function badCucumberFeatures(): iterable
      */
     private static function getCucumberFeatures(string $folder): iterable
     {
-        foreach (new FilesystemIterator(self::TESTDATA_PATH . $folder) as $file) {
+        $fileIterator = new FilesystemIterator(self::TESTDATA_PATH . $folder);
+        /**
+         * @var iterable<string, SplFileInfo> $fileIterator
```
stof, line 143: isn't phpstan smart enough to infer it ?

src/Keywords/KeywordsInterface.php
```diff
@@ -14,6 +14,14 @@
  * Keywords holder interface.
  *
  * @author Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * @phpstan-type TLanguage string
```
stof, line 18: do we really need to create such alias instead of using `string` directly ?

uuf6429: By itself it's pointless, but when used in other structures, it should be helpful.

For instance, `TMultiLanguageKeywords` could be defined as `array<string, ...>`, in which case I find it harder to convey that the key is the language.

Besides, it would make it easier to transition to finer-grained types in the future, if such a thing exists (e.g. if we're able to define the format of the string).

tests/Cucumber/CompatibilityTest.php
uuf6429: I'll double check, but I think I had to do it specifically because it couldn't.

If I may hazard a guess, maybe there might be something missing/wrong in the PHPStan stubs.

tests/FileReaderTrait.php
uuf6429: I'm also fine with a static class approach, I'll go for that.

uuf6429: I realised that we already have `symfony/filesystem` as a dev dependency, whereas `symfony/finder` is also included, but not as an explicit dependency, so might as well use both instead.

uuf6429: Damnit... `symfony/filesystem` introduced `readFile` in v7.1, which supports PHP 8.2+, but we want to support PHP 8.1 as well. 😩

tests/Cache/FileCacheTest.php
```diff
@@ -15,8 +15,11 @@
 use Behat\Gherkin\Node\FeatureNode;
 use Behat\Gherkin\Node\ScenarioNode;
 use PHPUnit\Framework\TestCase;
-use Symfony\Component\Filesystem\Filesystem;
+use Tests\Behat\Gherkin\Filesystem;
 
+/**
+ * @todo Use VFS
```
uuf6429, line 21: @acoulton I will handle this and the other case in a new PR ASAP.

tests/Keywords/CucumberKeywordsTest.php
```diff
@@ -11,23 +11,27 @@
...
         return new CucumberKeywords(__DIR__ . '/../Fixtures/i18n.yml');
     }
 
-    protected function getKeywordsArray()
+    protected static function getKeywordsArray(): array
     {
-        return Yaml::parse(file_get_contents(__DIR__ . '/../Fixtures/i18n.yml'));
+        $data = Filesystem::readFile(__DIR__ . '/../Fixtures/i18n.yml');
+
+        return Yaml::parse($data);
```
stof, line 31: you could use `Yaml::parseFile()` instead (which also mentions the file name in the error message, as it knows it)

tests/Cache/FileCacheTest.php
```diff
@@ -74,14 +71,7 @@ public function testBrokenCacheRead(): void
...
         );
-        $files = iterator_to_array(
-            (new Finder())
-                ->in($this->path)
-                ->depth(1)
-                ->name('*.feature.cache')
-                ->files(),
-            false
-        );
+        $files = Filesystem::find("$this->path/**/*.feature.cache");
```
stof, line 74: this `**` in the glob does not match the `->depth(1)` config of the Finder.

acoulton: True - but arguably for this test it is safer to assert that the previous `->write` has only created one file at any depth. Ultimately we're trying to make sure that the file we modify below is the one the cache is expecting to read.

tests/Cache/FileCacheTest.php
```diff
@@ -27,11 +30,6 @@ protected function setUp(): void
         $this->cache = new FileCache($this->path = sys_get_temp_dir() . uniqid('/gherkin-test'));
     }
 
-    protected function tearDown(): void
-    {
-        (new Filesystem())->remove($this->path);
```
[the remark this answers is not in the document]
acoulton: I assume because symfony/filesystem is removed and there's not much point in implementing our own directory removal since they become redundant with #298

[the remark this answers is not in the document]
uuf6429: Correct.

tests/Cache/FileCacheTest.php
uuf6429: I'm not sure why we're even discussing this since I introduced the finder implementation in one commit (instead of glob with **) and then changed it back (to **) in a subsequent one. 🤔

.github/workflows/build.yml
```diff
@@ -55,9 +55,9 @@ jobs:
           ini-file: "development"
           coverage: none
 
-      - name: Update Symfony version
+      - name: Update symfony/yaml version
         if: matrix.symfony-version != ''
-        run: composer require --no-update "symfony/symfony:${{ matrix.symfony-version }}"
+        run: composer require --dev "symfony/yaml:${{ matrix.symfony-version }}"
```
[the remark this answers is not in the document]
uuf6429: Ah, if I understand right, we just want to change the composer.json file in this step and leave the installation (and update) to the next step instead. Fine, makes sense to me.

I'm not sure how `symfony/flex` would make any difference though.. I think I'd rather keep it standard - `symfony/yaml` could have been any other random package we wanted to test with (that's also why I removed the original `symfony/symfony`dependency).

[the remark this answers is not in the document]
acoulton: To be honest, since (I think? Certainly as far as the composer.json indicates)  we are only using the Symfony components in our test suite I'm not actually sure we need to support/build with multiple versions? Can't we just pick one that works with our php versions and use that?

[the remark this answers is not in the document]
uuf6429: @acoulton `symfony/yaml` is an optional dependency (check the "suggest" in `composer.json`) to parse features from yaml file.

In addition to unrelated tests, the pipeline is testing each major version of that component.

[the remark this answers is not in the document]
acoulton: Ah, I missed that. In that case yes I agree, better to just vary the symfony/yaml direct in this instance. symfony/flex would only be useful if we had multiple Symfony components that need to switch versions IMO.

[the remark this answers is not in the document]
stof: The drawback of this approach is that it replaces entirely the requirement in our composer.json, which would ignore it.

Thinking about it, this could be also solved by using the `--with` flag in the `composer update` command to apply an additional temporary constraint in the solver (which composer will validate against our normal requirement instead of ignoring the normal requirement entirely)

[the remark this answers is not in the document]
uuf6429: @stof now I see what you mean. In that case we can simplify the pipeline further: https://github.com/Behat/Gherkin/pull/297/commits/238553a236e99ecc942aa5bb09047d867d1b0316
I also took the liberty to make it track and submit coverage, which seems to work well:
https://github.com/Behat/Gherkin/actions/runs/14539416015/job/40794264765?pr=297

.github/workflows/build.yml
```diff
@@ -39,28 +39,25 @@ jobs:
...
+            composer-flags: "--with=symfony/yaml:^5.4"
           - php: 8.1
-            symfony-version: '^6.4'
+            composer-flags: "--with=symfony/yaml:^6.4"
     steps:
       - uses: actions/checkout@v4
-
-      - name: Setup PHP
-        uses: shivammathur/setup-php@v2
+      - uses: shivammathur/setup-php@v2
```
acoulton, line 49: I would prefer to keep the blank line between each step, I think that's more readable.

.github/workflows/build.yml
```diff
@@ -39,28 +39,25 @@ jobs:
...
         php: [ 8.1, 8.2, 8.3, 8.4 ]
         composer-flags: [ "" ]
-        symfony-version: [ "" ]
         include:
           - php: 8.1
-            symfony-version: '^5.4'
+            composer-flags: "--with=symfony/yaml:^5.4"
           - php: 8.1
-            symfony-version: '^6.4'
+            composer-flags: "--with=symfony/yaml:^6.4"
```
acoulton, line 46: Good solution, thanks

.github/workflows/build.yml
```diff
@@ -39,28 +39,25 @@ jobs:
...
-        run: composer update ${{ matrix.composer-flags }}
-
-      - name: Run tests (phpunit)
-        run: ./vendor/bin/phpunit
+          coverage: "xdebug"
+      - run: composer update ${{ matrix.composer-flags }}
+      - run: ./vendor/bin/phpunit --coverage-clover=coverage.xml --log-junit junit.xml --colors=always --testdox
+      - uses: codecov/codecov-action@v4
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
```
acoulton, line 58: There is no `CODECOV_TOKEN` configured in this repository. The build output says `Fork detected, tokenless uploading used` so I assume it's ignoring this parameter at the moment but I guess it may not work once it merges to master? That secret also wouldn't be available in PR builds.

I'm also not clear on whether / how we need to set up codecov for ourselves - presumably someone needs to create an "official" behat profile on codecov to be able to control which repos can upload data etc? From a quick look at the codecov actions it looks like they support OIDC (e.g. GITHUB_TOKEN) permissions which IMO would be preferable to a repository secret. However since that means granting `id_token:write` to the github action, we should really split the codecov upload into a separate job (using artifacts from the main build).

It might be better to figure this out in a separate PR so we can get this merged?

acoulton (changes requested):

.github/workflows/build.yml
uuf6429: I've copied that part from existing projects where it already works, and seeing as it did work (notice the `https://app.codecov.io/gh/Behat` URL), I think it should be fine to merge this as is (especially since I wouldn't change anything in the way it works).

Of course if it gets any more complicated, it would make sense to split it.. but so far I don't see any reason to.

.github/workflows/build.yml
uuf6429: I mainly did that to be consistent with the other jobs.

However, I find it less readable with the blank lines.. this is essentially a build script, so it's kinda weird to have each line in a script separate by a blank line. I would rather leave that to separate jobs.

Additionally, I also remove the step names since they were mostly repeating what the step did almost verbatim.

acoulton: No issue with removing the step names.

The difference with a build script is you would usually indent arguments to a command that spanned multiple lines. With an actions yaml file they are on the same indentation so the only marker for the start of a new step is the `-` at the start of the line. Not a huge issue if things are just a one-line `run` but as soon as you need to introduce `if`, `env`, etc then it becomes harder to see imo.

.github/workflows/build.yml
acoulton: So codecov allow completely unauthenticated upload?

If so and the token is not actually required then it would be better to remove the `token: ${{ secrets.CODECOV_TOKEN }}` lines to avoid confusion. Not least then if it stops working in future it will be clearer that something on codecov has changed rather than trying to understand why the secret (which never existed) is undefined.

Also I noticed on https://app.codecov.io/gh/Behat/Gherkin/commit/238553a236e99ecc942aa5bb09047d867d1b0316 it reports "6 flag errors" - all appear to be "Multiple flags detected. Please ensure one flag per upload." Not sure what that's about, do we need to do more to identify/separate the different coverage reports from each actions job?

uuf6429: I think I'd still keep it like that to encourage a 'good' habit in the sense that readers should expect the dash to indicate a new command, rather than a blank line which could mean anything (keep in mind that a long string could be split into multiple lines, including blank lines).
```yaml
- run: |
    this is the

    same command
```
Although my main point would be that conciseness in terms of whitespace tends to be more understandable and show the bigger picture.

.github/workflows/build.yml
uuf6429: I think I'd still keep it like that to encourage a 'good' habit in the sense that readers should expect the dash to indicate a new command, rather than a blank line which could mean anything (keep in mind that a long string could be split into multiple lines, including blank lines).
```yaml
- run: |
    this is the

    same command
```
Although my main point would be that conciseness in terms of whitespace tends to be more understandable and show the bigger picture.

uuf6429:
I've removed pipeline/workflow changes so we can get this PR merged ASAP.

acoulton (approved):
Thanks @uuf6429

uuf6429:
@carlos-granados @stof could one of you please give this PR a quick review?

It's kinda blocking the rest of the upcoming changes.

stof (approved):

uuf6429:
Thanks!

PR #298, opened 2025-04-14
Use vfs in tests

uuf6429:
As the title says. :)

This branch is based on `uuf6429:chore/phpstan-level-9-tests`, so please check the last commit only.


tests/Filesystem.php
```diff
@@ -25,8 +25,19 @@ public static function readFile(string $fileName): string
     /**
      * @return list<string>
      */
-    public static function find(string $pattern): array
+    public static function findRecursively(string $path, string $pattern): array
     {
-        return glob($pattern) ?: [];
```
uuf6429: `glob()` doesn't work with streams, so I took the next best approach: iterators + fnmatch.

tests/Filesystem.php
```diff
@@ -25,8 +25,20 @@ public static function readFile(string $fileName): string
...
+        /**
+         * @var iterable<string, \SplFileInfo> $fileIterator
+         */
+        $fileIterator = new \RecursiveIteratorIterator(new \RecursiveDirectoryIterator($path), \RecursiveIteratorIterator::CHILD_FIRST);
```
acoulton, line 33: I think these should be imported rather than `\FQCN`?

tests/Filesystem.php
```diff
@@ -25,8 +25,20 @@ public static function readFile(string $fileName): string
...
+
+        $found = [];
+        foreach ($fileIterator as $file) {
+            if (fnmatch($pattern, $file->getFilename())) {
```
acoulton, line 37: Obviously `glob` wasn't strict, but I think we're generally expecting this to only return files rather than directories. So it might be worth being explicit / concrete about that and checking `->isFile()` as well?

acoulton (commented):
LGTM, just a couple of small comments

tests/Filesystem.php
stof: the Finder-based implementation (replaced with `glob` in #297) was strict.

stof: An easy way to improve that is to add `\RecursiveIteratorIterator::LEAVES_ONLY` in the flags

acoulton: Good point - but this replaces a mix of different approaches to finding files (in the tests), others of which were just using `glob` and assuming there would only be files in the result.

tests/Filesystem.php
uuf6429: A bit annoying we haven't set that up in php-cs-fixer... I for one keep remembering which one it should be.

tests/Filesystem.php
uuf6429: Both usages of this function seem to deal mainly with files, so I think I'll just rename the function and go for `isFile()`.

acoulton (approved):
I've only checked the vfsStream commits in this branch, looks good thanks @uuf6429

acoulton:
@uuf6429 it'd be great to have this, I think it also just needs a rebase now? Cheers

uuf6429:
@acoulton I just had to sort out a conflict; it should be fine now.

acoulton (approved):
Great, thanks 😀

acoulton:
@carlos-granados that's a good point.

This one branched off from a discussion a while ago on https://github.com/Behat/Gherkin/pull/295#discussion_r1976443793 so I had that context - but you're absolutely right that as a standalone PR / to someone outside the repository it would be very helpful to have a more detailed commit message / PR description with the "why" as well as the "what".

uuf6429:
@carlos-granados you're totally right - in this case it's as @acoulton said. I think that what is happening is that I jump too quickly into such changes and then forget about adding a description. I must admit that I keep PR tabs open, but completely scrolled to the bottom. 😅 

I'll try to keep this in mind, but please also feel free to point out these situations - I'd gladly add more context, even for the sake of future readers of these PRs.

acoulton:
@uuf6429 thanks - I think given there's a few iterative commits/merges here it might be worth you just rebasing and squashing that all together with a new/clear commit message?

acoulton:
We did discuss it, I think it depends.

Commits like this I agree should ideally be squashed (and I don't always remember to do so, which I should).

However, if commits are well structured & intentional then I think they should be merged as-is, as there is value in being able to see that history in future blame/log/etc. It's not always ideal to have to go to a GitHub PR to understand the components of a group of changes (and there's no guarantee GitHub will be here/have them forever - having them in the git log avoids us being platform dependent).

IMO ideally squashing/tidying commits and their messages would be part of preparing the branch for review/merge, rather than something for maintainers to do.

uuf6429:
> However, if commits are well structured & intentional then I think they should be merged as-is, as there is value in being able to see that history in future blame/log/etc.

It feels like such cases should be separate branches/PRs then. At most, maybe consider having the non-squashing approach for "epic" branches - but in most cases it feels like it should be 1 branch -> 1 objective change.

> It's not always ideal to have to go to a GitHub PR to understand the components of a group of changes

GitHub (or git maybe?) merges the different commit messages into the squashed commit description, so there is still some history, somewhat. But I wouldn't rely on it.

On a related note, I have trouble reading the commit history even with well-written messages with merges because the commits are interspersed with other merges.

uuf6429:
@carlos-granados can you please check this PR again?

acoulton:
@uuf6429 sorry I didn't get a chance to respond to your comments.

> It feels like such cases should be separate branches/PRs then. At most, maybe consider having the non-squashing approach for "epic" branches - but in most cases it feels like it should be 1 branch -> 1 objective change.

I disagree. Unrelated things should obviously be separate branches, but often there are changes that merit being described individually but that also deserve the context of being on a branch.

> >It's not always ideal to have to go to a GitHub PR to understand the components of a group of changes

> GitHub (or git maybe?) merges the different commit messages into the squashed commit description, so there is still some history, somewhat. But I wouldn't rely on it.

It doesn't correlate the messages with which parts of the diff they relate to though, which makes them much less useful.

It also puts the job of deciding which bits of the commit message are important to keep / rewrite / take from the PR description onto the maintainers instead of the original author. For example, here is the current default "Squash & Merge" message for this PR (even after you updated the description).

```
Use vfs in tests (#298)

* Use VFS in tests

* Minor CR fixes

* Remove unnecessary teardown

* Remove redundant import
```

That is much less useful to the end user or the project history than if you had rebased this to a single commit, and put what is now in the PR description as the commit message. If you had done that, we could just do a (non-squash) merge and everyone would be happy.

Equally, if you had got multiple commits with longer / clearer messages then to squash merge we would have to decide if/how to edit those down to a single message. Maybe for example the most important parts of the PR came after some small enabling commits, so if we're squashing that to a single message we might need to change the order so people can easily see the big thing. Again that isn't an issue in a normal merge.

> On a related note, I have trouble reading the commit history (even with well-written messages) of merged PRs because the commits are interspersed with other merges.

That is a valid point. We should probably more often be rebasing branches as part of / before merging - both so that we know they're green before merge, and so that if the merge has multiple commits they all sit linear in the history.

uuf6429:
> I disagree. Unrelated things should obviously be separate branches, but often there are changes that merit being described individually but that also deserve the context of being on a branch.

But the branch is gone in either case - the only leftover is the platform-specific PR. I still believe that each individual commit should be its own branch. That would also help if you're using `git bisect` because it somewhat ensures or at least encourages that each commit is in working order.

> It doesn't correlate the messages with which parts of the diff they relate to though, which makes them much less useful.

I know, but as said, if the commits are intended to be individual, either avoid squashing or split into different branches. Requiring contributors' commits and messages to be excessively granular is not realistic and feels like another reason to slow things down for no real benefit.

> It also puts the job of deciding which bits of the commit message are important to keep / rewrite / take from the PR description onto the maintainers instead of the original author. For example, here is the current default "Squash & Merge" message for this PR (even after you updated the description).

GitHub also provides various options for the default behaviour, by the way. From my perspective, I don't really care as much about the commit description.

> We should probably more often be rebasing branches as part of / before merging

You can configure GitHub to disable the different types of merging, to force a specific way.

Additionally, I find the argument to avoid or replace (convenient) platform(github)-specific functionality somewhat counterproductive. I would rather have the PR number/link in one commit message than multiple commits with the perfect message since it's easier to follow through things that can't be recorded in git (e.g. review conversations, correlate the PR/commit to the release etc). It's also far from practical to think about replacing GitHub (especially considering workflows).

PS: I just checked a few contributed PRs in this and the Behat/Behat repo and I didn't really find them useful e.g. I don't see it useful to have an "Added a test for missing scenario" commit message in the main history. I do believe that those messages will invariably lack context.

In any case I'll just squash this PR, but I still find that a waste of time (compared to what could be a 1-click convention).

carlos-granados (approved):
Thanks @uuf6429, your explanation helped me understand what was going on here

acoulton:
@uuf6429 I think there's a bit we'll have to just agree to disagree on. Couple of small things that are worth mentioning:

* git branches are still (permanently) visible in the git log after they are merged - the only thing that's lost is the branch name.
* we might not plan to ever move off GitHub, but services/platforms change all the time and you never know what changes you might be forced to make in future. Also, the simple PR and issue links won't always necessarily work outside the current repo (on a hard fork, for example) so there are definite benefits to having the basic history in the git log rather than externally.

That said, I wasn't aware GitHub have now added the ability to customise the default squash commit message. I'd be happy to set that to the PR title & description, and then to squash merge by default in future unless the maintainers think there's value in the individual commits.

carlos-granados:
@acoulton I think that squash and merge by default is OK as long as the option to do a normal merge is still available for those times when we might want to keep the commit history

acoulton:
@carlos-granados I've changed the default squash merge message here and Behat/Behat to pull request title and description.

Turns out there's no way to change the default merge method (short of disabling all the other options) - it just follows what you did last time :disappointed:  So we'll just have to try to remember...

PR #299, opened 2025-04-19
Improve GH workflow

### LINKED ISSUE #283: Update GH workflow
This is mostly a note for myself for later, but everyone is welcome contribute. :)

- enable coverage (xdebug)
- send coverage to codecov
- use codecov labels
- improve build name/title display (`PHP 8.1, sf ^6.4` instead of `Build and test (8.1, ^6.4)`)
- remove `composer-flags` since it seems to be unused
- enable colors where relevant (e.g. composer install, phpunit output etc)
- consider renaming `build.yml` to `ci.yml` (minor/nitpick)

uuf6429:
Fixes #283.

Someone with access should follow the comment below, to set up codecov for the Behat org.


.github/workflows/build.yml
```diff
@@ -39,28 +39,25 @@ jobs:
...
         include:
           - php: 8.1
-            symfony-version: '^5.4'
+            composer-flags: "--with=symfony/yaml:^5.4"
           - php: 8.1
-            symfony-version: '^6.4'
+            composer-flags: "--with=symfony/yaml:^6.4"
     steps:
       - uses: actions/checkout@v4
-
```
acoulton, line 50: @uuf6429 and I have different opinions on whether the actions yaml is more readable with (my preference) or without (his) a blank line between each step as [previously discussed here](https://github.com/Behat/Gherkin/pull/297#discussion_r2051428957). I'm happy to go with what other maintainers decide.

.github/workflows/build.yml
```diff
@@ -39,28 +39,25 @@ jobs:
...
-        run: composer update ${{ matrix.composer-flags }}
-
-      - name: Run tests (phpunit)
-        run: ./vendor/bin/phpunit
+          coverage: "xdebug"
+      - run: composer update ${{ matrix.composer-flags }} --ansi --no-progress
+      - run: composer test -- --coverage-clover=coverage.xml --log-junit junit.xml --testdox
+      - uses: codecov/codecov-action@v4
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
```
acoulton, line 58: @uuf6429 I have authorised the codecov app (for Behat/Gherkin only at this point, we can add it to other repositories as and when required).

I'm still not clear why the workflow specifies a `secrets.CODECOV_TOKEN` that is not defined (and therefore empty) and whether that will continue to work once this PR is merged. I'm somewhat surprised if Codecov allow anyone to upload coverage reports unauthenticated, but maybe that's the case. Can you provide any more info on how this works? If we don't in fact require a secret then we should remove these `token` arguments from the yaml to avoid confusion.

acoulton (changes requested):

.github/workflows/build.yml
uuf6429: As I understand, there were 3 options:
1. per-repository secret
2. per-organisation secret
3. github app/plugin

so I'm guessing that the secret is no longer needed with the 3rd option. I'll remove it and we can see what happens. 🤷

uuf6429: Seems to still be working fine. My guess is that:
- codecov already knows about this project (it even lists [other behat org projects](https://app.codecov.io/gh/Behat))
- codecov understands that this is a PR run (it even shows that in the branch dropdown, prefixing my branches with my username)

uuf6429: I guess easiest way to see this is to get it merged (after the other point(s) are also considered).

@carlos-granados can you also give this PR a look please?

.github/workflows/build.yml
carlos-granados: I think that having blank lines between steps in a job is much more readable

carlos-granados (commented):
Other than my comment about blank lines everything else looks good

.github/workflows/build.yml
uuf6429: OK, I'll go for it and update the others as well.

acoulton (approved):
Thanks @uuf6429

carlos-granados (approved):
Thanks @uuf6429

acoulton:
@uuf6429 well, the [codecov steps in master are passing](https://github.com/Behat/Gherkin/actions/runs/14650820375/job/41115951396) (in that they're exiting 0) but I don't think they're actually working, the logs for both say:

`error - 2025-04-24 20:15:32,297 -- Upload failed: {"message":"Token required because branch is protected"}`

There's also an error  `Could not pull latest version information: SyntaxError: Unexpected token '<', "<!DOCTYPE "... is not valid JSON` which looks like their action is trying to hit an API but getting an HTML page? Not sure if that's related.

Are you able to look into what needs to be done to get this working?

uuf6429:
@acoulton how about we add back the token? In the past that's what [I've been using](https://github.com/minkphp/webdriver-classic-driver/blob/main/.github/workflows/ci.yml#L140).

uuf6429:
Although interestingly, there are actually results for the master branch:
https://app.codecov.io/gh/Behat/Gherkin

Strange.

acoulton:
@uuf6429 but like I've been saying, there is no `secrets.CODECOV_TOKEN` in this repo - if we added that line back to the github yaml it'll just pass an empty string. I can't see how that will help. 

What I'm asking is, how do we get a CODECOV_TOKEN - or, preferably, is there a way to authenticate with Codecov using [github actions OIDC authentication](https://docs.github.com/en/actions/security-for-github-actions/security-hardening-your-deployments/about-security-hardening-with-openid-connect) to eliminate the downsides of providing long-lived secrets to actions jobs?

Issue #260, opened 2024-08-30
"Automatic Cucumber tag update" not triggered for 2 years


arnapou:
Context
--------

- I was looking at the `composer.json` file and saw a declared repository [cucumber/cucumber](https://github.com/cucumber/cucumber)
- Because curiosity, I followed the link ... which redirected me to [cucumber/common](https://github.com/cucumber/common) 🤔 
- Thus, I dove into the github actions `.github/workflows/update.yaml` + `bin/update_cucumber`.
Apparently it creates `Automatic Cucumber tag update to <TAG>`.

Problem
--------

When looking at the real [cucumber/gherkin](https://github.com/cucumber/gherkin) repo, I found that the last tag is **29.0** from 3 weeks ago.

But the `composer.json` is using the **24.0** from 2 years ago 😨.

There is probably something to watch & update there, isn't it ?

stof:
the workflow ran recently: https://github.com/Behat/Gherkin/actions/workflows/update.yml

However, it detected 24.1.0 as latest version apparently. The script probably need to be updated to account for the fact that cucumber has split gherkin from their previous monorepo again.

arnapou:
I have not the bandwidth to help (PR) etc ... at the moment.
At least, I wanted to raise the question.

Sorry for not being able to help more at the moment 🤗

acoulton:
I've started to work on this, also looking at #227 in parallel.

acoulton:
This is now working again :)

Issue #227, opened 2021-02-17
Cucumber update PRs don't trigger CI


ciaranmcnulty:
See comment on #222

acoulton:
@stof yes, that's my plan when I can get back to this.

One challenge is I don't have access to create / manage a github app for the Behat org, or to configure secrets for github actions. Do you?

The other is that we may need a separate repo to send the PRs from (e.g. Behat/Gherkin-automation-prs) because currently the branch protection rules cover everything so we can't create or delete arbitrary branches in this repo.

stof:
We would need to work with @ciaranmcnulty or @everzet so that they grant admin access on the organization to active maintainers instead of only write access.

everzet:
Sorry for being absent - had mad couple of weeks. @acoulton happy to make you org admin if you're up for it?

acoulton:
@everzet no need to apologise. Hope you're getting some time to chill over the festive season. That'd be great if you can make me an admin, thanks.

everzet:
@acoulton done

acoulton:
@everzet thanks very much for that - I'm just back at work today so will take a look at getting the update flow working again when I can.

acoulton:
This is now resolved, see #306

PR #300, opened 2025-04-24
fix: Reinstate & improve automatic updates of Gherkin tests & translations

acoulton:
Relates to #260 and #227 

I have rewritten the `update_cucumber` script and the github actions workflow for this, as detailed in the commit messages.

I've tested the script itself locally all the way up to the current cucumber/gherkin release and it is working as expected and producing the right composer patches and commit messages.

I've tested as much of the github actions flow as I feasibly can in an external repo. There's obviously the chance there are issues in the wiring up (especially of the custom github app / token permissions etc) but for what this is I think it's easier if we merge it and troubleshoot anything that comes up once it's running.


acoulton:
Also to note that both the actions scripts I've added (which I couldn't find direct solutions for on Marketplace) are things we actually need at work. Once we see them working, I'll probably look at extracting them to standalone actions, rather than permanently having them as scripts in this repo.

.github/workflows/update.yml
```diff
@@ -11,42 +11,88 @@ jobs:
...
-          php-version: 8.1
+          php-version: 8.4
           coverage: none
 
-      - uses: actions/checkout@v4
+      - name: Check whether to update or create PR
+        id: check-existing-pr
+        uses: actions/github-script@v7
+        with:
+          script: |
```
acoulton, line 28: I realise this is quite hard to follow in unified diff format. It's much easier to view in split diff.

This needs to run *before* the checkout, because it controls which branch we pull - hence why it is inline in the action.yml rather than a separate script. I envisage this eventually becoming a standalone action.

.github/workflows/update.yml
```diff
@@ -11,42 +11,88 @@ jobs:
...
-        run: bin/cucumber_changelog ${{ steps.cucumber.outputs.cucumber_version }}
-        if: steps.cucumber.outputs.cucumber_version
+            if (pr) {
+              core.info(`Found existing PR ${pr.number} for ${pr.head.ref}`)
+              core.setOutput('exists', 'true')
+              core.setOutput('checkout_branch', pr.head.ref)
+              core.setOutput('commit_branch', pr.head.ref)
+            } else {
+              const timestamp = Math.floor(Date.now() / 1000)
+              const newBranch = `do-update-${timestamp}`
```
acoulton, line 63: This is meant to be using the `branchPrefix` variable, apologies - I'll fix that.

.github/workflows/update.yml
```diff
@@ -11,42 +11,88 @@ jobs:
...
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              branchPrefix
+            })
 
-      - name: Find changelog
-        id: changelog
-        run: bin/cucumber_changelog ${{ steps.cucumber.outputs.cucumber_version }}
-        if: steps.cucumber.outputs.cucumber_version
+            if (pr) {
```
acoulton, line 56: We first check to see if there's an existing `cucumber-update-*` branch in this repo with an open PR. If there is, that will be checked out and the next cucumber update (if any) applied to the same branch.

If not, we create a new branch based off master. We use a unique suffix on the branch name to avoid any conflict with an old update (if e.g. a previous PR was merged/closed but the branch was not deleted).

.github/workflows/update.yml
```diff
@@ -11,42 +11,88 @@ jobs:
...
+        if: steps.cucumber.outputs.cucumber_updated == 'yes'
+
+      - uses: actions/create-github-app-token@v2
+        id: app-token
+        if: steps.cucumber.outputs.cucumber_updated == 'yes'
+        with:
+          app-id: ${{ vars.GHERKIN_UPDATER_APP_ID }}
+          private-key: ${{ secrets.GHERKIN_UPDATER_APP_KEY }}
+
+      - run: .github/commit-and-push-gherkin-update.sh
```
acoulton, line 91: I don't want to use the current `peter-evans/create-pull-request` action because that is explicitly designed to always create a single new commit off the main branch and force push it. If there are multiple gherkin releases before we get round to merging the branch, I would much prefer to see these as separate updates each with their own build results etc.

There are other "auto commit and push" actions, but they all have limitations / design decisions that made them not quite right for this. For example, they often use the github token provided in the initial `actions/checkout` step - but in our case it's cleaner and more secure to only create the custom app token right at the end, if we know we have something to push.

.github/workflows/update.yml
```diff
@@ -11,42 +11,88 @@ jobs:
...
+        if: steps.cucumber.outputs.cucumber_updated == 'yes'
+        env:
+          GH_TOKEN: ${{ steps.app-token.outputs.token }}
+          BRANCH_NAME: ${{ steps.check-existing-pr.commit_branch }}
+          COMMIT_MSG: ${{ steps.cucumber.outputs.commit_msg }}
+          PR_EXISTS: ${{ steps.check-existing-pr.outputs.exists }}
+          PR_TITLE: Update to cucumber/gherkin ${{ steps.cucumber.outputs.cucumber_version }}
```
acoulton, line 98: The PR title will always be updated to the latest Gherkin release it includes

.github/commit-and-push-gherkin-update.sh
```diff
@@ -0,0 +1,28 @@
...
+git add -u
+git -c user.name='github-actions[bot]' \
+    -c user.email='41898282+github-actions[bot]@users.noreply.github.com' \
+    commit \
+    -m "$COMMIT_MSG"
+
+echo "Authenticating and pushing"
+# This will reconfigure git to push with the token provided in the GH_TOKEN env var
+gh auth setup-git
+git push -u origin "$BRANCH_NAME"
```
acoulton, line 20: Design decision: the update branches will be in the main `Behat/Gherkin` repository (as they were before). We'll need to relax the branch protection rules to allow us to delete `cucumber-update-*` branches after they're merged. This does mean they'll appear in packagist.

Alternatively we could create a `Behat/Gherkin-Automation-PRS` repo as a fork and configure our github app with the permissions to push the branch there instead. However that complicates things a bit and I think it's simpler to keep everything together?

carlos-granados (approved):
Thanks @acoulton, great job.  As you said I think that the best course of action will be to merge this, test it and fix it if any issues are found

PR #301, opened 2025-04-24
Add back codecov token

uuf6429:
Follow up of https://github.com/Behat/Gherkin/pull/299

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -64,7 +64,10 @@ jobs:
 
       - uses: codecov/codecov-action@v4
         with:
+          token: ${{ secrets.CODECOV_TOKEN }}
           files: coverage.xml
           flags: php${{ matrix.php }}${{ matrix.composer-flags }}
 
       - uses: codecov/test-results-action@v1
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}


acoulton:
Moving my comment from https://github.com/Behat/Gherkin/pull/299#issuecomment-2828821981#

@uuf6429 but like I've been saying, there is no `secrets.CODECOV_TOKEN` in this repo - if we add this line back to the github yaml it'll just pass an empty string. I can't see how that will help. 

What I'm asking is, how do we get a CODECOV_TOKEN - or, preferably, is there a way to authenticate with Codecov using [github actions OIDC authentication](https://docs.github.com/en/actions/security-for-github-actions/security-hardening-your-deployments/about-security-hardening-with-openid-connect) to eliminate the downsides of providing long-lived secrets to actions jobs?

acoulton:
> Although interestingly, there are actually results for the master branch:
https://app.codecov.io/gh/Behat/Gherkin
> Strange.

Very strange indeed - the build logs from their action clearly say that uploading the report failed, yet they seem to have accepted and published it anyway at their end. Feels buggy to me, but 🤷 

Maybe if it's actually working (and most importantly, not causing the build steps to fail) we see what happens over the next little while. Unless you'd like to spend the time trying to figure it out.

uuf6429:
> What I'm asking is, how do we get a CODECOV_TOKEN

I got that from codecov dashboard in the past. I don't know about GH OIDC to answer that.

In my account I can see the following:
{89793502-3D6A-43A8-A95F-E06521AB6A24}

If the same thing has been set for the behat org, then I guess there's nothing more to do.. but I agree that the console message is misleading at best.

uuf6429:
Might be related: https://github.com/codecov/codecov-action/issues/1709

Can you verify that setting in the behat org so that we know it's not allowing uploads by implicit association?

uuf6429:
@acoulton when @stoff merged in my phpstan-test-fixes PR to master, it seems that codecov didn't pick up those changes (even though codecov did run in the workflow - and fail with the same error as before):
https://app.codecov.io/gh/Behat/Gherkin/commits

This is leading me to believe that maybe the behaviour we saw was due to users already being authorized from codecov's perspective.

Maybe check if that setting is enabled for the behat org?

acoulton (approved):

carlos-granados (approved):

acoulton:
@uuf6429 that seems to be working now as far as I can tell

PR #302, opened 2025-04-26
fix: Ensure assertions in the update_cucumber script are executed

acoulton:
By default shivammathur/setup-php configures PHP with a production ini file, which disables zend assertions.

This means that calls like `assert(preg_match(...,..., $matches))` are compiled out, leaving $matches undefined. The logic in the update_cucumber script requires the assertions to run, so we use the development ini file and have the script fail if they're not enabled.

### DIFF
--- a/bin/update_cucumber
+++ b/bin/update_cucumber
@@ -12,6 +12,12 @@
 ini_set('error_reporting', E_ALL);
 set_error_handler(static fn ($code, $msg, $file, $line) => throw new ErrorException($msg, $code, $code, $file, $line));
 
+if (ini_get('zend.assertions') !== '1') {
+    // Ensure our assertions are actually processed. Otherwise any code inside the assert() call is compiled out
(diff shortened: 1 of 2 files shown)


carlos-granados (approved):
Thanks @acoulton

bin/update_cucumber
```diff
@@ -12,6 +12,12 @@
 ini_set('error_reporting', E_ALL);
 set_error_handler(static fn ($code, $msg, $file, $line) => throw new ErrorException($msg, $code, $code, $file, $line));
 
+if (ini_get('zend.assertions') !== '1') {
+    // Ensure our assertions are actually processed. Otherwise any code inside the assert() call is compiled out
+    echo "This script must be run with zend.assertions set to 1\n";
+    exit;
```
stof, line 18: this should exit with a failure exit code

acoulton: 🤦‍♂️indeed it should. Will fix, thanks

PR #304, opened 2025-04-30
ci: Fix workflow definition for gherkin update and add actionlint

acoulton:
Annoyingly, I had another typo / oversight in the actions file (incorrect references to the outputs of a step) which caused the run to fail this morning https://github.com/Behat/Gherkin/actions/runs/14749065681/job/41402076670

This PR fixes that, and also introduces a new action to run [actionlint](https://github.com/rhysd/actionlint) on all our workflow files if they're changed. That would have caught this mistake, as proved in [this build on my fork](https://github.com/acoulton/Gherkin/actions/runs/14750347837).

Additionally, I've added a manual workflow_dispatch trigger to the gherkin update to make it easier for us to troubleshoot this action in future without having to wait for the daily cron.


acoulton:
@carlos-granados apologies for another iteration on this - I *think* this will work now. Certainly the action looks like it did what it was meant to this morning up until it failed on the empty value at the final step.

carlos-granados (approved):
No worries @acoulton

PR #307, opened 2025-04-30
ci: Update scripts for phpstan level 9

uuf6429:


bin/update_cucumber
```diff
@@ -20,8 +20,8 @@ if (ini_get('zend.assertions') !== '1') {
 
 $updater = new
 /**
- * @phpstan-type CurrentVersionArray array{tag_name: string, hash: string, composer_version: string}
- * @phpstan-type GitHubReleaseArray array{tag_name: string, hash: string, html_url: string, changelog: string}
+ * @phpstan-type TCurrentVersionArray array{tag_name: string, hash: string, composer_version: string}
+ * @phpstan-type TGitHubReleaseArray array{tag_name: string, hash: string, html_url: string, changelog: string}
```
uuf6429, line 24: @acoulton I decided to keep the same convention of prefixing the alias types with a 'T' in some languages this convention is used to denote 'types' (as opposed to classes). I've been doing that to avoid conflicts with real PHP classes (although the 'Array' suffix would have adequately solved that in this case - with added readability).

bin/update_cucumber
```diff
@@ -77,13 +80,15 @@ class {
             && ($repository['package']['name'] ?? null) === 'cucumber/gherkin-monorepo'
             && isset($repository['package']['version'])
             && is_array($repository['package']['source'] ?? null)
-            && isset($repository['package']['source']['reference']),
+            && isset($repository['package']['source']['reference'])
+            && is_string($repository['package']['source']['reference']),
```
uuf6429, line 84: Without the string check the returned array structure could theoretically contain `mixed` type, whereas we want specifically a string. A few other similar cases have been tackled below.

bin/update_cucumber
```diff
@@ -109,27 +117,44 @@ class {
     }
 
     /**
-     * @return list<GitHubReleaseArray>
+     * @phpstan-return non-empty-list<TGitHubReleaseArray>
```
uuf6429, line 120: I noticed that later on we seem to assume the there must be at least one revision (the line with `return array_pop($releases);` would return `null` instead of `array|false`).

Of course if it is expected to have 0 releases, then maybe instead of the assertion and non-empty-list, the return line should be `return array_pop($releases) ?? false;` (or throw an exception). @acoulton what are your thoughts?

bin/update_cucumber
```diff
@@ -109,27 +117,44 @@ class {
...
         // GitHub requires API requests to send a user_agent header for tracing
         ini_set('user_agent', 'https://github.com/Behat/Gherkin updater');
         $releases = $this->jsonDecodeArray($this->getContents("https://api.github.com/repos/$repo/releases"));
 
+        assert($releases !== []);
+
         // Releases API is in order of github release ID, this may not be version order (e.g. major releases may be
         // interspersed with point releases to older versions). So sort it most recent -> oldest by version number.
-        usort($releases, static fn (array $a, array $b) => version_compare($b['tag_name'], $a['tag_name']));
+        usort($releases, static function (mixed $a, mixed $b) {
```
uuf6429, line 132: At this point, PHPStan still sees `$release` as an "array of stuff". If we want to avoid this excessive type checking within sorting, maybe we should move sorting after the array is formally defined (e.g. instead if directly returning the next `array_map`, we put it in a variable, sort and then return it).

I'm fine with either approach.

bin/update_cucumber
acoulton: I guess... I suppose it depends how much detail we want to go into for a CI script. Either it's going to fail on an assertion or it's going to blow up / produce an unexpected or invalid PR diff (which we'll see before merging) - not sure that it really matters which as ultimately the job will fail and need to be debugged?

Bearing in mind that we are parsing a file that is under our control and has a very defined schema...

bin/update_cucumber
```diff
@@ -77,13 +80,15 @@ class {
...
-            && isset($repository['package']['source']['reference']),
+            && isset($repository['package']['source']['reference'])
+            && is_string($repository['package']['source']['reference']),
             'First composer.json repository should be our custom gherkin package with expected structure',
         );
 
         $composerVersion = $repository['package']['version'];
         assert(
-            is_array($composerConfig['require-dev'] ?? null)
+            is_string($composerVersion)
```
acoulton, line 90: Does this matter? If it's not a string it won't `===` the value at the end of this assert anyway?

bin/update_cucumber
```diff
@@ -100,6 +105,9 @@ class {
         ];
     }
 
+    /**
+     * @return array<array-key, mixed>
```
acoulton, line 109: What does this provide over the native PHP `: array` return type?

bin/update_cucumber
acoulton: Well we know that there will be at least one revision, because we are querying github for a specific package that already has releases. So we are never going to get a list shorter than the one we currently get.

As per the inline comment, the array_pop is a workaround for the fact that the `24.1.0` release is not listed on the new repo, so the script would not be able to find the original "current version" in the list of releases to know that `25.0.0` is the next one. It should blow up (rather than returning false) if somehow it got an empty array. Once #306 is merged that workaround can be deleted anyway as it will be irrelevant.

bin/update_cucumber
acoulton: Yeah I think if we *have* to check these are all strings (again noting that it would also be fine for this to just blow up if somehow github's schema changed and they gave us something that couldn't go into version_compare) then we should array_map and then sort.

bin/update_cucumber
```diff
@@ -205,12 +230,15 @@ class {
...
-        $body = preg_replace('_https://github\.com_', 'https://redirect.github.com', $body);
+        $replacements = [
+            // Remove `#` prefixes on issues and PR numbers (otherwise these will incorrectly autolink to our own repo).
+            '/#(\d+)/' => '$1',
+            // Remove `@` prefixes on github username tags so that they don't create a notification / backlink
+            '/@(\w+)/' => '$1',
+            // Prevent any github links from spamming issues and PRs on other repos with backlinks
+            '_https://github\.com_' => 'https://redirect.github.com',
+        ];
+        $body = preg_replace(array_keys($replacements), array_values($replacements), $nextVersion['changelog']);
```
acoulton, line 241: Is this refactor required for some phpstan reason? If not, personally I prefer it as individual statements.

acoulton (commented):
Thanks @uuf6429 - a few comments / questions.

bin/update_cucumber
uuf6429: OK, in that case I'll keep the code as is.

bin/update_cucumber
```diff
@@ -62,6 +62,9 @@ class {
         );
     }
 
+    /**
+     * @phpstan-return TCurrentVersionArray
```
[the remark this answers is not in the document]
uuf6429: In case another tool is used that does not understand `phpstan-[import-]type` (and would therefore see `TCurrentVersionArray` as an undefined symbol). @stof once raised that point, which I think makes sense in principle.

bin/update_cucumber
uuf6429: I've enabled PHPStan strict checks, which disallows just using `array`.

The reason is that while in some cases (like that one) there is no advantage, PHPStan cannot catch cases where there should have been a better better/detailed type hint.

bin/update_cucumber
uuf6429: Kinda, since `preg_replace()` return `string|null` where $subject (of next call) requires `string`: https://phpstan.org/r/fca54ed7-62df-4da3-b3fb-7c0bb9ef7275

Another option is to typecast the result or the parameter:
```php
$body = (string) preg_replace('/#(\d+)/', '$1', $nextVersion['changelog']);
```

Let me know which one you prefer (although personally I'd prefer this approach - even better if PHP had a native `preg_translate` function).

bin/update_cucumber
uuf6429: Yes, PHPStan is surprisingly thorough!

Here's why:
- it was not asserting that `$composerVersion` (=> therefore it's defined as mixed)
- it didn't assert that `$composerConfig['require-dev']['cucumber/gherkin-monorepo']` is a string (=> therefore also seen as mixed)
- it's possible that `mixed === mixed` (e.g. arrays)
- later one we concatenate `$composerVersion` which at that point is still considered `mixed`

bin/update_cucumber
uuf6429: OK, I'll go for it.

bin/update_cucumber
acoulton: ah, OK. Let's stick with the array syntax then.

acoulton (approved):
Thanks @uuf6429

PR #308, opened 2025-05-04
refactor: Simplify tag filtering implementation

uuf6429:
- added helper methods to construct feature/scenario with one different property (decreases boilerplate code)
- simplified some code:
  - some filtering could be done with lambda or method reference
  - replaced if condition with ternary
  - short-circuited tag matching (theoretically improves tag filtering speed)
- php-cs-fixer decided to fix some code style 🤷

### DIFF
--- a/src/Node/FeatureNode.php
+++ b/src/Node/FeatureNode.php
@@ -224,22 +224,22 @@ class FeatureNode implements KeywordNodeInterface, TaggedNodeInterface
     }
 
     /**
-     * Returns a copy of this features but with a different set of scenarios.
+     * Returns a copy of this feature but with a different set of scenarios.
      *
      * @param array<array-key, ScenarioInterface> $scenarios
      */
     public function withScenarios(array $scenarios): self
     {
         return new self(
-            $this->getTitle(),
-            $this->getDescription(),
-            $this->getTags(),
-            $this->getBackground(),
+            $this->title,
+            $this->description,
+            $this->tags,
+            $this->background,
             array_values($scenarios),
-            $this->getKeyword(),
-            $this->getLanguage(),
-            $this->getFile(),
-            $this->getLine()
+            $this->keyword,
+            $this->language,
+            $this->file,
+            $this->line,
         );
     }
 
--- a/src/Node/FeatureNode.php
+++ b/src/Node/FeatureNode.php
@@ -224,6 +224,26 @@ class FeatureNode implements KeywordNodeInterface, TaggedNodeInterface
     }
 
     /**
+     * Returns a copy of this features but with a different set of scenarios.
+     *
+     * @param array<array-key, ScenarioInterface> $scenarios
+     */
+    public function withScenarios(array $scenarios): self
+    {
+        return new self(
+            $this->getTitle(),
+            $this->getDescription(),
+            $this->getTags(),
+            $this->getBackground(),
+            array_values($scenarios),
+            $this->getKeyword(),
+            $this->getLanguage(),
+            $this->getFile(),
+            $this->getLine()
+        );
+    }
+
+    /**
      * Returns whether the file path is an absolute path.
      *
      * @param string|null $file A file path
@@ -243,7 +263,6 @@ class FeatureNode implements KeywordNodeInterface, TaggedNodeInterface
                 && $file[1] === ':'
                 && strspn($file, '/\\', 2, 1)
             )
-            || parse_url($file, PHP_URL_SCHEME) !== null
-        ;
+            || parse_url($file, PHP_URL_SCHEME) !== null;
     }
 }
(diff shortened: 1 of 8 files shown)


src/Node/FeatureNode.php
```diff
@@ -223,6 +223,26 @@ public function getLine()
...
+        return new self(
+            $this->getTitle(),
+            $this->getDescription(),
+            $this->getTags(),
+            $this->getBackground(),
+            array_values($scenarios),
+            $this->getKeyword(),
+            $this->getLanguage(),
+            $this->getFile(),
+            $this->getLine()
```
acoulton, line 242: Personally for this kind of `clone with` function, I would reference the properties rather than the getters e.g.:

```
return new self(
  $this->title,
  $this->description,
  /// etc
);
```

It's more performant, clearer IMO, and also guarantees the cloned object has the exact same internal state (e.g. if in future a getter was given behaviour other than just returning the raw property value).

tests/Filter/ComplexFilterTest.php
```diff
@@ -0,0 +1,62 @@
...
+        $scenarios = [new ScenarioNode('Scenario#1', [], [], '', 1)];
+        $feature = new FeatureNode('Feature#1', null, [], null, $scenarios, '', '', null, 1);
+
+        $filter = $this->createComplexFilter([$feature], $scenarios);
+
+        $filtered = $filter->filterFeature($feature);
+
+        $this->assertCount(1, $scenarios = $filtered->getScenarios());
+        $this->assertSame('Scenario#1', $scenarios[0]->getTitle());
+    }
```
acoulton, line 31: I'm not sure this is really testing much, it could pass just by `isScenarioMatch` returning true? 

So either:
* if you're just trying to prove the interface, the class implemented in `createComplexFilter` is more complicated than it needs to be and should just return `true` from the match methods.
* if you think we need to test the logic of abstract ComplexFilter separately to the classes that extend it, there should be more test cases to exercise the stub class properly (e.g. with no matching features/scenarios, with a subset of matches, etc).

It depends on the motivation for adding this test?

uuf6429: > it could pass just by isScenarioMatch returning true?

I was thinking I might want to test other scenarios, but now that I think about it it's overcomplicated for no reason. I'll simplify it into one test (essentially I just want to test `filterFeature()` because it was never covered by tests since classes extending this abstract class seem to always override `filterFeature()`).

src/Node/FeatureNode.php
uuf6429: I was considering that but then I worried it might not be backward compatible, e.g. if the end user somehow managed to swap the `$scenario` instance with their own subclass the behaviour would differ and possibly.

But then again those properties are private, so if the user supplied wrong parent-constructor-arguments it's already broken in principle.

I'll fix this as suggested, let me know what you think about the above in the meantime.

acoulton: Yeah, I think obviously it would be better if our classes were `final` in a future major, considering that the filtering process will always clone these back to our own implementations even if the user has subclassed, which could be confusing.

That said, I think it's unlikely anyone would have replaced the hardcoded parts of the `Parser` to instantiate custom classes.

Considering that the properties are all private, I think it's best to clone them direct.

tests/Filter/ComplexFilterTest.php
```diff
@@ -0,0 +1,41 @@
...
+            public function isFeatureMatch(FeatureNode $feature): bool
+            {
+                return true;
+            }
+        };
+
+        $filtered = $nonFilteringFilter->filterFeature($feature);
+
+        $this->assertCount(1, $scenarios = $filtered->getScenarios());
+        $this->assertSame('Scenario#1', $scenarios[0]->getTitle());
```
acoulton, line 39: Thinking about it more, I'm still not sure I see the value of this test.

I could make it pass just by changing the `ComplexFilter` implementation to be a no-op (which should surely break something):

```
public function filterFeature(FeatureNode $feature) {
  return $feature;
}
```

If it's worth having a test at all, then I think it probably at least needs to make sure the method does something? Even just a second example with an anonymous class that returns `false` and proves we get an empty feature?

uuf6429: Hey, I hit a bit of a snag.

Seems that the original `ComplexFilter` never invoked abstract method `isFeatureMatch()`. Funnily enough, the only implementation of this filter (`TagFilter`) also doesn't invoke `isFeatureMatch()`.

This doesn't make much sense to me, but it means there isn't anything to test regarding `isFeatureMatch()` with respect to `ComplexFilter`.

Agreed?

acoulton: I guess `isFeatureMatch()` is just present to satisfy the `FeatureFilterInterface`? I agree if we don't provide an actual implementation of that then there's no need to test it.

I'm still not actually sure why we're testing an abstract class at all - normally I would just test them through their concretions. I suppose at least it will tell us if we break BC for end-users that have extended it.

acoulton (commented):
Lgtm, thanks @uuf6429

acoulton (approved):

carlos-granados (approved):
Good stuff, thanks @uuf6429

tests/Filter/ComplexFilterTest.php
stof: We test that abstract class because it is part of our public API (end users can extend it, it is not an internal one)

Issue #209, opened 2021-02-04
Feature description whitespace does not match Cucumber


ciaranmcnulty:
We just seem to trim it slightly differently

Removing this line will add a failing test
https://github.com/Behat/Gherkin/blob/master/tests/Behat/Gherkin/Cucumber/CompatibilityTest.php#L39

ciaranmcnulty:
Would fixing this be a BC break? We can make a new method that returns the description with the same whitespace as cucumber but that seems over the top. Do people depend on this for something or do they blindly trim it?

stof:
I suggest discussing that with cucumber

acoulton:
I'm actually not sure that our description trimming is better than theirs. I think both are questionable...

With this file:

```gherkin
Feature: Escaped pipes
    The \-character will be considered as an escape in table cell
    iff it is followed by a |-character, a \-character or an n.

  Scenario: They are the future
```

cucumber/gherkin parses the description with the first line trimmed but subsequent lines keeping the original indent:

```
string(125) "The \-character will be considered as an escape in table cell
    iff it is followed by a |-character, a \-character or an n."
```

We parse it with exactly 2 spaces trimmed from each line:
```
string(125) "  The \-character will be considered as an escape in table cell
  iff it is followed by a |-character, a \-character or an n."
```

We trim it here: https://github.com/Behat/Gherkin/blob/e07986ef3e9b6621dcd7069831d96fef9c76d4ac/src/Parser.php#L243-L247

I think we *might* have been intending to completely unident the description lines, but the `$token['indent']` refers to the indentation of the `Feature:` line (which will almost always be zero), so we generally just remove 2 spaces.

Notably the [Behat PrettyFeaturePrinter](https://github.com/Behat/Behat/blob/master/src/Behat/Behat/Output/Node/Printer/Pretty/PrettyFeaturePrinter.php#L100) then re-indents each line of the description (by default by 2 spaces).

It seems to me like either:
* We should follow what cucumber/gherkin do, but this will unexpectedly cause second and following lines to be indented by more than in the original file.
* We should strip all indentation and leave it up to the end-user to indent if required (it's surely only necessary for something like a console PrettyPrinter - if outputting to HTML or whatever then the leading whitespace is irrelevant)?

I don't think I'd consider changing how we trim whitespace to be a BC break.

stof:
[...]

PR #309, opened 2025-05-04
test: Make CucumberND name reading consistent

uuf6429:
According [to the spec](https://github.com/cucumber/messages/blob/e1537b07e511feb6405ed9aa00261ff79d8a9710/jsonschema/GherkinDocument.json#L160C19-L162C23), 'name' is always required and must not be null (simply search for 'name' and you'll see that it is defined as "type: string" and is always listed in "required").

The description is in a similar situation, except that we are also trimming it on top, [for historical reasons](https://github.com/Behat/Gherkin/issues/209) (to make the CucumberND input compatible with our parser).

Technically, these changes are not backword compatible, but not following the spec should be considered undefined behaviour, or?

### DIFF
--- a/src/Loader/CucumberNDJsonAstLoader.php
+++ b/src/Loader/CucumberNDJsonAstLoader.php
@@ -50,8 +50,8 @@ class CucumberNDJsonAstLoader implements LoaderInterface
         $featureJson = $json['gherkinDocument']['feature'];
 
         return new FeatureNode(
-            $featureJson['name'] ?? null,
-            $featureJson['description'] ?? null,
+            $featureJson['name'],
+            $featureJson['description'],
             self::getTags($featureJson),
             self::getBackground($featureJson),
             self::getScenarios($featureJson),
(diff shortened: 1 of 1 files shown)


uuf6429:
PS: It seems this class [was intended for tests](https://github.com/Behat/Gherkin/issues/286#issuecomment-2732822866)...that might change our perspective on handling it.

acoulton (approved):
I think considering the use/purpose of this class, and the fact that allowing for these missing keys was a departure from the spec, this change is OK.

In general terms I agree that normally, usage that departs from the spec is not covered by BC. There'd need to be a strong reason otherwise (e.g. it was compliant with older versions of the spec, we have evidence that it's common practice, or we have tests asserting behaviour in that specific usecase).

acoulton:
@uuf6429 please could you rebase so we can make sure this is still green with the latest cucumber/gherkin & other changes?

acoulton (approved):
Thanks!

carlos-granados (approved):
Look good, thanks @uuf6429

PR #310, opened 2025-05-04
Improve tests expecting exceptions

uuf6429:
This started out from [this suggestion](https://github.com/Behat/Gherkin/pull/294/files#r2000707025), but then I saw a few other issues which I also fixed:
- exception message should be checked first since failures would be more informative (original issue)
  - also applies to message pattern matching
- many `expectException()`+`expectExceptionMessage()` cases could be replaced with `expectExceptionObject()`
- in a few cases we expected the exception too early, which could have hidden the real source of failure
- minor code style improvement: blank line between expectation and action

### DIFF
--- a/tests/Filter/TagFilterTest.php
+++ b/tests/Filter/TagFilterTest.php
@@ -15,7 +15,7 @@ use Behat\Gherkin\Node\ExampleTableNode;
 use Behat\Gherkin\Node\FeatureNode;
 use Behat\Gherkin\Node\OutlineNode;
 use Behat\Gherkin\Node\ScenarioNode;
-use Exception;
+use ErrorException;
 use PHPUnit\Framework\TestCase;
 
 class TagFilterTest extends TestCase
@@ -303,13 +303,13 @@ class TagFilterTest extends TestCase
     private function expectDeprecationError(): void
     {
         set_error_handler(
-            static function ($errno, $errstr) {
+            static function (int $errNo, string $errStr, string $errFile, int $errLine) {
                 restore_error_handler();
-                throw new Exception($errstr, $errno);
+                throw new ErrorException($errStr, $errNo, filename: $errFile, line: $errLine);
             },
             E_ALL
         );
 
-        $this->expectException(Exception::class);
+        $this->expectException(ErrorException::class);
     }
 }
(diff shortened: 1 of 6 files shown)


tests/Node/TableNodeTest.php
```diff
@@ -303,18 +307,18 @@ public function testMergeRowsFromTablePassSeveralTablesShouldBeMerged(): void
...
             10 => ['42', 'everzet', 'qwerty'],
             13 => ['2', 'antono', 'pa$sword'],
         ]);
-
         $new = new TableNode([
             25 => ['id', 'QWE', 'password'],
             210 => ['242', '2everzet', '2qwerty'],
         ]);
 
+        $this->expectException(NodeException::class);
```
uuf6429, line 320: Probably out of scope of this PR, but shouldn't we also check the exception message in these cases?

acoulton: Yeah, I think so - might as well do it while we're looking at this?

acoulton (approved):
Thanks @uuf6429 this is good as-is, but if you have time to add the expected TableNode messages as well that'd be great.

tests/Filter/TagFilterTest.php
```diff
@@ -303,12 +303,13 @@ public function testTagFilterThatIsAllWhitespaceIsIgnored(): void
     private function expectDeprecationError(): void
     {
         set_error_handler(
-            static function ($errno, $errstr) {
+            static function (int $errNo, string $errStr, string $errFile, int $errLine) {
                 restore_error_handler();
-                throw new Exception($errstr, $errno);
+                throw new ErrorException($errStr, $errNo, filename: $errFile, line: $errLine);
```
uuf6429, line 308: I changed it to `ErrorException` so that we do not capture unrelated exceptions by mistake (although the risk was already relatively low since we checked the message).

Additionally, I added the originating file/line parameters, which I think could be useful.

tests/Node/TableNodeTest.php
uuf6429: Done!

acoulton (approved):
Great, thanks :)

carlos-granados (approved):
Looking good, thanks @uuf6429

PR #311, opened 2025-05-04
test: Add excluded cucumber parity tests for cucumber/gherkin up to 32.0.4

acoulton:
The newer cucumber/gherkin releases have brought in a few more examples that are not handled by our parser.

From reviewing these, each of them has some potential details to consider, so I propose we add them to the exclusion lists (so that we can merge the cucumber/gherkin update branch and i18n changes) and then tackle them separately.

I did notice that one of the excluded tests (and a couple of new ones) were only problematic because we handle whitespace in feature descriptions differently to gherkin - as tracked in #209. Rather than exclude those completely, I've changed the test to normalise the whitespace so that we can still assert the rest of the example matches the expectation.

### DIFF
--- a/tests/Cucumber/CompatibilityTest.php
+++ b/tests/Cucumber/CompatibilityTest.php
@@ -59,6 +59,7 @@ class CompatibilityTest extends TestCase
      */
     private array $parsedButShouldNotBe = [
         'invalid_language.feature' => 'Invalid language is silently ignored',
+        'unexpected_end_of_file.feature' => 'EOF after tags is ignored',
     ];
 
     /**
--- a/tests/Cucumber/CompatibilityTest.php
+++ b/tests/Cucumber/CompatibilityTest.php
@@ -48,7 +48,6 @@ class CompatibilityTest extends TestCase
         'spaces_in_language.feature' => 'Whitespace not supported around language selector',
         'incomplete_feature_3.feature' => 'file with no feature keyword not handled correctly',
         'rule_without_name_and_description.feature' => 'Rule is wrongly parsed as Description',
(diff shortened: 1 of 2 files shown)


tests/Cucumber/CompatibilityTest.php
```diff
@@ -158,6 +160,15 @@ private function normaliseFeature(?FeatureNode $featureNode): ?FeatureNode
             return null;
         }
 
+        if ($featureNode->getDescription() !== null) {
+            // We currently handle whitespace in feature descriptions differently to cucumber
+            // https://github.com/Behat/Gherkin/issues/209
+            // We need to be able to ignore that difference so that we can still run cucumber tests that
+            // include a description but are covering other features.
+            $trimmedDescription = preg_replace('/^\s+/m', '', $featureNode->getDescription());
```
uuf6429, line 168: Hey, if we have to fix this here, doesn't that mean we should have just removed the counter hack from [CucumberNDJsonAstLoader](https://github.com/Behat/Gherkin/blob/e07986ef3e9b6621dcd7069831d96fef9c76d4ac/src/Loader/CucumberNDJsonAstLoader.php#L54)?

acoulton: Urgh, I missed that - the test implies that all the normalization happens here.

I agree, we should remove that hack from the NDJSON loader so it returns the exact cucumber parse result

acoulton: Fixed, thanks for pointing that out.

carlos-granados (approved):
Looking good, thanks @acoulton

PR #312, opened 2025-05-04
change: Throw ParserException if file ends with tags

acoulton:
**Based on #311 - only the last commit is new in this branch**.

For parity with cucumber/gherkin, do not accept feature files that end with `@tags` with no subsequent content.

cucumber/gherkin is relatively permissive about incomplete files in some cases (for example, it is explicitly valid to have a file with a Feature: but no scenarios, or a Scenario: with no steps). I suspect this is intentional to avoid parser errors while iterating on a feature.

However, the presence of `@tags` at the end of the file is unusual and suggests that the file is incomplete or damaged. cucumber/gherkin treats this as an exception and so should we.

### DIFF
--- a/tests/ParserTest.php
+++ b/tests/ParserTest.php
@@ -48,14 +48,18 @@ final class ParserTest extends TestCase
     {
         $parser = $this->createGherkinParser();
 
-        $parser->parse(
-            <<<'FEATURE'
-            Feature:
-            Scenario:
-            Given step
-            @skipped
-            FEATURE
-        );
+        try {
+            $parser->parse(
+                <<<'FEATURE'
+                Feature:
+                Scenario:
+                Given step
+                @skipped
+                FEATURE,
+            );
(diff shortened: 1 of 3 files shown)


tests/ParserTest.php
```diff
@@ -48,14 +48,18 @@ public function testParserResetsTagsBetweenFeatures(): void
...
+                <<<'FEATURE'
+                Feature:
+                Scenario:
+                Given step
+                @skipped
+                FEATURE,
+            );
+        } catch (ParserException) {
+            // expected - features cannot end with tags
+        }
```
acoulton, line 62: This test relied on us parsing tags at the end of a file, but I don't think that was expected end-user behaviour. It was simply to prove that the parser state resets correctly on moving to the next file, and since we already know that tags are reset between features / scenarios / etc, the test had to create an artificial case where the tag was not attached to any node.

PR #313, opened 2025-05-05
change: Throw ParserException if file ends with tags

acoulton:
For parity with cucumber/gherkin, do not accept feature files that end with `@tags` with no subsequent content.

cucumber/gherkin is relatively permissive about incomplete files in some cases (for example, it is explicitly valid to have a file with a Feature: but no scenarios, or a Scenario: with no steps). I suspect this is intentional to avoid parser errors while iterating on a partially-complete feature file.

However, the presence of `@tags` at the end of the file is unusual and suggests that the file is incomplete or damaged. cucumber/gherkin treats this as an exception and so should we.

### DIFF
--- a/src/Parser.php
+++ b/src/Parser.php
@@ -569,6 +569,15 @@ class Parser
         }
 
         $nextType = $this->predictTokenType();
+
+        if ($nextType === 'EOS') {
+            throw new ParserException(sprintf(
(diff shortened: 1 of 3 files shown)


carlos-granados (approved):
Looks good, thanks @acoulton

PR #316, opened 2025-05-06
add: Parse `Scenario` as a synonym for `Scenario Outline`

acoulton:
In cucumber/gherkin, the `Scenario:` and `Scenario Outline:` keywords are interchangeable. Either may have (or not have) an `Examples:` table.

We still need to differentiate between `OutlineNode` and `ScenarioNode` in the parsed result, for BC and because Behat treats these differently at runtime.

However, we can safely parse a `Scenario:` that contains an `Examples:` table (which would have previously been invalid) to an `OutlineNode` - exactly as if the user had written `Scenario Outline:`.

This does require refactoring how we parse tags, because:

* Previously, we could make assumptions based on whether we were expecting (or had already seen) an Examples table.
* Now, we can't tell from the tags alone whether they belong to an Examples table, or mark the end of the current Scenario and the start of the next.

Therefore, we now scan ahead to predict which node the tags are attached to. This also allows us to throw a clearer exception if tags are found in unexpected places in the document.

Relates to #153


src/Parser.php
```diff
@@ -43,8 +44,6 @@ class Parser
     private $tags = [];
     private $languageSpecifierLine;
 
-    private $passedNodesStack = [];
```
acoulton, line 46: This was only used to track recursion into nodes for the purpose of making assumptions about whether we could take tags (and looks like it wasn't always unwound). This is now handled differently.

src/Parser.php
```diff
@@ -302,11 +299,8 @@ protected function parseBackground()
         $line = $token['line'];
 
         if (count($this->popTags()) !== 0) {
-            throw new ParserException(sprintf(
-                'Background can not be tagged, but it is on line: %d%s',
-                $line,
-                $this->file ? ' in file: ' . $this->file : ''
-            ));
+            // Should not be possible to happen, parseTags should have already picked this up.
+            throw new UnexpectedTaggedNodeException($token, $this->file);
```
acoulton, line 303: I have kept this for the moment out of caution, I don't think it can happen now (it doesn't with any of the test examples).

src/Parser.php
```diff
@@ -391,8 +351,22 @@ protected function parseScenario()
...
+        $node = $this->parseScenarioOrOutlineBody($this->expectTokenType('Outline'));
+
+        if (!($node instanceof OutlineNode && $node->hasExamples())) {
+            throw new ParserException(sprintf(
+                'Outline should have examples table, but got none for outline "%s" on line: %d%s',
+                $node->getTitle(),
+                $node->getLine(),
+                $this->file ? ' in file: ' . $this->file : ''
+            ));
+        }
```
acoulton, line 363: NB cucumber/gherkin doesn't throw in this case (see #153) - I have kept it for now but plan to look at that separately.

src/Parser.php
```diff
@@ -438,16 +419,43 @@ protected function parseOutline()
...
-                rtrim($title),
-                $line,
-                $this->file ? ' in file: ' . $this->file : ''
-            ));
+        if ($examples !== []) {
+            return new OutlineNode(rtrim($title) ?: null, $tags, $steps, $examples, $keyword, $line);
         }
 
-        return new OutlineNode(rtrim($title) ?: null, $tags, $steps, $examples, $keyword, $line);
+        return new ScenarioNode(rtrim($title) ?: null, $tags, $steps, $keyword, $line);
```
acoulton, line 426: Note that at the moment calling `->getKeyword()` on the parsed node returns a hardcoded result. So if the feature file used `Scenario:` but had an examples table then e.g. Behat's pretty printer would still print `Scenario Outline:`. 

We could add a new property into the nodes to record the actual keyword that we found, but I would be a little nervous about BC impact if the result of `OutlineNode->getKeyword()` can vary in case anyone is matching on / doing anything specific with that string.

tests/ParserExceptionsTest.php
```diff
@@ -205,7 +205,7 @@ public function testWrongTagPlacement(): void
         GHERKIN;
 
         $this->expectExceptionObject(
-            new ParserException('Expected Scenario, Outline or Background, but got Step on line: 6')
+            new ParserException('Step can not be tagged, but it is on line: 6')
```
acoulton, line 208: I think this new exception is clearer than the previous anyway. I have left the assertion as `ParserException` (rather than change to the actual `UnexpectedTaggedNodeException`) to prove that this is all still BC.

src/Parser.php
```diff
@@ -438,16 +419,43 @@ protected function parseOutline()
...
+     * @throws UnexpectedTaggedNodeException if there is not a taggable node
+     */
+    private function predictNextTaggedNodeType(): string
+    {
+        $deferred = [];
+        try {
+            while (true) {
+                $deferred[] = $next = $this->lexer->getAdvancedToken();
+
+                if (in_array($next['type'], ['Tag', 'Comment', 'Newline'], true)) {
```
carlos-granados, line 441: Nitpick: I would probably assign `$next['type']` to a variable instead of accessing it several times

carlos-granados (approved):
Approved, just a tiny comment, feel free to ignore it

tests/Cucumber/CompatibilityTest.php
```diff
@@ -46,7 +46,6 @@ class CompatibilityTest extends TestCase
         'extra_table_content.feature' => 'Table without right border triggers a ParserException',
         'incomplete_scenario_outline.feature' => 'Scenario and Scenario outline not yet synonyms',
         'padded_example.feature' => 'Scenario and Scenario outline not yet synonyms',
```
stof, line 48: what about those 2 that were skipped for the same reason ?

stof (commented):
should we throw a ParserException if we find steps _after_ an examples table ? currently, the parser seems to allow mixing them AFAICT (unless the token type determination returns a different result for such case).

acoulton:
> should we throw a ParserException if we find steps after an examples table?

Yeah, probably - it does seem that (as now) we don't detect / fail in that situation. I'll double check what cucumber/gherkin does.

tests/Cucumber/CompatibilityTest.php
acoulton: They now fail for unrelated reasons.

* `incomplete_scenario_outline.feature` because we throw if an Outline has no Examples table, but cucumber/gherkin does not (tracked in #153 and noted in the diff above).
* `padded_example.feature` because we don't trim table padding (specifically &nbsp; I think) the same as cucumber/gherkin - no issue for this yet (I plan to create one), and it has BC implications.

I'll update the reason descriptions in the testcase

acoulton:
I checked and `@cucumber/gherkin` does indeed throw a Parser error if `Steps` follow `Examples:` - https://codesandbox.io/p/devbox/gherkin-parser-ast-forked-f9j44t?workspaceId=ws_Mrsa3f4At4n3w9MazShoyf

I have added a test for that and updated the Parser to detect and throw in this situation.

stof:
> I have added a test for that and updated the Parser to detect and throw in this situation.

I suggest contributing such a test for bad data to the cucumber testsuite as well, to make it part of the shared tests in the future.

src/Parser.php
stof: I think solving #153 fully is just a matter of removing that exception, by _fully_ making them synonyms.

acoulton: I agree, I just mean it's a separate PR/change. It relates to other cases where we currently throw on partially-complete feature files that gherkin accepts and I think we should consider that in the round.

stof: well, to me, accepting outlines without examples (returning them as a ScenarioNode) is part of making `Scenario` and `Scenario Outline` be synonyms. Right now, you are making `Scenario` a synonym of `Scenario Outline` but `Scenario Outline` is not a synonym of `Scenario`, which is quite weird. Synonyms are meant to be bidirectional relations to me.

stof: I would say that turning a parsing exception into a supported case won't break CI of anyone: if they have files matching that case, it will potentially _fix_ their CI. Turning invalid files into valid ones is generally not considered a BC break (otherwise, it would be impossible to add any new syntax)
And this is also exactly the kind of changes you are doing for files using `Scenario` that have an `Examples` table today (which are also a ParserException).

But anyway, I won't block the merge on that if you decide to keep it split.

acoulton: Fair points and broadly I agree, I just think there's a philosophical difference between changing behaviour that is undefined, and changing behaviour that we have explicit tests for. Hence I prefer the discipline of considering the second type of changes separately. The next PR will be very soon :smile:

src/Parser.php
```diff
@@ -404,50 +378,94 @@ protected function parseOutline()
...
-        return new OutlineNode(rtrim($title) ?: null, $tags, $steps, $examples, $keyword, $line);
+        return new ScenarioNode(rtrim($title) ?: null, $tags, $steps, $keyword, $line);
+    }
+
+    /**
+     * Peek ahead to find the node that the current tags belong to.
+     *
+     * @throws UnexpectedTaggedNodeException if there is not a taggable node
+     */
+    private function predictNextTaggedNodeType(): string
```
carlos-granados, line 443: This function actually does two things: validates that the next tagged node is correct and returns it. But the function name does not reflect the validation part, so when we call it and don't use the return value it looks weird. Also `predict` sounds a bit magic in my head 😆 . What about `validateAndGetNextTaggedNodeType()`?

carlos-granados (commented):
@acoulton looking great overall, just a small comment

acoulton:
@carlos-granados good catch, thanks. The method started off just looking ahead, and I named it `predict` to follow the naming of other methods in this class & the Lexer. But then moved the validation & exception inside the method and didn't rename it. Fixed now.

PR #323, opened 2025-05-16
change: Do not throw on some unexpected Feature / Language tags

acoulton:
Historically, we have thrown exceptions if:

* A `Feature:` line is found after the first `Feature:` line
* A language comment is found after the first language comment.

This is not the case in cucumber/gherkin, which stops looking for features and languages after it finds them - they are just treated as text / comments and are valid in places where those are allowed.

I have changed our behaviour to match cucumber/gherkin. As with #322, the new testdata files will be contributed back to cucumber/gherkin in due course.


src/Parser.php
```diff
@@ -82,7 +80,7 @@ public function parse($input, $file = null)
         }
 
         $feature = null;
-        while ('EOS' !== ($predicted = $this->predictTokenType())) {
```
acoulton, line 85: This `$predicted` variable was already unused.

src/Parser.php
```diff
@@ -640,17 +616,8 @@ protected function parseLanguage()
...
-        } elseif ($token['line'] !== $this->languageSpecifierLine) {
-            throw new ParserException(sprintf(
-                'Ambiguous language specifiers on lines: %d and %d%s',
-                $this->languageSpecifierLine,
-                $token['line'],
-                $this->file ? ' in file: ' . $this->file : ''
-            ));
-        }
+        // Switch the lexer to the specified language and tell it to treat any further language tags as comments
+        $this->lexer->analyse($this->input, $token['value'], true);
```
acoulton, line 620: Theoretically, if someone had injected a Lexer of their own which does not recognise the new `$hasSeenLanguageTag = false` then this could cause an infinite loop.

I'm not sure how much of a risk that is, and how best to avoid it.

Our own Lexer *cannot* now return a second language tag when used with our own Parser (anything after the first one will just be a comment) - so we could keep some state tracking in the Parser so that it still throws on a second language tag?

acoulton:
Urgh, it's annoying that codecov has given this a red check (and therefore marked the whole PR as ❌ in the PR list). Coverage has dropped slightly because I've deleted code that was covered but is not required, meaning the pre-existing uncovered code is now a bigger % of the total.

@uuf6429 do you know how to avoid that happening? Is there a way to get the codecov report comment without it applying a commit status?

src/Lexer.php
```diff
@@ -58,7 +59,7 @@ public function __construct(KeywordsInterface $keywords)
      *
      * @throws LexerException
      */
-    public function analyse($input, $language = 'en')
+    public function analyse($input, $language = 'en', $hasSeenLanguageTag = false)
```
stof, line 62: the new argument should have a native `bool` type (adding a new parameter is already a BC break for child classes, so better avoid a second later BC break for the same argument by adding the type separately)

stof: If we want to avoid a BC break, we could instead add a method (which could even be tagged as an `@internal` one to switch that flag just after calling `analyse` (which will still work as `analyse` itself does not parse things depending on that flag).

tests/Cucumber/extra_testdata/good/language_tag_after_feature_started.feature
```diff
@@ -0,0 +1,4 @@
+Feature: Some feature
+# language: ru
+
+    Given something wrong
```
stof, line 4: This test case where `Given something wrong` is actually part of the description rather than being a step is confusing IMO. I suggest adding a `Scenario:` line. Otherwise, this test would _also_ be covering the weird edge case of "step keyword in feature description" that has its dedicated test. It makes tests more reliable (even more important for bad tests actually, as our CompatibilityTest does not guarantee that we fail for the same reason than cucumber as they accept any failure)

src/Lexer.php
stof: Btw, I'm wondering whether we really need to restart an analysis from scratch. Maybe we could have a `setLanguage()` method that toggles this flag, changes the language in `$this->keywords` and resets the 2 keyword-related caches (so that they are rebuilt on next usage), allowing us to continue parsing from where we are, as done in cucumber.
I think the only potential issue is related to the deferred objects and stashed tokens, but I don't know whether we have a way to ever use those properties before changing the language, as it is expected to come early in the file, before any meaningful steps. Maybe `setLanguage` could assert that there is no deferred object and no stashed token (reporting a parser exception about an invalid state when changing language) to ensure this assumption does not lead to weird invalid parsing (and if there's a valid case that invalidates this assumption, we could likely obtain the gherkin file triggering it to verify the behavior).

tests/Cucumber/extra_testdata/good/language_tag_after_feature_started.feature
stof: same for all other test cases that are not about "step keyword in feature description" IMO

src/Lexer.php
acoulton: Urg, good point on the BC - I've gone a bit code-blind this afternoon. I'll have a look next week and see what I can come up with. I'm a little nervous about changing the flow of re-starting analyse or not since these are quite tightly coupled non-final classes with a lot of protected methods and we don't know how anyone has extended them (though I doubt many have, in practice).

tests/Cucumber/extra_testdata/good/language_tag_after_feature_started.feature
acoulton: Good shout. These are all based on the structures we had in the existing ParserExceptions examples but I agree they're confusing. Will fix.

stof:
@acoulton this codecov status is not blocking the merge. To me, getting a red status if code coverage regresses allows to draw attention on this. But if we have a good justification for it (exactly your case of removing unnecessary covered code rather than reducing the ratio by adding more covered code), we can still merge it.

src/Lexer.php
stof: I suggest we also deprecate extending the lexer (there is no valid point doing it IMO), so that we can make maintenance easier in the future.

stof: Not restarting the analysis might indeed be too risky if we care about potential extension. In that case, I suggest using the solution in my second comment: having an internal method `markLanguageAsFound()` or something like that called just after `analyze`.

uuf6429:
> Urgh, it's annoying that codecov has given this a red check (and therefore marked the whole commit & PR as ❌ in the list views). Coverage has dropped slightly because I've deleted code that was covered but is not required, meaning the pre-existing uncovered code is now a bigger % of the total.
>
> @uuf6429 do you know how to avoid that happening? Is there a way to get the codecov report comment without it applying a commit status?

I'm not aware of anything, but I wouldn't be bothered unless it's blocking us or something.

It does kinda make sense in principle - the check is supposed to _warn_ us when coverage is lower, no matter what - but that doesn't mean we _have_ to act.

src/Lexer.php
```diff
@@ -35,6 +35,7 @@ class Lexer
     private $stashedToken;
     private $inPyString = false;
     private $pyStringSwallow = 0;
+    private $allowLanguageTag = true;
     private $featureStarted = false;
```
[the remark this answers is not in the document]
carlos-granados: Agree, I think it would be better

src/Parser.php
carlos-granados: Yes, keeping a tracker so that we can throw on a second language tag seems good. Otherwise people might expect that having two language tags should work and might be left wondering when nothing happened

acoulton:
@stof @uuf6429 on the coverage - I guess. It's just a bit annoying GH doesn't support a "warning" status, I don't mind seeing the coverage failure in the overall PR summary at the bottom, but the red X on each commit makes it harder to see if there are actual test / codestyle etc failures. Probably just have to live with it.

acoulton:
I have looked at the language parsing again based on @stof's feedback and I now think it is actually safer (and better) to just move all the language handling inside the `Lexer`, and to continue reading the document from the current location rather than analysing it again.

* With the "out-of-the-box" Parser & Lexer, this all works as expected (as proved by the tests).
* With an "out-of-the-box" Lexer and a custom Parser: worst-case is the third-party Parser will just follow the old behaviour of calling analyse() again for the first language tag. The Lexer will still handle the language change correctly and re-start from the beginning as it does now, then continue through the document.

The only potential behaviour change is if the user has:
* An "out-of-the-box" Parser
* A custom class that extends Lexer **and** extends `scanLanguage` (or `getNextToken` so that `scanLanguage` is not called)
* One or more feature files that contain language tags.

In this case, our new Parser will expect the Lexer to handle the language change internally. If it doesn't, then the language change will be ignored & parsing will fail due to unknown keywords.

I think that combination of circumstances is extremely unlikely. It will also be hard to protect against / work around pre-emptively without risking other bugs. So we should probably just accept that (very small) risk.

src/Lexer.php
```diff
@@ -74,10 +75,24 @@ public function analyse($input, $language = 'en')
...
+
+    private function setLanguage(string $language): void
+    {
+        if ($this->stashedToken !== null) {
+            throw new LexerException('Cannot set gherkin language when there is already a stashed token');
+        }
+
+        if ($this->deferredObjects !== []) {
+            throw new LexerException('Cannot set gherkin language when there are already deferred tokens');
+        }
```
acoulton, line 94: I've included these defensive methods in case of use with unexpected `Parser` implementations. They're not really possible to test (other than by testing the Lexer directly with a fake/unexpected sequence of method calls, which I think is probably not necessary).

We could alternatively use `assert()` here - but there's no guarantee that assert functions will be enabled in end-user environments, so it may not provide much benefit.

acoulton: @stof I have switched to a LogicException with a clearer message.

I tried to write a test to trigger that, but it is impossible using the public API of the Lexer "out of the box":

- `stashedToken` is only populated during `predictToken()`, and only if there is not already a stashed token. `predictToken()` internally does not actually "predict", it parses the next token and advances / updates the internal state. It just caches the result so that the next `getAdvancedToken` will return it instead of reading the next token in the file. Therefore if a language tag is present, our new `setLanguage` will be called during the `predictToken` call, but only if `stashedToken` is null.
- `deferedTokens` is only set if the user has called `deferToken` to put a token back onto the queue (or, theoretically, inject some custom token for some reason). However, if there are deferred tokens then both `predictToken` and `getAdvancedToken` simply return them in sequence before they read any more tokens from the input. Therefore, if there are any `deferredTokens`, `scanLanguage` will never be called and so can never call `setLanguage`.
- The only other way to call `setLanguage` is from calling `analyse` - but this fully resets the state of the Lexer including any stashed and deferred tokens, so again can't cause a problem.

I have left the exception in, as I think there's still a small risk someone has extended the Lexer class and this is a reasonable change to our internal logic flow through the protected methods. However I've excluded from code coverage.

tests/Cucumber/extra_testdata/good/feature_keyword_in_scenario_description.feature
```diff
@@ -0,0 +1,4 @@
+Feature: One
+  Scenario: Something
+
+Feature: Two
```
stof, line 4: Do we also have a test (maybe upstream?) covering a file that would try to put 2 _meaningful_ features ? In that case, the `Feature` keyword would not be part of the description of something (as steps don't have description)

acoulton: @stof does this one cover it (also added in this PR, but as a "bad" example)? https://github.com/Behat/Gherkin/pull/323/files#diff-82e1f68b1699d4413bab996af3f48d6d66c34e1cece5802ddc5ce31c8e931ad5

It's a special kind of input, but it doesn't require special handling. After the first `Feature:` tag then a line that begins `Feature:` is just text content, and therefore valid / invalid depending on whether the current position in the document allows free text. This appears to be what cucumber/gherkin is doing.

stof: yes, this covers it and matches what I thought.

The existing code of the Parser indeed handles it (thanks to its handling of unexpected token at that position), but I think it is good to actually cover it with a "bad" example.

stof (approved):

stof:
The code coverage reduction seems to be because one case of throwing an exception in `parseFeature` (saying `Expected Scenario, Outline or Background, but got %s on line: %d%s`) is not covered anymore.
Can you check which kind of bad input can reach this error ? Maybe this is a case of using an `Examples` table in a `Feature` for instance. Adding a "bad" example for it might restore code coverage.

acoulton:
@stof I'm not convinced it *can* be triggered with content that is actually invalid in `cucumber/gherkin`.

For example:
```
Feature: Whatever

  Examples:
    | some |
    | thing  |
```

Does indeed trigger that exception in our parser. However, that actually parses as valid with `cucumber/gherkin` - the examples and table just become part of the feature description. 

Essentially AFAICS, we mostly take an opposite approach to them - we parse things that look like tags as tags and see if they're valid in the context, where they only look for tags that are valid in context and treat everything else as text. So there are various things that are invalid for us (and look odd) but are accepted by them.

The only way I can make `cucumber/gherkin` reject unexpected content in the `Feature` node is with things like tags, a second background, free text in places where it's not allowed, etc - but we already detect & throw for those in different places.

stof:
OK, so the newly untested code is code that corresponds to a parity mismatch with upstream.
This should be recorded as issues about parity gap. And it might make sense to contribute more of those `*_keyword_in_*_description` examples upstream for the various kind of keywords that don't end a description, to ensure all of them are covered.

acoulton:
@stof I've consolidated the two separate `throw` statements into a single one with just the message varying, similar to what we now do when parsing a scenario / scenario outline.

This actually reduces coverage for this specific patch, but increases it overall. 

In trying to trigger this, I've also identified that we currently accept a `Background` after a `Scenario`, which cucumber/gherkin does not. I'll log that separately to come back to.

stof (approved):
Let's merge this.

Next step is to submit those testdata upstream.

PR #324, opened 2025-05-16
Parse `Scenario Outline` with no examples as a `Scenario`

acoulton:
Continues #153 and builds on #294 (where I gave a wrong steer on the NDJson loading) and #316.

Instead of throwing an exception if a `Scenario Outline` has no examples, we will just parse it as a `Scenario` - this matches the behaviour of cucumber/gherkin.

### DIFF
--- a/tests/Cucumber/CompatibilityTest.php
+++ b/tests/Cucumber/CompatibilityTest.php
@@ -174,12 +174,26 @@ class CompatibilityTest extends TestCase
             // include a description but are covering other features.
             $feature->getDescription() === null ? null : preg_replace('/^\s+/m', '', $feature->getDescription()),
             $feature->getTags(),
-            $feature->getBackground(),
+            $this->normaliseBackground($feature->getBackground()),
             array_map($this->normaliseScenario(...), $feature->getScenarios()),
             $feature->getKeyword(),
             $feature->getLanguage(),
             $feature->getFile(),
             $feature->getLine(),
+        );
+    }
+
+    private function normaliseBackground(?BackgroundNode $background): ?BackgroundNode
+    {
+        if ($background === null) {
+            return $background;
+        }
+
+        return new BackgroundNode(
+            $background->getTitle(),
+            array_map($this->normaliseStep(...), $background->getSteps()),
+            $background->getKeyword(),
+            $background->getLine(),
         );
     }
 
--- a/tests/Cucumber/CompatibilityTest.php
+++ b/tests/Cucumber/CompatibilityTest.php
@@ -14,6 +14,7 @@ use Behat\Gherkin\Exception\ParserException;
 use Behat\Gherkin\Keywords;
 use Behat\Gherkin\Lexer;
 use Behat\Gherkin\Loader\CucumberNDJsonAstLoader;
+use Behat\Gherkin\Node\BackgroundNode;
 use Behat\Gherkin\Node\FeatureNode;
 use Behat\Gherkin\Node\OutlineNode;
 use Behat\Gherkin\Node\ScenarioInterface;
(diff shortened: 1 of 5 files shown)


tests/Cucumber/CompatibilityTest.php
```diff
@@ -171,6 +170,13 @@ private function normaliseFeature(?FeatureNode $featureNode): ?FeatureNode
             $this->setPrivateProperty($featureNode, 'description', $trimmedDescription);
         }
 
+        if ($featureNode->getBackground()) {
+            foreach ($featureNode->getBackground()->getSteps() as $step) {
+                $this->setPrivateProperty($step, 'keywordType', '');
+                $this->setPrivateProperty($step, 'arguments', []);
+            }
+        }
+
```
acoulton, line 179: We need to do the same fixes to be able to compare steps in a Background that we already do for steps in a Scenario (for now, at least).

stof: this will conflict with #319 which changes this normalization logic. As the other PR is approved, I suggest merging it before finishing the work here.

acoulton:
I think we may have an unexpected issue / edge case here.

It is valid to have a feature like this:

```gherkin
  Scenario Outline: no steps, no examples
    When I do something

  Scenario Outline: no steps, no table
    When I do something

    Examples:

  Scenario Outline: no steps, only table header
    When I do something

    Examples:
    | what |
```

With this change, the first Scenario Outline (which would previously have been invalid) is parsed as a `ScenarioNode`, and the "When I do something" step will run.

However for the second two, I think (haven't checked yet):
* They will be parsed as an `OutlineNode` because they do have an `Examples:` child even though it is empty.
* The `ExampleTableNode->getColumnsHash()` will return an empty array (because the table is either empty or contains only a header row).
* Therefore `OutlineNode->getExamples()` will also produce an empty array, and nothing will run.

It is valid for `getExamples()` to produce an empty array - e.g. if all the examples have been filtered out by a tag or similar. So there shouldn't be any existing code that expects that to always be non-empty.

But it might be surprising to a user that adding just the `Examples:`  keyword at the end of an otherwise-valid `Scenario` would prevent it from running at all?

We could filter `ExampleTableNode` that are empty in the gherkin file before we decide whether to return an `OutlineNode` or a `ScenarioNode`. However that would then prevent anything from seeing / rendering them e.g. in the pretty printer output.

I'll have a look next week to see how this all actually plays out in Behat and in the official cucumber implementations. I may be overthinking on a Friday afternoon!

tests/Cucumber/CompatibilityTest.php
acoulton: Agreed, I'd meant to note that in my annotation.

acoulton: FYI #319 is not merged yet because it's so far only approved by me and per our policy we want 2 approvals on PRs that are not from a maintainer.

acoulton:
I have checked this with Behat and with cucumber-js (see example on [codesandbox](https://codesandbox.io/p/devbox/cucumber-js-demo-forked-jkpxrq?file=%2Ffeatures%2Fgreeting.feature%3A6%2C38)).

As expected, as soon as there is an `Examples:` line at the end of a Scenario, it becomes an OutlineNode with no actual examples. This means that (in both Behat and cucumber-js):

* None of the steps run
* Formatters like progress / junit etc therefore don't include it in any of the number of scenarios / number of steps etc.
* The pretty format just outputs the scenarios before & after. No examples actually ran, so none of the scenario outline is printed. This is the same as if you'd e.g. used tags to filter out all the examples for an outline.

So although it might seem unusual that a user could turn a working feature into a no-op just by adding `Examples:` at the end of a scenario, that is consistent with how the other cucumber runners work, and does not produce any errors in Behat. Therefore I think this is OK.

stof:
@acoulton I'm fine with the behavior then, as it is consistent with upstream.

I guess a gherkin linter could report a warning for such empty examples table. For instance, [`gherkin-lint`](https://github.com/gherkin-lint/gherkin-lint) does it in its `no-scenario-outlines-without-examples` rule

stof (approved):

acoulton:
@stof thanks. Yeah I'd been wondering a bit about the overall impact of cucumber/gherkin's relatively relaxed parsing of feature files - makes sense when first iterating on a feature, but perhaps not want you want in long-term CI. But I agree that linting is a better solution than trying to create some sort of in-built `strict` mode so in due course we can point people in that direction.

PR #325, opened 2025-05-21
test: Compare step arguments when checking gherkin parity

### LINKED ISSUE #320: CucumberNDJsonAstLoader does not parse tables and pystrings for tests
From https://github.com/Behat/Gherkin/issues/319#issuecomment-2876628868 -
currently we are not comparing `arguments` (TableNode and PyString) in the cucumber CompatibilityTest.

This is because because [CucumberNDJsonAstLoader](https://github.com/Behat/Gherkin/blob/master/src/Loader/CucumberNDJsonAstLoader.php#L139) doesn't attempt to load/parse step arguments. 

It looks like those are present in the NDJson (as either `docString` or `dataTable` keys) so I'm not sure whether there was a problem parsing them originally, or whether it was just felt that it was not required.

If possible, we should parse those arguments and compare them to ensure that our parser treats these the same as cucumber/gherkin.

acoulton:
Originally, the NDJson loader did not attempt to create PyString / TableNode objects from the cucumber/gherkin AST. As a result, the CompatibilityTest just removed arguments our parser had included with any StepNodes, before comparing to the cucumber result.

This commit instead hydrates this testdata from the gherkin AST. This allows us to check parity of parsing multiline step arguments.

Fixes #320


acoulton:
Coverage drops slightly because the excluded `docstrings.feature` contains the only example of a docstring like:

~~~
  ```
  first line
  """
  third line
  ```
~~~

stof (approved):

stof:
I'm fine with the drop of code coverage if we assume this is temporary until the new identified parity issues are fixed.

If the fix is not done, it would be worth adding a testdata covering a subset of the upstream pystrings.feature (to cover the basic case)

acoulton:
@stof good plan - I've copied over the docstrings.feature and just commented out the parts that are broken, that restores the coverage.

stof (approved):

PR #331, opened 2025-05-21
Use a custom object comparator to ignore the keywordType of StepNode

stof:
The keywordType of our StepNode is not a concept that exists in the Gherkin AST of cucumber. It actually corresponds to the pickle step type in their architecture.
Using a dedicated object comparator (registered only for cucumber compatibility tests) allows to ignore that property without having to normalize steps, which makes the code a lot clearer.

The handling of the feature description is kept in a normalization step as this is related to a parity difference with cucumber which is meant to be solved at some point instead.


stof:
the code coverage reduction is due to the fact that the normalization logic (now that it was not done with Reflection anymore) was calling the `getKeyword` method of a bunch of classes, while those simple getters are not called anymore (`assertEquals` compares the properties values without using the getters).

Issue #285, opened 2025-02-10
Can table nodes really contain scalar cell values?


uuf6429:
If we look at the following code:
https://github.com/Behat/Gherkin/blob/3d7bfdd381b0b0bdb1e841dac97864477b89e270/src/Behat/Gherkin/Node/TableNode.php#L64 it gives the impression that we can have tables with scalar values, even though AFAIK we parse a table from a string without any type conversion (therefore can only have string cells).

Why is this significant? While implementing PHPStan, I started with scalar cell values, but later on this becomes more and more ambiguous.

For example:
1. We calculate the cell length just two lines below: https://github.com/Behat/Gherkin/blob/3d7bfdd381b0b0bdb1e841dac97864477b89e270/src/Behat/Gherkin/Node/TableNode.php#L68 Assuming the type gets converted, the length of, for instance `false` is 0, because PHP casts it to an empty string (instead of `'0'`).
2. Elsewhere we replace tokens with `str_replace()` - which produces some odd results: https://3v4l.org/QTtAK
   https://github.com/Behat/Gherkin/blob/e016902ab9dcca51d72d22c64d1424a2d03e7b14/src/Behat/Gherkin/Node/ExampleNode.php#L259

----

So, should I...
1. keep scalar and fix/improve the strange usages (e.g. explicit typecast in str_replace)
2. OR should I replace `is_scalar` with `is_string`?

In any case, how much of a BC break is it?

uuf6429:
Seems to relate to https://github.com/Behat/Gherkin/pull/121 - it doesn't explain or provide a good reason why scalar is allowed (other than tests passing numbers). Also slightly relates to https://github.com/Behat/Gherkin/issues/134, which was closed as duplicate (even though I can't find the duplicate).

acoulton:
So I think this depends on how you view the `TableNode` class.

* If it is something that is only created by parsing Gherkin, then you're right, the values can only be strings, and tightening that (and fixing the unit tests that pass ints) would be OK
* If it is something that end-users should also be able to create, then tightening it up would indeed be a BC break.

I don't think the project has ever marked TableNode as an internal / consume-only class, so I think we have to keep supporting the second case.

[...]

PR #332, opened 2025-05-21
Add precise type info to solve phpstan errors

stof:
This solves all phpstan level 6 errors except in 3 files:

- ArrayLoader and CucumberNDJsonAstLoader will be handled in separate PRs as adding types for them involves describing the shape of the array structures they process. Handling them separately will simplifying the review (as each of them will probably have a diff close to the full size of the diff of this PR due to that) => see #333 and #334 
- TableNode, because the types to use to fix errors depend on the discussion in #285

### DIFF
--- a/src/Node/FeatureNode.php
+++ b/src/Node/FeatureNode.php
@@ -24,7 +24,7 @@ class FeatureNode implements KeywordNodeInterface, TaggedNodeInterface
     use TaggedNodeTrait;
 
     /**
-     * @param string[] $tags
+     * @param list<string> $tags
      * @param ScenarioInterface[] $scenarios
      * @param string|null $file the absolute path to the feature file
      */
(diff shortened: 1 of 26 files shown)


src/Node/FeatureNode.php
```diff
@@ -24,7 +24,7 @@ class FeatureNode implements KeywordNodeInterface, TaggedNodeInterface
     use TaggedNodeTrait;
 
     /**
-     * @param string[] $tags
+     * @param list<string> $tags
```
stof, line 27: this one is technically a fix for a level 7 error: `getTags` in TaggedNodeInterface requires returning a list, and the readonly promoted property is what gets returned, so we need to expect the same in the constructor (which is quite fine as the Parser produces a list anyway)

acoulton (approved):
Thanks @stof

PR #333, opened 2025-05-21
Add types for the CucumberNDJsonAstLoader

stof:
To make maintenance easier, I defined phpstan type aliases matching exactly the JSON Schema of Cucumber Messages, including properties we don't use for now (like the `id` generated for them)

`TLocation` matches the `Location` schema in https://github.com/cucumber/messages/blob/main/jsonschema/Location.json (referenced by the GherkinDocument for basically all types)
Other aliases matches definitions in https://github.com/cucumber/messages/blob/main/jsonschema/GherkinDocument.json (with a `T` prefix, following our convention)
`TEnvelope` is a partial declaration of the type from https://github.com/cucumber/messages/blob/main/jsonschema/Envelope.json defining only the `gherkinDocument` type (I did not want to define types for all other parts of Cucumber Messages as we would not load them anyway, as our use case is loading an AST file).

A bunch of usages of `array_values` and `??` are removed because phpstan detects that they are unnecessary (`array_values` on a values that is already a list, or accessing a key that is required to be present)

### DIFF
--- a/src/Loader/CucumberNDJsonAstLoader.php
+++ b/src/Loader/CucumberNDJsonAstLoader.php
@@ -11,6 +11,7 @@
 namespace Behat\Gherkin\Loader;
 
 use Behat\Gherkin\Exception\NodeException;
+use Behat\Gherkin\Node\ArgumentInterface;
 use Behat\Gherkin\Node\BackgroundNode;
 use Behat\Gherkin\Node\ExampleTableNode;
 use Behat\Gherkin\Node\FeatureNode;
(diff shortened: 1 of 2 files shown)


acoulton (approved):
Great, thanks @stof 

Out of interest, did you produce those phpstan types by hand, or is there a tool to generate them from a JSON Schema?

stof:
I did them by hand.

uuf6429:
Thanks for the work, but I'd like to sadly point out that I've had a PR with all this prepared since months, but the other PRs were blocking it. 😞 

Anyway, comparing this to mine, I have a few questions:
1. It seems that there are a few cases where the json schema defines some fields optional, but this PR defined the type as required (and removed the `?? []`), such as `TRule.children`. In the short term it's fine (tests are passing), but I suppose we might hit some test that requires it.. so wouldn't it make sense to stick to the schema as much as possible?
2. I noticed that you used fallbacks when using PHPStan types (`@phpstan-return TXX ... @return array`) - should we go for this or just stick to one (e.g. just `@phpstan-return` when an alias is used)?

stof:
`TRule.children` is actually not used anywhere in the loader (as well

acoulton:
> Thanks for the work, but I'd like to sadly point out that I've had a PR with all this prepared since months, but the other PRs were blocking it. 😞

Sorry to hear that, appreciate that will be frustrating. I know you'd been looking at PHPStan but I / we weren't aware you had work beyond what you'd already submitted.

I know there's an issue with GitHub not supporting dependent PRs which makes life hard for things like this.

However in future I'd suggest still either submitting a draft PR for all your work-in-progress with a note that it's not ready for review / you're planning to split it up. Or at least opening an issue and keeping it updated with the status of what you have done locally. That will help us to make sure we're not cutting across anything you're doing.

PR #334, opened 2025-05-21
Add precise types for the ArrayLoader

stof:


stof:
not ready yet. When analyzing my need type using phpstan level 7+, I'm getting issues reported due to phpstan messing up the array shapes. I reported them at https://github.com/phpstan/phpstan/issues/13061 and https://github.com/phpstan/phpstan/issues/13059. Maybe we'll get a fix soon that would allow defining them properly.

The supported type for `ArrayLoader::loadExamplesHash` even seems ambiguous due to the first shortcut it has (or maybe it needs to be more complex by forcing the example table hash itself to use `int<1, max>` due to `0` not being a line number, which might match our etalons in fixtures)

stof:
I'm marking this PR as ready. It allows to enable the level 6 of phpstan. The extra errors reported at level 7 can be handled later (especially as they are more about cases that should be improved in phpstan itself)

acoulton (approved):

PR #335, opened 2025-05-21
Use the `Yaml::parseFile` API to handle Yaml files

stof:
This lets the Yaml component deal with the case of failures when reading the file (which it does).


stof:
This has the benefit of solving some errors reported by phpstan at level 7 due to the `string|false` return type of `file_get_contents`

acoulton (approved):
Thanks :)

PR #338, opened 2025-05-22
Remove the line length from the NewLine token value

stof:
This is the only token type using an integer value, and this value is never used by the Parser.


acoulton (approved):
I suppose in theory an external caller could be using this value, but I think that's very unlikely for what it is.

PR #339, opened 2025-05-22
Add precise types for the TableNode

### LINKED ISSUE #285: Can table nodes really contain scalar cell values?

stof:
This documents that TableNode expects string values for all cells in the table, which is what parsing a gherkin document produces and what the code already assumes internally. Closes #285

### DIFF
--- a/src/Node/ExampleNode.php
+++ b/src/Node/ExampleNode.php
@@ -198,14 +198,16 @@ class ExampleNode implements ScenarioInterface, NamedScenarioInterface
      */
     protected function replaceTableArgumentTokens(TableNode $argument)
     {
-        $table = $argument->getTable();
-        foreach ($table as $line => $row) {
-            foreach (array_keys($row) as $col) {
-                $table[$line][$col] = $this->replaceTextTokens($table[$line][$col]);
+        $replacedTable = [];
+        foreach ($argument->getTable() as $line => $row) {
+            $replacedRow = [];
+            foreach ($row as $value) {
+                $replacedRow[] = $this->replaceTextTokens($value);
             }
+            $replacedTable[$line] = $replacedRow;
         }
 
-        return new TableNode($table);
+        return new TableNode($replacedTable);
     }
 
     /**
--- a/src/Node/TableNode.php
+++ b/src/Node/TableNode.php
@@ -20,6 +20,8 @@ use ReturnTypeWillChange;
  * Represents Gherkin Table argument.
  *
  * @author Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * @template-implements IteratorAggregate<int, array<string, string>>
  */
 class TableNode implements ArgumentInterface, IteratorAggregate
 {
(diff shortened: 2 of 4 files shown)


src/Node/ExampleNode.php
```diff
@@ -198,14 +198,16 @@ protected function replaceArgumentsTokens(array $arguments)
      */
     protected function replaceTableArgumentTokens(TableNode $argument)
     {
-        $table = $argument->getTable();
-        foreach ($table as $line => $row) {
-            foreach (array_keys($row) as $col) {
-                $table[$line][$col] = $this->replaceTextTokens($table[$line][$col]);
```
stof, line 204: I rewrote this by building a new array from scratch instead of copying the array (due to copy on write) and then overriding each value in place, because phpstan is not smart enough to identify that the `$col` key is an existing key of the list here, and so it preserves the fact that the array is a list.

src/Node/TableNode.php
```diff
@@ -96,11 +98,11 @@ public static function fromList(array $list)
             throw new NodeException('List is not a one-dimensional array.');
         }
 
-        array_walk($list, function (&$item) {
-            $item = [$item];
-        });
+        $table = array_map(function ($item) {
```
stof, line 101: using `array_map` is a lot easier to analyze by phpstan than changing array values by reference with `array_walk` (for which phpstan basically gives up)

tests/Node/TableNodeTest.php
```diff
@@ -333,9 +348,10 @@ public function testGetTableFromListWithMultidimensionalArrayArgument(): void
             new NodeException('List is not a one-dimensional array.')
         );
 
+        // @phpstan-ignore argument.type (we are explicitly testing an invalid instantiation)
         TableNode::fromList([
-            [1, 2, 3],
-            [4, 5, 6],
+            ['1', '2', '3'],
+            ['4', '5', '6'],
```
stof, line 354: using string values here inside the array corresponds to using `fromList` with the type expected by the constructor, which I find clearer (any nested array is rejected anyway).

acoulton (approved):
Thanks @stof

src/Node/TableNode.php
stof: For some reason, I though that arrow functions in PHP were much more recent that they actually are.

acoulton: I've done that before :laughing:

PR #341, opened 2025-05-22
change: Ignore content after table right-hand `|` (instead of throwing)

acoulton:
For parity with cucumber/gherkin - their `extra_table_content` testdata notes that "It is not recommended to use this feature, but it is how the implementation currently works."

The implementation is taken from the equivalent code in cucumber/gherkin StringGherkinLine.

This should be BC for end-users - previously we treated table rows that did not end with a `|` as plain text, which in turn triggered a ParserException as there is nowhere that plain text can immediately follow a table row.

Now these will instead be parsed, and just the right-hand "column" / content will be ignored.


src/Lexer.php
```diff
@@ -580,14 +586,20 @@ protected function scanTableRow()
...
             return null;
         }
 
+        $rawColumns = preg_split(self::CELL_PATTERN, $line);
+        assert($rawColumns !== false);
+
+        // Safely remove elements before the first and last separators
+        array_shift($rawColumns);
+        array_pop($rawColumns);
+
```
acoulton, line 601: This is taken from [cucumber/gherkin's PHP implementation](https://github.com/cucumber/gherkin/blob/679a87e21263699c15ea635159c6cda60f64af3b/php/src/StringGherkinLine.php#L75-L79).

We will probably want to borrow more of that in due course to solve the table escaping issues, but they need to be behind a feature flag to avoid a BC break.

stof (approved):

PR #344, opened 2025-05-22
Improve type-hinting of the lexer and parser

uuf6429:
Extracted/recovered changes from my phpstan-level-9 PR that:
1. improves type hinting of the parser, lexer and by coincidence, the keywords interface.
2. fixes a few php-level-9 issues

### DIFF
--- a/src/Keywords/KeywordsInterface.php
+++ b/src/Keywords/KeywordsInterface.php
@@ -14,10 +14,6 @@ namespace Behat\Gherkin\Keywords;
  * Keywords holder interface.
  *
  * @author Konstantin Kudryashov <ever.zet@gmail.com>
- *
- * @phpstan-type TLanguage string
- * @phpstan-type TKeywordsArray array{feature: string, background: string, scenario: string, scenario_outline: string, examples: string, given: string, when: string, then: string, and: string, but: string}
- * @phpstan-type TMultiLanguageKeywords array<TLanguage, TKeywordsArray>
  */
 interface KeywordsInterface
 {
--- a/src/Lexer.php
+++ b/src/Lexer.php
@@ -27,7 +27,17 @@ use function assert;
  *
  * @final since 4.15.0
  *
- * @phpstan-type TToken array{type: string, line: int, value: string|null, deferred: bool, tags?: list<string>, keyword?: string, keyword_type?: string, text?: string, indent?: int, columns?: list<string>}
+ * @phpstan-type TStepKeywordsType 'Given'|'When'|'Then'|'And'|'But'
+ * @phpstan-type TGeneralKeywordsType 'Feature'|'Background'|'Scenario'|'Outline'|'Examples'|'Step'
+ * @phpstan-type TKeywordsType TGeneralKeywordsType|TStepKeywordsType
+ * @phpstan-type TTokenType 'Text'|'Comment'|'EOS'|'Newline'|'PyStringOp'|'TableRow'|'Tag'|'Language'|TGeneralKeywordsType
+ * @phpstan-type TToken TStringValueToken|TNullValueToken|TKeywordToken|TStepToken|TTagToken|TTableRowToken
+ * @phpstan-type TStringValueToken array{type: TTokenType, value: string, line: int, deferred: bool}
+ * @phpstan-type TNullValueToken array{type: TTokenType, value: null, line: int, deferred: bool}
+ * @phpstan-type TKeywordToken array{type: TTokenType, value: null|string, line: int, deferred: bool, keyword: string, indent: int}
+ * @phpstan-type TStepToken array{type: TTokenType, value: string, line: int, deferred: bool, keyword_type: string, text: string}
+ * @phpstan-type TTagToken array{type: TTokenType, value: null, line: int, deferred: bool, tags: list<string>}
+ * @phpstan-type TTableRowToken array{type: TTokenType, value: null, line: int, deferred: bool, columns: list<string>}
  */
 class Lexer
 {
@@ -37,6 +47,7 @@ class Lexer
      * @see https://github.com/cucumber/gherkin/blob/679a87e21263699c15ea635159c6cda60f64af3b/php/src/StringGherkinLine.php#L14
      */
     private const CELL_PATTERN = '/(?<!\\\\)(?:\\\\{2})*\K\\|/u';
+
     private readonly DialectProviderInterface $dialectProvider;
     private GherkinDialect $currentDialect;
     private GherkinCompatibilityMode $compatibilityMode = GherkinCompatibilityMode::LEGACY;
@@ -52,7 +63,7 @@ class Lexer
     /**
      * A cache of keyword types associated with each keyword.
      *
-     * @var array<string, non-empty-list<string>>|null
+     * @phpstan-var array<string, non-empty-list<TStepKeywordsType>>|null
      */
     private ?array $stepKeywordTypesCache = null;
     /**
@@ -232,12 +243,13 @@ class Lexer
     /**
      * Constructs token with specified parameters.
      *
-     * @param string $type Token type
      * @param string|null $value Token value
+     *
+     * @phpstan-param TTokenType $type Token type
      *
      * @return array
      *
-     * @phpstan-return TToken
+     * @phpstan-return ($value is non-empty-string ? TStringValueToken : TNullValueToken)
      */
     public function takeToken(string $type, ?string $value = null)
     {
@@ -290,7 +302,7 @@ class Lexer
     }
 
     /**
-     * Returns stashed token or null if hasn't.
+     * Returns stashed token or null if there isn't one.
      *
      * @return array|null
      *
@@ -305,7 +317,7 @@ class Lexer
     }
 
     /**
-     * Returns deferred token or null if hasn't.
+     * Returns deferred token or null if there isn't one.
      *
      * @return array|null
      *
@@ -353,11 +365,12 @@ class Lexer
      * Scans for token with specified regex.
      *
      * @param string $regex Regular expression
-     * @param string $type Expected token type
+     *
+     * @phpstan-param TTokenType $type Expected token type
      *
      * @return array|null
      *
-     * @phpstan-return TToken|null
+     * @phpstan-return TStringValueToken|null
      */
     protected function scanInput(string $regex, string $type)
     {
@@ -375,11 +388,12 @@ class Lexer
      * Scans for token with specified keywords.
      *
      * @param string $keywords Keywords (separated by "|")
-     * @param string $type Expected token type
+     *
+     * @phpstan-param TTokenType $type Expected token type
      *
      * @return array|null
      *
-     * @phpstan-return TToken|null
+     * @phpstan-return TKeywordToken|null
      *
      * @deprecated
      */
@@ -421,7 +435,9 @@ class Lexer
     /**
      * @param list<string> $keywords
      *
-     * @phpstan-return TToken|null
+     * @phpstan-param TTokenType $type
+     *
+     * @phpstan-return TKeywordToken|null
      */
     private function scanTitleLine(array $keywords, string $type): ?array
     {
@@ -449,7 +465,7 @@ class Lexer
      *
      * @return array|null
      *
-     * @phpstan-return TToken|null
+     * @phpstan-return TNullValueToken|null
      */
     protected function scanEOS()
     {
@@ -463,7 +479,7 @@ class Lexer
     /**
      * Returns a regex matching the keywords for the provided type.
      *
-     * @param string $type Keyword type
+     * @phpstan-param TKeywordsType $type Keyword type
      *
      * @return string
      *
@@ -502,7 +518,7 @@ class Lexer
      *
      * @return array|null
      *
-     * @phpstan-return TToken|null
+     * @phpstan-return TKeywordToken|null
      */
     protected function scanFeature()
     {
@@ -529,7 +545,7 @@ class Lexer
      *
      * @return array|null
      *
-     * @phpstan-return TToken|null
+     * @phpstan-return TKeywordToken|null
      */
     protected function scanBackground()
     {
@@ -549,7 +565,7 @@ class Lexer
      *
      * @return array|null
      *
-     * @phpstan-return TToken|null
+     * @phpstan-return TKeywordToken|null
      */
     protected function scanScenario()
     {
@@ -570,7 +586,7 @@ class Lexer
      *
      * @return array|null
      *
-     * @phpstan-return TToken|null
+     * @phpstan-return TKeywordToken|null
      */
     protected function scanOutline()
     {
@@ -591,7 +607,7 @@ class Lexer
      *
      * @return array|null
      *
-     * @phpstan-return TToken|null
+     * @phpstan-return TKeywordToken|null
      */
     protected function scanExamples()
     {
@@ -611,7 +627,7 @@ class Lexer
      *
      * @return array|null
      *
-     * @phpstan-return TToken|null
+     * @phpstan-return TStepToken|null
      */
     protected function scanStep()
     {
@@ -652,7 +668,7 @@ class Lexer
      *
      * @return array|null
      *
-     * @phpstan-return TToken|null
+     * @phpstan-return TNullValueToken|null
      */
     protected function scanPyStringOp()
     {
@@ -689,7 +705,7 @@ class Lexer
      *
      * @return array|null
      *
-     * @phpstan-return TToken|null
+     * @phpstan-return TStringValueToken|null
      */
     protected function scanPyStringContent()
     {
@@ -709,7 +725,7 @@ class Lexer
      *
      * @return array|null
      *
-     * @phpstan-return TToken|null
+     * @phpstan-return TTableRowToken|null
      */
     protected function scanTableRow()
     {
@@ -749,7 +765,7 @@ class Lexer
      *
      * @return array|null
      *
-     * @phpstan-return TToken|null
+     * @phpstan-return TTagToken|null
      */
     protected function scanTags()
     {
@@ -779,7 +795,7 @@ class Lexer
      *
      * @return array|null
      *
-     * @phpstan-return TToken|null
+     * @phpstan-return TStringValueToken|null
      */
     protected function scanLanguage()
     {
@@ -816,7 +832,7 @@ class Lexer
      *
      * @return array|null
      *
-     * @phpstan-return TToken|null
+     * @phpstan-return TStringValueToken|null
      */
     protected function scanComment()
     {
@@ -840,7 +856,7 @@ class Lexer
      *
      * @return array|null
      *
-     * @phpstan-return TToken|null
+     * @phpstan-return TNullValueToken|null
      */
     protected function scanNewline()
     {
@@ -859,7 +875,7 @@ class Lexer
      *
      * @return array
      *
-     * @phpstan-return TToken
+     * @phpstan-return TStringValueToken
      */
     protected function scanText()
     {
@@ -873,6 +889,8 @@ class Lexer
      * Returns step type keyword (Given, When, Then, etc.).
      *
      * @param string $native Step keyword in provided language
+     *
+     * @phpstan-return TStepKeywordsType
      */
     private function getStepKeywordType(string $native): string
     {
@@ -901,6 +919,8 @@ class Lexer
 
     /**
      * @param list<string> $keywords
+     *
+     * @phpstan-param TStepKeywordsType $type
      */
     private function addStepKeywordTypes(array $keywords, string $type): void
     {
--- a/src/Lexer.php
+++ b/src/Lexer.php
@@ -27,10 +27,9 @@ use function assert;
  *
  * @final since 4.15.0
  *
- * @phpstan-type TStepKeywordsType 'Given'|'When'|'Then'|'And'|'But'
- * @phpstan-type TGeneralKeywordsType 'Feature'|'Background'|'Scenario'|'Outline'|'Examples'|'Step'
- * @phpstan-type TKeywordsType TGeneralKeywordsType|TStepKeywordsType
- * @phpstan-type TTokenType 'Text'|'Comment'|'EOS'|'Newline'|'PyStringOp'|'TableRow'|'Tag'|'Language'|TGeneralKeywordsType
+ * @phpstan-type TStepKeyword 'Given'|'When'|'Then'|'And'|'But'
+ * @phpstan-type TTitleKeyword 'Feature'|'Background'|'Scenario'|'Outline'|'Examples'|'Step'
+ * @phpstan-type TTokenType 'Text'|'Comment'|'EOS'|'Newline'|'PyStringOp'|'TableRow'|'Tag'|'Language'|TTitleKeyword
  * @phpstan-type TToken TStringValueToken|TNullValueToken|TKeywordToken|TStepToken|TTagToken|TTableRowToken
  * @phpstan-type TStringValueToken array{type: TTokenType, value: string, line: int, deferred: bool}
  * @phpstan-type TNullValueToken array{type: TTokenType, value: null, line: int, deferred: bool}
@@ -63,7 +62,7 @@ class Lexer
     /**
      * A cache of keyword types associated with each keyword.
      *
-     * @phpstan-var array<string, non-empty-list<TStepKeywordsType>>|null
+     * @phpstan-var array<string, non-empty-list<TStepKeyword>>|null
      */
     private ?array $stepKeywordTypesCache = null;
     /**
@@ -378,6 +377,8 @@ class Lexer
             return null;
         }
 
+        assert($matches[1] !== '');
+
         $token = $this->takeToken($type, $matches[1]);
         $this->consumeLine();
 
@@ -479,7 +480,7 @@ class Lexer
     /**
      * Returns a regex matching the keywords for the provided type.
      *
-     * @phpstan-param TKeywordsType $type Keyword type
+     * @phpstan-param TTitleKeyword|TStepKeyword $type Keyword type
      *
      * @return string
      *
@@ -652,6 +653,7 @@ class Lexer
         $text = ltrim(mb_substr($trimmedLine, mb_strlen($matchedKeyword)));
 
         $nodeKeyword = $this->compatibilityMode->shouldRemoveStepKeywordSpace() ? trim($matchedKeyword) : $matchedKeyword;
+        assert($nodeKeyword !== '');
 
         $token = $this->takeToken('Step', $nodeKeyword);
         $token['keyword_type'] = $this->getStepKeywordType($matchedKeyword);
@@ -715,7 +717,7 @@ class Lexer
 
         $token = $this->scanText();
         // swallow trailing spaces
-        $token['value'] = preg_replace('/^\s{0,' . $this->pyStringSwallow . '}/u', '', $token['value'] ?? '');
+        $token['value'] = (string) preg_replace('/^\s{0,' . $this->pyStringSwallow . '}/u', '', $token['value'] ?? '');
 
         return $token;
     }
@@ -875,7 +877,7 @@ class Lexer
      *
      * @return array
      *
-     * @phpstan-return TStringValueToken
+     * @phpstan-return TStringValueToken|TNullValueToken
      */
     protected function scanText()
     {
@@ -890,7 +892,7 @@ class Lexer
      *
      * @param string $native Step keyword in provided language
      *
-     * @phpstan-return TStepKeywordsType
+     * @phpstan-return TStepKeyword
      */
     private function getStepKeywordType(string $native): string
     {
@@ -920,7 +922,7 @@ class Lexer
     /**
      * @param list<string> $keywords
      *
-     * @phpstan-param TStepKeywordsType $type
+     * @phpstan-param TStepKeyword $type
      */
     private function addStepKeywordTypes(array $keywords, string $type): void
     {
--- a/src/Lexer.php
+++ b/src/Lexer.php
@@ -28,15 +28,15 @@ use function assert;
  * @final since 4.15.0
  *
  * @phpstan-type TStepKeyword 'Given'|'When'|'Then'|'And'|'But'
- * @phpstan-type TTitleKeyword 'Feature'|'Background'|'Scenario'|'Outline'|'Examples'|'Step'
- * @phpstan-type TTokenType 'Text'|'Comment'|'EOS'|'Newline'|'PyStringOp'|'TableRow'|'Tag'|'Language'|TTitleKeyword
- * @phpstan-type TToken TStringValueToken|TNullValueToken|TKeywordToken|TStepToken|TTagToken|TTableRowToken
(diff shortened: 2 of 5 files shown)


src/Keywords/KeywordsInterface.php
```diff
@@ -18,13 +18,17 @@
...
+ * @phpstan-type TGeneralKeywordsType 'Feature'|'Background'|'Scenario'|'Outline'|'Examples'|'Step'
+ * @phpstan-type TKeywordsType TGeneralKeywordsType|TStepKeywordsType
  */
 interface KeywordsInterface
 {
     /**
      * Sets keywords holder language.
      *
-     * @param string $language Language name
+     * @param TLanguage $language Language name
```
stof, line 31: please use the `phpstan-` prefixed tag when using type aliases

src/Keywords/KeywordsInterface.php
```diff
@@ -33,77 +37,77 @@ public function setLanguage($language);
     /**
      * Returns Feature keywords (separated by "|").
      *
-     * @return string
```
stof, line 36: should be kept to indicate the native return type we'll add in the future.

src/Keywords/KeywordsInterface.php
```diff
@@ -18,13 +18,17 @@
  * @phpstan-type TLanguage string
  * @phpstan-type TKeywordsArray array{feature: string, background: string, scenario: string, scenario_outline: string, examples: string, given: string, when: string, then: string, and: string, but: string}
  * @phpstan-type TMultiLanguageKeywords array<TLanguage, TKeywordsArray>
+ * @phpstan-type TKeywordsString string
+ * @phpstan-type TStepKeywordsType 'Given'|'When'|'Then'|'And'|'But'
```
uuf6429, line 22: As I understand, these are non-translatable keyword "identifiers" that are later on mapped to translations. If that is the case, then this should be fine.

src/Lexer.php
```diff
@@ -21,7 +21,18 @@
...
  * @author Konstantin Kudryashov <ever.zet@gmail.com>
  *
- * @phpstan-type TToken array{type: string, line: int, value: string|null, deferred: bool, tags?: list<string>, keyword?: string, keyword_type?: string, text?: string, indent?: int, columns?: list<string>}
+ * @phpstan-import-type TKeywordsString from KeywordsInterface
+ * @phpstan-import-type TKeywordsType from KeywordsInterface
+ * @phpstan-import-type TStepKeywordsType from KeywordsInterface
+ * @phpstan-import-type TGeneralKeywordsType from KeywordsInterface
+ *
+ * @phpstan-type TTokenType 'Text'|'Comment'|'EOS'|'Newline'|'PyStringOp'|'TableRow'|'Tag'|'Language'|TGeneralKeywordsType
+ * @phpstan-type TTokenValue int|string|null
```
stof, line 30: the token value cannot be a string anymore since #338

src/Lexer.php
```diff
@@ -288,37 +295,35 @@ protected function getDeferredToken()
...
+            ?? $this->scanScenario()
+            ?? $this->scanBackground()
+            ?? $this->scanOutline()
+            ?? $this->scanExamples()
+            ?? $this->scanFeature()
+            ?? $this->scanTags()
+            ?? $this->scanTableRow()
+            ?? $this->scanNewline()
+            ?? $this->scanText()
+            ?? throw new LogicException('End of token stream encountered unexpectedly.');
```
stof, line 318: scanText cannot return null.

src/Lexer.php
```diff
@@ -21,7 +21,18 @@
...
+ * @phpstan-import-type TStepKeywordsType from KeywordsInterface
+ * @phpstan-import-type TGeneralKeywordsType from KeywordsInterface
+ *
+ * @phpstan-type TTokenType 'Text'|'Comment'|'EOS'|'Newline'|'PyStringOp'|'TableRow'|'Tag'|'Language'|TGeneralKeywordsType
+ * @phpstan-type TTokenValue int|string|null
+ * @phpstan-type TToken array{type: TTokenType, value: TTokenValue, line: int, deferred: bool}
+ * @phpstan-type TKeywordToken array{type: TTokenType, value: TTokenValue, line: int, deferred: bool, keyword: string, indent: int}
+ * @phpstan-type TStepToken array{type: TTokenType, value: TTokenValue, line: int, deferred: bool, keyword_type: string, text: string}
+ * @phpstan-type TTagToken array{type: TTokenType, value: TTokenValue, line: int, deferred: bool, tags: list<string>}
+ * @phpstan-type TTableRowToken array{type: TTokenType, value: TTokenValue, line: int, deferred: bool, columns: list<string>}
```
uuf6429, line 35: Having granular token alias types allows us to better describe what should be returned in the relevant methods and what to expect when calling them - that makes some assertions redundant.

src/Lexer.php
```diff
@@ -720,9 +695,7 @@ protected function scanNewline()
     /**
      * Scans text from input & returns it if found.
      *
-     * @return array
-     *
-     * @phpstan-return TToken
+     * @phpstan-return TToken|null
```
stof, line 698: should be reverted. This method does not return null

src/Parser.php
```diff
@@ -201,10 +218,8 @@ protected function parseExpression()
     protected function parseFeature()
     {
         $token = $this->expectTokenType('Feature');
-        \assert(\array_key_exists('keyword', $token));
-        \assert(\array_key_exists('indent', $token));
 
-        $title = trim($token['value'] ?? '');
+        $title = trim((string)($token['value'] ?? ''));
```
stof, line 222: All those changes should be reverted

src/Lexer.php
```diff
@@ -42,11 +53,17 @@ class Lexer
...
      */
     private array $keywordsCache = [];
     /**
-     * @var array<string, list<string>>
+     * @var array{}|array{
+     *     Given: list<string>,
+     *     When: list<string>,
+     *     Then: list<string>,
+     *     And: list<string>,
+     *     But: list<string>,
```
uuf6429, line 65: From the code I understood that this array should have that exact structure, which is why I opted to go for a strict approach instead.

tests/ParserTest.php
```diff
@@ -132,7 +133,8 @@ public static function parserErrorDataProvider(): iterable
 
         yield 'invalid content encoding' => [
             'expectedException' => new ParserException('Lexer exception "Feature file is not in UTF8 encoding" thrown for file /fake.feature'),
-            'featureText' => mb_convert_encoding('🔥 Все буде добре 🔥', 'EUC-JP', 'UTF-8'),
+            'featureText' => mb_convert_encoding('🔥 Все буде добре 🔥', 'EUC-JP', 'UTF-8')
+                ?: throw new RuntimeException('Could not encode string data for testing successfully'),
```
stof, line 137: I suggest reverting this. It will be solved by https://github.com/phpstan/phpstan-src/pull/4010 which makes the phpstan analysis smarter.

src/Lexer.php
```diff
@@ -147,8 +164,6 @@ public function getLanguage()
     /**
      * Returns next token or previously stashed one.
      *
-     * @return array
```
uuf6429, line 150: This and similar cases are still to be discussed.

src/Parser.php
```diff
@@ -104,22 +112,31 @@ public function parse($input, $file = null)
...
+     *                 : ($type is TGeneralKeywordsType
+     *                     ? TKeywordToken
+     *                     : TToken)))
+     * )
      *
      * @throws ParserException
      */
     protected function expectTokenType($type)
     {
-        $types = (array) $type;
```
uuf6429, line 113: I think this change should be safebecause:
1. phpdoc param was always string and it never said anything about possibly passing an array
2. currently we never pass an array - only a single string

src/Keywords/KeywordsInterface.php
uuf6429: Just to be clear, is the point of this just to have the correct type for the future (and should therefore only add simple types)?

So for example:
```php
    /**
     * @phpstan-return TMultiLanguageKeywords
     */
    abstract protected static function getKeywordsArray(): array;
```
Should I add `@return array<string, whatever>` in that case?

stof: I would keep the simple case matching the future native type (which would be removed once we add a native return type). There is no point defining a precise type without type alias in `@return` while using an equivalent type alias in `@phpstan-return`. It only doubles the maintenance work (with phpstan checking only one of them)

src/Lexer.php
```diff
@@ -567,19 +552,17 @@ protected function scanPyStringContent()
             return null;
         }
 
-        $token = $this->scanText();
+        $token = $this->scanText() ?? throw new LogicException('End of token stream encountered unexpectedly.');
```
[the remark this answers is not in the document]
stof: should be reverted as scanText does not return null

src/Lexer.php
```diff
@@ -21,7 +21,18 @@
...
+ * @phpstan-import-type TKeywordsType from KeywordsInterface
+ * @phpstan-import-type TStepKeywordsType from KeywordsInterface
+ * @phpstan-import-type TGeneralKeywordsType from KeywordsInterface
+ *
+ * @phpstan-type TTokenType 'Text'|'Comment'|'EOS'|'Newline'|'PyStringOp'|'TableRow'|'Tag'|'Language'|TGeneralKeywordsType
+ * @phpstan-type TTokenValue int|string|null
+ * @phpstan-type TToken array{type: TTokenType, value: TTokenValue, line: int, deferred: bool}
+ * @phpstan-type TKeywordToken array{type: TTokenType, value: TTokenValue, line: int, deferred: bool, keyword: string, indent: int}
+ * @phpstan-type TStepToken array{type: TTokenType, value: TTokenValue, line: int, deferred: bool, keyword_type: string, text: string}
+ * @phpstan-type TTagToken array{type: TTokenType, value: TTokenValue, line: int, deferred: bool, tags: list<string>}
```
stof, line 34: the value or a tag token is always a string. Same for the step token.

If we define more precise types for the different tokens, we could as well do it fully (replacing all the `\assert(\is_string($token['value']))` I added in #342 for such cases.

src/Lexer.php
acoulton: I'm not certain about this. Once we get to the public interface of the class, we basically have to go back to saying it can return any of the supported token types - the methods that generate specific types are only ever called internally.

So the Parser then still has to differentiate based on the values in the array that it gets. Theoretically also the Parser accepts any Lexer (the class is not final) so there is no actual guarantee that the tokens we get at runtime follow the virtual types defined here.

I think I would probably prefer keeping the `TToken` definition fairly generic and using runtime assertions - but I am newer to working with PHPStan, so willing to be challenged.

However as @stof says **if** we do define specific types then we should be fully explicit. Not just about the value type, but also for example the `type` key for each of these tokens is actually a fixed value - you can't get a `TTagToken` with a `type` value of `Comment` etc.

src/Lexer.php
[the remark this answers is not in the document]
acoulton: I would much prefer to have changes to actual runtime code committed separately to things that are just phpstan / phpdoc comments, with a suitable message. It is much easier to review that way, much less risk that something slips through in a big diff, and also makes it easier to deal with any changes following code review.

src/Parser.php
```diff
@@ -267,10 +282,8 @@ protected function parseFeature()
     protected function parseBackground()
     {
         $token = $this->expectTokenType('Background');
-        \assert(\array_key_exists('keyword', $token));
-        \assert(\array_key_exists('indent', $token));
 
-        $title = trim($token['value'] ?? '');
+        $title = trim((string) ($token['value'] ?? ''));
```
acoulton, line 286: Can be reverted, as other cases.

src/Parser.php
```diff
@@ -331,13 +344,11 @@ protected function parseOutline()
...
     /**
-     * @phpstan-param TToken $token
+     * @phpstan-param TKeywordToken $token
      */
     private function parseScenarioOrOutlineBody(array $token): OutlineNode|ScenarioNode
     {
-        \assert(\array_key_exists('keyword', $token));
-        \assert(\array_key_exists('indent', $token));
-        $title = trim($token['value'] ?? '');
+        $title = trim((string) ($token['value'] ?? ''));
```
acoulton, line 351: Can be reverted, as other cases.

src/Parser.php
```diff
@@ -447,14 +458,6 @@ private function validateAndGetNextTaggedNodeType(): string
...
     {
         $token = $this->expectTokenType('Step');
-        \assert(\is_string($token['value']));
-        \assert(\array_key_exists('keyword_type', $token));
-        \assert(\array_key_exists('text', $token));
-
-        $keyword = $token['value'];
-        $keywordType = $token['keyword_type'];
-        $text = trim($token['text']);
-        $line = $token['line'];
```
acoulton, line 461: I'd prefer not to have this variable inlining in the same commit diff as everything else.

I also think we've lost something here - `$token['value']` must always be present for a step token. Casting to a string, instead of the previous `assert(is_string` means a null will now be happily accepted when that should be invalid?

src/Parser.php
```diff
@@ -569,7 +570,7 @@ protected function popTags()
     /**
      * Checks the tags fit the required format.
      *
-     * @param string[] $tags
+     * @param array<array-key, string> $tags
```
acoulton, line 573: Fairly sure this should be a list?

src/Lexer.php
stof: @acoulton indeed. I edited my comment

src/Keywords/KeywordsInterface.php
stof, line 22: I'm wondering whether we should use such a type alias here or define a `StepKeywordType` class with constants for each type and then use `StepKeywordType::*` as type in place needing it (while also being able to use the constants instead of literal strings in places

acoulton: I was actually wondering whether we should introduce an enum for this now. 

We may be able to use the actual enum internally in some places, and where we still need to use strings the phpstan `value-of<StepKeywordType>` would work the same as `StepKeywordType::*`. I think there are also some methods with parameters that could be `StepKeywordType|value-of<StepKeywordType>` where we could safely accept either and convert internally as required. 

That would get us phpstan and runtime benefits from using an enum now without a BC break, and provide a cleaner path for making more use of the enum in a future major?

src/Lexer.php
```diff
@@ -42,11 +53,17 @@ class Lexer
     private int $lineNumber;
     private bool $eos;
     /**
-     * @var array<string, string>
+     * @phpstan-var array<TKeywordsType, TKeywordsString>
      */
     private array $keywordsCache = [];
     /**
-     * @var array<string, list<string>>
+     * @var array{}|array{
```
stof, line 60: This property is refactored entirely in #350

src/Lexer.php
uuf6429: Good catch 😮

src/Parser.php
uuf6429: Done: https://github.com/Behat/Gherkin/pull/352

src/Keywords/KeywordsInterface.php
```diff
@@ -18,13 +18,17 @@
  * @phpstan-type TLanguage string
  * @phpstan-type TKeywordsArray array{feature: string, background: string, scenario: string, scenario_outline: string, examples: string, given: string, when: string, then: string, and: string, but: string}
  * @phpstan-type TMultiLanguageKeywords array<TLanguage, TKeywordsArray>
+ * @phpstan-type TKeywordsString string
```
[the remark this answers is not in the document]
stof: The `@return array` makes a difference as explained in https://github.com/Behat/Gherkin/pull/333#issuecomment-2902505496
It helps with tooling reporting deprecations when a child class overrides the method, to push them to add a native return type (as child classes must add them before parent classes). This `@return array` would be removed once we have a native return type instead.

> Having alias types for primitives helps correlating values, perhaps facilitating future type checks with regards to PHPStan

I don't see any future where phpstan would treat its type aliases as opaque types instead of actual type aliases like today, as this is a totally different feature with a totally different usage (by definition of being opaque, a value of an opaque type can never be created outside the code owning that opaque type and so knowing what is its underlying implementation, and only that owner code can also act on such value for the same reason). And most of the primitive aliases you are adding here don't correspond to a case of such opaque type but to a public API involving a string.

> indicate a potential for a replacement with value objects (or similar).

Most of those primitive type aliases you are adding end up appearing as array keys, where value objects are forbidden in PHP, which actually indicates they cannot be replaced.

src/Lexer.php
```diff
@@ -21,7 +21,19 @@
  *
  * @author Konstantin Kudryashov <ever.zet@gmail.com>
  *
- * @phpstan-type TToken array{type: string, line: int, value: string|null, deferred: bool, tags?: list<string>, keyword?: string, keyword_type?: string, text?: string, indent?: int, columns?: list<string>}
+ * @phpstan-import-type TLanguage from KeywordsInterface
+ * @phpstan-import-type TKeywordsString from KeywordsInterface
+ * @phpstan-import-type TKeywordsType from KeywordsInterface
+ * @phpstan-import-type TStepKeywordsType from KeywordsInterface
```
stof, line 27: The StepKeywordType exposed in the `StepNode::getKeywordType` is not related to a string literal belonging to KeywordsInterface. They are a different concept (even if somewhat related).

Note that in `cucumber/messages`, they also have a concept of step keyword type (we don't test equivalence in our CompatibilityTest because our concept is actually a mix between their step keyword type and their pickle step keyword type, because we don't have a separate concept of pickle between the AST and the Behat runner), and they consider `And` and `But` to both be of the `Conjunction` type instead of being separate types once parsed.

src/Lexer.php
```diff
@@ -21,7 +21,19 @@
...
+ * @phpstan-import-type TGeneralKeywordsType from KeywordsInterface
```
stof, line 28: Getting this `TGeneralKeywordsType` from KeywordsInterface is also creating a dependency that does not exist in practice. The token types are a concept of the Lexer, not of the KeywordsInterface.
A good proof is that you define this type alias in KeywordsInterface but you never actually use it in this file. This is a very good indication that the type alias is defined in a place where it does not belong.

src/Parser.php
```diff
@@ -104,22 +113,31 @@ public function parse($input, $file = null)
...
+     * @phpstan-return (
+     *     $type is 'TableRow'
+     *         ? TTableRowToken
+     *         : ($type is 'Tag'
+     *             ? TTagToken
+     *             : ($type is 'Step'
+     *                 ? TStepToken
+     *                 : ($type is TGeneralKeywordsType
+     *                     ? TKeywordToken
+     *                     : TNullValueToken|TStringValueToken)))
```
stof, line 125: `Text` is always a `TStringValueToken`

src/Parser.php
stof: this particular method does not care about keys, but we indeed always pass a list in practice in our own code.

I did not made the parameter stricter in my previous PR because this is a protected method. A child class could be passing a non-list value, and the implementation supports it without any change (as it ignores keys when looping anyway).

src/Keywords/KeywordsInterface.php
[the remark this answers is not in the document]
acoulton: Thanks for explaining @uuf6429 but I agree with @stof, I don't think the string type alias provides much benefit. I also think it could be more confusing for debugging / someone coming to the code new - it implies the methods get some sort of special value which is true in theory but in practice is from a very wide set of possibilities (or from a public API).

Please revert this aspect of your changes.

src/Lexer.php
```diff
@@ -42,20 +55,26 @@ class Lexer
...
+     *     When: list<string>,
+     *     Then: list<string>,
+     *     And: list<string>,
+     *     But: list<string>,
+     * }
      */
     private array $stepKeywordTypesCache = [];
     /**
-     * @phpstan-var list<TToken>
+     * @phpstan-var list<TNullValueToken|TStringValueToken>
```
acoulton, line 72: Sorry, I'm really not seeing the benefit of these changes. The deferred / stashed tokens can by definition be any type of token, with any of the possible additional array keys. `TToken` makes clear that we don't know what type of token we have (indeed deferred tokens are passed through the public API and could in theory be anything). `TNullValueToken|TStringValueToken` implies that it is somehow a subset?

I'd understand more if these were real classes where e.g. `TStepToken` extends `TStringValueToken` extends `TToken` - but they don't. IMO this is adding a lot of complexity (and changes and code reviews) as well as potentially false confidence, and I'm not sure what we gain in return?

src/Keywords/KeywordsInterface.php
acoulton: True that we can't return the enum at runtime in 4.x, however if we are making that class final we could easily do so in 5.x. - and we then only change internal API rather than also changing a (public) const into an enum?

src/Lexer.php
```diff
@@ -21,7 +21,19 @@
  *
  * @author Konstantin Kudryashov <ever.zet@gmail.com>
  *
- * @phpstan-type TToken array{type: string, line: int, value: string|null, deferred: bool, tags?: list<string>, keyword?: string, keyword_type?: string, text?: string, indent?: int, columns?: list<string>}
```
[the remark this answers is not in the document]
stof: Different projects can have different policies there. I don't think there is an established convention yet.

src/Keywords/KeywordsInterface.php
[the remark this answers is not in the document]
uuf6429: > I don't see any future where phpstan would treat its type aliases as opaque types instead of actual type aliases like today

That is not necessarily true. PHPStan already has specialized types with respect to strings: `non-empty-string`, string literals and so on. If a new type was defined `string-matching-pattern<THE_PATTERN>` it could be easily used with `TLanguage` (or we could have even defined `TLanguage` as `'en'|'fr'` etc) - at that point it would be a matter of redefining the type, not reintroducing it all over the place.

Introducing this later is possible, but a waste of time considering I had to check how these values are passed already... which I guess also wastes potential contributors' time.

> I also think it could be more confusing for debugging / someone coming to the code new

How can it be more confusing? We have 700+ occurrences of `string`, how can aliasing a few of them with a different name (where relevant) possibly make it worse? Consider cases like `$keywords` array, the key is the language. There is no documentation or a hint whatsoever that the key is (or has to be) a language, and as @stoff noted, cannot be solved in any other way because the key has to be a string.

It's extremely easy to see how the type is defined and anyone with a little bit of knowledge of PHPStan should catch on quick.

----

But anyway, I don't mind removing it... it indeed doesn't add a huge advantage at the moment.

src/Lexer.php
uuf6429: Are you suggesting that I undo this change?

src/Lexer.php
uuf6429: Define "these changes".. in some cases, such as this one, I was forced to change it because I had to split `TToken`.

Now that I think about it, maybe I should define `TToken` as `TNullValueToken|TStringValueToken|etc`?

src/Lexer.php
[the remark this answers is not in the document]
uuf6429: We've changed the definition of that alias anyway (by removing `int`), by the way.

src/Lexer.php
```diff
@@ -111,6 +132,9 @@ public function analyse($input, $language = 'en')
         $this->setLanguage($language);
     }
 
+    /**
+     * @phpstan-param TLanguage $language
+     */
```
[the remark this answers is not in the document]
uuf6429: Removed as discussed.

src/Lexer.php
uuf6429: Yep, I fixed those others. 👍

src/Lexer.php
uuf6429: I didn't touch `StepNode::getKeywordType()`... do you mean it should `@phpstan-return Lexer::TStepKeywordsType`?

src/Lexer.php
```diff
@@ -567,19 +552,17 @@ protected function scanPyStringContent()
             return null;
         }
 
-        $token = $this->scanText();
+        $token = $this->scanText() ?? throw new LogicException('End of token stream encountered unexpectedly.');
         // swallow trailing spaces
-        $token['value'] = preg_replace('/^\s{0,' . $this->pyStringSwallow . '}/u', '', $token['value'] ?? '');
+        $token['value'] = preg_replace('/^\s{0,' . $this->pyStringSwallow . '}/u', '', (string) ($token['value'] ?? ''));
```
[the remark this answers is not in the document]
uuf6429: Right you are. Looking forward to increasing PHPStan level, it would also catch pointless casts.

src/Lexer.php
uuf6429: I went back to that union approach, it should be cleaner and less wrong. :)

src/Lexer.php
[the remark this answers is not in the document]
uuf6429: I've reintroduced it as a union of all token types, since it makes sense in places like `getAdvanceToken()` etc, so I believe that resolves this conversation. Please reopen if you feel differently.

src/Keywords/KeywordsInterface.php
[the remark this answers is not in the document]
stof: The issue with type aliases is that they might help readability if used properly. But the tooling won't be able to catch mistakes in their usages (as `string` is still a `string`).

> There is no documentation or a hint whatsoever that the key is (or has to be) a language

Adding documentation on the structure of a cache property can still be done. Our phpdoc can have description.

> (or we could have even defined `TLanguage` as `'en'|'fr'` etc)

A language can actually be any non-empty string (and #350 uses `non-empty-string` as type for them in the new API).

src/Lexer.php
stof: I would say yes. It will avoid conflicts between the 2 PRs.

src/Parser.php
uuf6429: I changed it back to `array<array-key` ... Please reopen if there is any disagreement.

src/Lexer.php
```diff
@@ -31,6 +41,7 @@ class Lexer
      * @see https://github.com/cucumber/gherkin/blob/679a87e21263699c15ea635159c6cda60f64af3b/php/src/StringGherkinLine.php#L14
      */
     private const CELL_PATTERN = '/(?<!\\\\)(?:\\\\{2})*\K\\|/u';
+
```
acoulton, line 44: To avoid conflict with #350
```suggestion
```

src/Lexer.php
```diff
@@ -42,7 +53,7 @@ class Lexer
     private int $lineNumber;
     private bool $eos;
     /**
-     * @var array<string, string>
+     * @phpstan-var array<TKeywordsType, string>
```
acoulton, line 56: To avoid conflict with #350 where this prop is removed
```suggestion
     * @var array<string, string>
```

src/Lexer.php
```diff
@@ -338,11 +351,12 @@ protected function scanInput(string $regex, string $type)
      * Scans for token with specified keywords.
      *
      * @param string $keywords Keywords (separated by "|")
-     * @param string $type Expected token type
+     *
+     * @phpstan-param TTokenType $type Expected token type
```
acoulton, line 355: This method is deprecated and no longer used after #350, I think we can leave the types unchanged.

```suggestion
     * @param string $type Expected token type
```

src/Lexer.php
```diff
@@ -338,11 +351,12 @@ protected function scanInput(string $regex, string $type)
...
      *
      * @return array|null
      *
-     * @phpstan-return TToken|null
+     * @phpstan-return TKeywordToken|null
```
acoulton, line 359:
```suggestion
     * @phpstan-return TToken|null
```

src/Lexer.php
```diff
@@ -396,7 +410,7 @@ protected function scanEOS()
     /**
      * Returns keywords for provided type.
      *
-     * @param string $type Keyword type
+     * @phpstan-param TKeywordsType $type Keyword type
```
acoulton, line 413: This method also becomes deprecated and unused in #350
```suggestion
     * @param string $type Keyword type
```

acoulton (changes requested):
@uuf6429 thanks very much, this is looking great to me now.

I've just suggested a few small reverts to things that are about to change in #350, to avoid creating unnecessary conflicts. I am happy to apply those for you if you want.

src/Lexer.php
uuf6429: This thing about conflicts is getting a bit out of hand. I've had to solve a truckload of conflicts already, so that's kind of part of the job. I don't think it makes sense to anticipate all of this since there's a chance that PR won't be merged.

IMO we should merge either PR first and just solve the conflicts later. I also think it makes sense to apply changes to deprecated methods since they'll remain a part of the codebase.

uuf6429: PS: We are essentially wasting time on conflicts already, so we're not gaining anything.

acoulton: @uuf6429 to be fair, I'd think it was unreasonable if I'd asked you to fix the conflicts. Picking them out and proposing them as changes (and offering to commit them) is not that different to waiting and one of us resolving the conflicts at merge time. And in this specific suggestion, it's not just that it's a conflict but that the PR is adding a redundant blank line (left over from previous additions/reverts).

The other PR will definitely be merged, there were only a couple of small details left to tackle when I reviewed this PR this weekend, and now only one small point blocking it.

Up to you what you think is most efficient, I have no objection to committing the reverts & merging this, or you can wait and rebase once #350 is merged.

uuf6429: The blank line was intentional :D Usually I'm putting one between each set of types of symbols in a class (uses, constants, properties, methods) - with methods and sometimes properties having a blank line for each symbol.

uuf6429: I'll wait until the other is merged, because then I can be sure these new changes are applied correctly throughout.

src/Lexer.php
```diff
@@ -27,7 +27,17 @@
  *
  * @final since 4.15.0
  *
- * @phpstan-type TToken array{type: string, line: int, value: string|null, deferred: bool, tags?: list<string>, keyword?: string, keyword_type?: string, text?: string, indent?: int, columns?: list<string>}
+ * @phpstan-type TStepKeywordsType 'Given'|'When'|'Then'|'And'|'But'
+ * @phpstan-type TGeneralKeywordsType 'Feature'|'Background'|'Scenario'|'Outline'|'Examples'|'Step'
+ * @phpstan-type TKeywordsType TGeneralKeywordsType|TStepKeywordsType
```
stof, line 32: is this type actually used anywhere ?

uuf6429: On an unrelated note, I don't like the naming of `TGeneralKeywordsType`.
The best alternative I can find is `TNonStepKeywordsType` (I thought about "TopLevel", but technically only "Feature" is top level).

stof: I agree this type is not worth it, especially as its only usage is on a deprecated method (I would even keep it as simply `string` in that case as it is deprecated)

uuf6429: `TTitleKeywordType` sounds good to me!

acoulton (approved):
LGTM, thanks for all your work on this @uuf6429

stof (approved):

PR #345, opened 2025-05-23
Define and change phpdoc order coding style

uuf6429:
The current/default style doesn't handle the `@phpstan-` cases, which as I see, we will have more often.

This CS change ensures they stay uniform.

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -26,6 +26,17 @@ return (new PhpCsFixer\Config())
         ],
         'concat_space' => ['spacing' => 'one'],
         'phpdoc_align' => ['align' => 'left'],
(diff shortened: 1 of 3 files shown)


stof (approved):

uuf6429:
@acoulton can you give this a look please? I'd like to use it in the draft PR.

stof:
@uuf6429 you can submit the other PR with coding standards matching this rule even if the tooling does not enforce it. I don't think it creates a blockage.

acoulton (approved):

acoulton:
I've added this to the changelog for 4.14.0 in #346, I'd like to get that merged & released before we merge any more PRs as I think it's already a big enough release.

Issue #336, opened 2025-05-22
Tags are not parsed the same than in cucumber


stof:
cucumber/gherkin keeps the `@` in the tag, while we remove the `@` prefix

This is normalized in the CucumberNDJsonAstLoader used by our compatibility tests: https://github.com/Behat/Gherkin/blob/281e044300418a83be5b7dd435ba5de9197f0db2/src/Loader/CucumberNDJsonAstLoader.php#L78

Removing that transformation makes some tests fail

stof:
Changing this will impact the pretty formatter of Behat (which is currently looping over tags to re-add the prefix) and the TagFilter implementation (in Gherkin).

The pretty formatter of Behat could probably be updated to add the `@` only when not already present.

PR #349, opened 2025-05-23
feat: Introduce compatibility mode & fix feature whitespace

### LINKED ISSUE #209: Feature description whitespace does not match Cucumber

acoulton:
Not to be merged until after the 4.14.0 release.

This is the beginning of a concept for providing a "parser mode" to allow users to opt-in to newer parser behaviour that is compatible with cucumber/gherkin but may not be 100% BC.

I have started by using the flag to fix #209 - subsequent PRs will then tackle the other remaining cucumber-parity issues.

Very open to feedback on both the concept and naming of things!


src/GherkinCompatibilityMode.php
```diff
@@ -0,0 +1,26 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Behat\Gherkin;
+
+enum GherkinCompatibilityMode: string
+{
+    case LEGACY = 'legacy';
+
+    case GHERKIN_32 = 'gherkin-32';
```
acoulton, line 17: I have tied this to a gherkin version, because we may in future want to use similar logic to support new gherkin majors.

The downside is that unless we solve all the cucumber parity issues in the first release, the behaviour in `gherkin-32` mode will change in subsequent releases. That may be acceptable if we advertise to users that the `gherkin-32` mode is expected to be unstable for a few releases - they can always stick with the default legacy mode.

I don't want to expose individual feature flags to end users as that will hugely complicate our ability to test / maintain the parser in different configurations.

tests/Cucumber/CompatibilityTest.php
```diff
@@ -28,44 +28,75 @@
...
-        'rule_with_tag.feature' => 'Rule keyword not supported',
-        'tags.feature' => 'Rule keyword not supported',
-        'descriptions.feature' => 'Examples table descriptions not supported',
-        'descriptions_with_comments.feature' => 'Examples table descriptions not supported',
-        'feature_keyword_in_scenario_description.feature' => 'Scenario descriptions not supported',
-        'padded_example.feature' => 'Table padding is not trimmed as aggressively',
-        'spaces_in_language.feature' => 'Whitespace not supported around language selector',
-        'rule_without_name_and_description.feature' => 'Rule is wrongly parsed as Description',
-        'incomplete_background_2.feature' => 'Background descriptions not supported',
+        'legacy' => [
```
acoulton, line 44: It would be nice to directly use the enum case values as the array keys, but they're not supported in constant expressions in PHP 8.1.

src/Parser.php
```diff
@@ -218,12 +224,28 @@ protected function parseFeature()
...
+                    continue;
+                }
+
+                // It must be part of the feature description (text & newlines later in the document will be consumed as
+                // part of parsing Background / Scenario before execution returns to this loop).
+                $description .= $node;
+                if ($node !== "\n") {
+                    // Text nodes do not end with a newline, add one. The final trailing newline is rtrimmed below.
+                    $description .= "\n";
+                }
```
acoulton, line 245: This is a little convoluted because for a feature file like:

```gherkin
Feature: Extra table content
  Tables are delimited by pipes on both sides.
  Anything that isn't enclosed is not part of
  the table.

  It is not recommended to use this feature, but
  it is how the implementation currently works.
```

The Lexer gives us the following tokens (note that the empty line is represented by "\n", slightly inconsistently to the lines with text).

```
"  Tables are delimited by pipes on both sides."
"  Anything that isn't enclosed is not part of"
"  the table."
"\n"
"  It is not recommended to use this feature, but"
"  it is how the implementation currently works."
"\n"
```

src/GherkinCompatibilityMode.php
```diff
@@ -0,0 +1,26 @@
...
+
+    public function shouldRemoveFeatureDescriptionPadding(): bool
```
stof, line 19: should we mark all those methods as internal ?

To me, the only public API of the enum should be the list of cases.

stof (commented):
should the compatibility be exposed (through an `@internal` getter) in the FeatureNode, to all checking it in the TagFilter for instance when solving #336 ?

src/Parser.php
```diff
@@ -218,12 +224,28 @@ protected function parseFeature()
...
             $node = $this->parseExpression();
 
-            if (is_string($node)) {
+            if (is_string($node) && $this->compatibilityMode->shouldRemoveFeatureDescriptionPadding()) {
                 $text = preg_replace('/^\s{0,' . ($token['indent'] + 2) . '}|\s*$/', '', $node);
                 $description .= ($description !== null ? "\n" : '') . $text;
                 continue;
             }
 
+            if (is_string($node)) {
```
carlos-granados, line 233: The code inside this if statement only runs if we are not in legacy compatibility mode but I had to think twice to realize this. I think the logic would be clearer if we did:

```
if (is_string($node)) {
    if ($this->compatibilityMode->shouldRemoveFeatureDescriptionPadding()) {
        ...
        continue;
   }     
   if ($node === "\n" && $description === null) {
       ...
   }
   ....
   continue;
}
```

carlos-granados (commented):
@acoulton I think this is a good idea in the right direction, just added a small comment

src/GherkinCompatibilityMode.php
acoulton: I've added a note that behaviour will change.

src/GherkinCompatibilityMode.php
acoulton: Good suggestion.

I'd thought of leaving it public as it may be useful e.g. for Behat to know whether padding was removed or not (to then know whether to add it back).

But actually I think we should try to solve any of these downstream without reference to the parsing mode (e.g. we can just look at whether there is / is not whitespace at the start of the first line).

Then it will be easier to remove these if & when they're no longer required.

acoulton:
> should the compatibility be exposed (through an @internal getter) in the FeatureNode, to all checking it in the TagFilter for instance when solving https://github.com/Behat/Gherkin/issues/336 ?

That's a good idea, it could be useful - but we might be able to solve things other ways (e.g. for that one just by looking for the `@` or not at the start of the tag string).

I think I'd prefer to wait until we definitely need it, rather than adding it now?

carlos-granados (approved):
Looking good for me now, thanks @acoulton

stof (approved):

PR #350, opened 2025-05-23
feat: Introduce a DialectProviderInterface matching the modern cucumber API

### LINKED ISSUE #156: Upstream gherkin breaks if language is not recognised
Cucumber repo has an example with an unknown language, which should cause a parse error. We default back to `en` I believe

### LINKED ISSUE #203: The Keywords API is based on cucumber gherkin 2
The syntax using `|` delimiters and `<` to decide whether a space is necessary or no at the end of the keyword is a legacy from cucumber/gherkin 2. In their 3.0 version, they changed their language system to use a list of keywords instead of a string needing to be split, and they include the space itself in the translation when it is necessary (i.e. in most languages).

Our update_i18n script is currently converting back to the cucumber/gherkin 2 format to create our own i18n.php to be compatible with our own keyboard API. The Lexer then need to undo most of these changes (and might not actually be compliant with upstream cucumber gherkin regarding the need for a space or no after the token due to the way this is implemented).

I suggest building a new dialect API (`Dialect` is the upstream name for that) relying on the modern API (an array of strings for keywords). We would then have a compat layer which would the `DialectProvider` API on top of the existing `Keywords` API (basically doing the conversion from gherkin 2 format to gherkin 3 format again), which would be used to wrap the provided Keywords instance passed to the lexer. But we would also support passing directly a dialect provider directly to the lexer instead, and we would ship an implementation consuming a file containing directly the new format (either a json provider loading the upstream `gherkin-languages.json` file directly or a provider loading a file returning a PHP array, as today). The usage of the Keywords API and its compat layer would of course be deprecated.

what do you think @ciaranmcnulty ?

stof:
~~When configuring the Lexer with a dialect provider, invalid language tags in the parsed files will fail instead of silently uswing English keywords, matching the cucumber behavior. Refs #156~~
~~When configuring the Lexer with a KeywordsInterface, the existing silent usage of English will still be done (as the Keywords implementation does that internally).~~
Failing for invalid languages during parsing will be implemented in a follow-up PR based on the compatiblity mode.

An implementation is provided based on the cucumber gherkin-languages.json file, which is now shipped in the package
unmodified (but is considered an internal implementation detail of the new class as we don't expose the path).
No configurable implementation is provided in the core. The DialectProviderInterface is simple enough to implement it fully based on your needs (I expect that the need to read translations from a file having the same structure than the cucumber gherkin-languages.json is not the use case for custom dialect providers, and even in such case, it is dead simple to implement it).

Closes #203

This PR does not provide a replacement for the KeywordsDumper feature for now (which is more a dumper for an example feature). This will be handled in a separate PR.
This PR also does not mark the KeywordsInterface as deprecated yet, but this is expected to happen once the replacement for the KeywordsDumper is available.

### DIFF
--- a/src/Dialect/GherkinDialect.php
+++ b/src/Dialect/GherkinDialect.php
@@ -1,0 +1,159 @@
+<?php
+
+/*
+ * This file is part of the Behat Gherkin Parser.
+ * (c) Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Behat\Gherkin\Dialect;
+
+/**
+ * @phpstan-type TDialectData array{
+ *     feature: non-empty-list<non-empty-string>,
+ *     background: non-empty-list<non-empty-string>,
+ *     scenario: non-empty-list<non-empty-string>,
+ *     scenarioOutline: non-empty-list<non-empty-string>,
+ *     examples: non-empty-list<non-empty-string>,
+ *     rule: non-empty-list<non-empty-string>,
+ *     given: non-empty-list<non-empty-string>,
+ *     when: non-empty-list<non-empty-string>,
+ *     then: non-empty-list<non-empty-string>,
+ *     and: non-empty-list<non-empty-string>,
+ *     but: non-empty-list<non-empty-string>,
+ * }
+ */
+final class GherkinDialect
+{
+    /**
+     * @var non-empty-list<non-empty-string>|null
+     */
+    private ?array $stepKeywordsCache = null;
+
+    /**
+     * @phpstan-param TDialectData $dialect
+     */
+    public function __construct(
+        private readonly string $language,
+        private readonly array $dialect,
+    ) {
+    }
+
+    public function getLanguage(): string
+    {
+        return $this->language;
+    }
+
+    /**
+     * @return non-empty-list<non-empty-string>
+     */
+    public function getFeatureKeywords(): array
+    {
+        return $this->dialect['feature'];
+    }
+
+    /**
+     * @return non-empty-list<non-empty-string>
+     */
+    public function getBackgroundKeywords(): array
+    {
+        return $this->dialect['background'];
+    }
+
+    /**
+     * @return non-empty-list<non-empty-string>
+     */
+    public function getScenarioKeywords(): array
+    {
+        return $this->dialect['scenario'];
+    }
+
+    /**
+     * @return non-empty-list<non-empty-string>
+     */
+    public function getScenarioOutlineKeywords(): array
+    {
+        return $this->dialect['scenarioOutline'];
+    }
+
+    /**
+     * @return non-empty-list<non-empty-string>
+     */
+    public function getRuleKeywords(): array
+    {
+        return $this->dialect['rule'];
+    }
+
+    /**
+     * @return non-empty-list<non-empty-string>
+     */
+    public function getGivenKeywords(): array
+    {
+        return $this->dialect['given'];
+    }
+
+    /**
+     * @return non-empty-list<non-empty-string>
+     */
+    public function getWhenKeywords(): array
+    {
+        return $this->dialect['when'];
+    }
+
+    /**
+     * @return non-empty-list<non-empty-string>
+     */
+    public function getThenKeywords(): array
+    {
+        return $this->dialect['then'];
+    }
+
+    /**
+     * @return non-empty-list<non-empty-string>
+     */
+    public function getAndKeywords(): array
+    {
+        return $this->dialect['and'];
+    }
+
+    /**
+     * @return non-empty-list<non-empty-string>
+     */
+    public function getButKeywords(): array
+    {
+        return $this->dialect['but'];
+    }
+
+    /**
+     * @return non-empty-list<non-empty-string>
+     */
+    public function getStepKeywords(): array
+    {
+        if ($this->stepKeywordsCache !== null) {
+            return $this->stepKeywordsCache;
+        }
+
+        $stepKeywords = [
+            ...$this->getGivenKeywords(),
+            ...$this->getWhenKeywords(),
+            ...$this->getThenKeywords(),
+            ...$this->getAndKeywords(),
+            ...$this->getButKeywords(),
+        ];
+
+        // Sort longer keywords before shorter keywords being their prefix
+        rsort($stepKeywords);
+
+        return $this->stepKeywordsCache = $stepKeywords;
+    }
+
+    /**
+     * @return non-empty-list<non-empty-string>
+     */
+    public function getExamplesKeywords(): array
+    {
+        return $this->dialect['examples'];
+    }
+}
--- a/src/Dialect/KeywordsDialectProvider.php
+++ b/src/Dialect/KeywordsDialectProvider.php
@@ -1,0 +1,100 @@
+<?php
+
+/*
+ * This file is part of the Behat Gherkin Parser.
+ * (c) Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Behat\Gherkin\Dialect;
+
+use Behat\Gherkin\Exception\NoSuchLanguageException;
+use Behat\Gherkin\Keywords\ArrayKeywords;
+use Behat\Gherkin\Keywords\KeywordsInterface;
+
+/**
+ * Adapter for the legacy keywords interface.
+ *
+ * @internal
+ */
+final class KeywordsDialectProvider implements DialectProviderInterface
+{
+    public function __construct(
+        private readonly KeywordsInterface $keywords,
+    ) {
+    }
+
+    public function getDialect(string $language): GherkinDialect
+    {
+        // The legacy keywords interface doesn't support detecting whether changing the language worked or no.
+        $this->keywords->setLanguage($language);
+
+        if ($this->keywords instanceof ArrayKeywords && $this->keywords->getLanguage() !== $language) {
+            throw new NoSuchLanguageException($language);
+        }
+
+        return $this->buildDialect($language);
+    }
+
+    public function getDefaultDialect(): GherkinDialect
+    {
+        // Assume a default dialect of `en` as the KeywordsInterface does not allow reading its language but returns the current data
+        $language = $this->keywords instanceof ArrayKeywords ? $this->keywords->getLanguage() : 'en';
+
+        return $this->buildDialect($language);
+    }
+
+    private function buildDialect(string $language): GherkinDialect
+    {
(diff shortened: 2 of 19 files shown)


stof:
the code coverage reduction is because `ArrayKeywords::getStepKeywords` becomes uncovered as it was only ever called by the Lexer (which now relies on the Dialect API instead).

src/Dialect/KeywordsDialectProvider.php
```diff
@@ -0,0 +1,92 @@
...
+final class KeywordsDialectProvider implements DialectProviderInterface
+{
+    public function __construct(
+        private readonly KeywordsInterface $keywords,
+    ) {
+    }
+
+    public function getDialect(string $language): GherkinDialect
+    {
+        // The legacy keywords interface doesn't support detecting whether changing the language worked or no.
```
stof, line 29: I'm wondering whether we should do a special case by adding an internal `ArrayKeywords::getLanguage` method to read the language, and then using it when the KeywordsInterface implementation extends ArrayKeywords (which is the case for 99.99% of usages at least IMO, as all our core implementations of KeywordsInterface extend it, except the new DialectKeywords internal implementation added in this PR but which has no reason to be wrapped again here except doing crazy things).
On the other hand, not being able to detect the failure here is what preserves the existing behavior, making the switch to the DialectProvider API a kind of opt-in for the strict validation of language comments.
What do you think @acoulton ?

stof (commented):
ParserTest and ParserExceptionsTest are still using the Keywords API as they were using locally-defined translations, not the i18n.php containing the official gherkin translations. As we are in the process of deleting those tests to replace them with compatibility tests contributed to cucumber, I decided it was not worth validating that the translations were still in sync.

src/Dialect/GherkinDialect.php
```diff
@@ -0,0 +1,159 @@
...
+ *     and: non-empty-list<non-empty-string>,
+ *     but: non-empty-list<non-empty-string>,
+ * }
+ */
+final class GherkinDialect
+{
+    /**
+     * @var non-empty-list<non-empty-string>|null
+     */
+    private ?array $stepKeywordsCache = null;
```
acoulton, line 33: I would potentially just make it `private array $stepKeywordsCache;` so that it is clearly not initialised. 

The getter can still safely do `if (isset($this->stepKeywordsCache))` or `$this->stepKeywordsCache ??=` etc to check if the cache is/needs to be initialised, but the code that populates the cache is guaranteed to assign an array, and it makes clear that the property is only expected to be accessed through the getter/initialiser method?

src/Dialect/KeywordsDialectProvider.php
acoulton: I think that's a good idea, and makes all the DialectProvider implementations consistent.

I don't think we can control BC entirely based on which language/dialect provider is injected - I suspect Behat will want to start injecting the new DialectProvider, but that doesn't mean the end-user wants to opt-in to the new strict behaviour?

I'd suggest that the `Lexer->setLanguage()` takes responsibility for the BC based on the GherkinCompatibilityMode - either with something like `$this->dialectProvider->supportsLanguage()`, or by catching the NoSuchLanguageException and switching back to `en` if parsing in legacy mode?

src/Keywords/DialectKeywords.php
```diff
@@ -0,0 +1,121 @@
...
+    private function getStepKeywordString(array $keywords): string
+    {
+        $legacyKeywords = [];
+        foreach ($keywords as $keyword) {
+            if (str_ends_with($keyword, ' ')) {
+                $legacyKeywords[] = substr($keyword, 0, -1);
+            } else {
+                $legacyKeywords[] = $keyword . '<';
+            }
+        }
```
acoulton, line 117: If we're dumping keywords for use in example features (I think that's the only usecase?) isn't it clearer to just leave them with/without the final ` ` rather than replicating the old `<` behaviour?

src/Lexer.php
```diff
@@ -108,9 +116,21 @@ public function analyse($input, $language = 'en')
         $this->allowMultilineArguments = false;
         $this->allowSteps = false;
 
-        $this->setLanguage($language);
+        if (\func_num_args() > 1) {
+            // @codeCoverageIgnoreStart
+            \assert($language !== '');
+            // TODO trigger deprecation (the Parser does not use this code path)
+            $this->setLanguage($language);
+        // @codeCoverageIgnoreEnd
```
acoulton, line 124: Minor formatting issue, should be indented?

src/Lexer.php
```diff
@@ -394,33 +443,36 @@ protected function scanEOS()
...
-        if (!isset($this->keywordsCache[$type])) {
-            $getter = 'get' . $type . 'Keywords';
-            $keywords = $this->keywords->$getter();
-
-            if ($type === 'Step') {
-                $padded = [];
-                foreach (explode('|', $keywords) as $keyword) {
-                    $padded[] = str_contains($keyword, '<')
-                        ? preg_quote(mb_substr($keyword, 0, -1, 'utf8'), '/') . '\s*'
-                        : preg_quote($keyword, '/') . '\s+';
```
acoulton, line 414: If I've understood right, `currentDialect->getStepKeywords()` will return the list of keywords, which may or may not have a trailing space, but that's not the same as the current behaviour where the return from this method matches *at least* zero-or-one spaces depending on how the keyword is defined?

src/Lexer.php
```diff
@@ -501,15 +597,27 @@ protected function scanStep()
             return null;
         }
 
-        $keywords = $this->getKeywords('Step');
-        if (!preg_match('/^\s*(' . $keywords . ')([^\s].*)/u', $this->line, $matches)) {
+        $trimmedLine = $this->getTrimmedLine();
+        $matchedKeyword = null;
+
+        foreach ($this->currentDialect->getStepKeywords() as $keyword) {
```
acoulton, line 603: Is there a potential performance hit here, compared to the previous regex approach? There are quite a lot of step keyword variants to check for every line of every feature.

acoulton:
@stof this is broadly looking great so far, I ran out of time to get to the end of it just now but have left some initial small comments / thoughts.

src/Keywords/DialectKeywords.php
stof: I'm replicating the old structure here to be compatible with the expectation of the KeywordsDumper (which expects the old format, but won't dump a `<` in the output)

src/Lexer.php
stof: php-cs-fixer made me un-indent it, because it considers that a comment just before the `else` is related to the `else` clause (comments are generally put before the thing they explain)

src/Lexer.php
```diff
@@ -343,9 +362,12 @@ protected function scanInput($regex, $type)
      * @return array|null
      *
      * @phpstan-return TToken|null
+     *
+     * @deprecated
      */
     protected function scanInputForKeywords($keywords, $type)
     {
+        // @codeCoverageIgnoreStart
```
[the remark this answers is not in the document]
stof: I would say the `@deprecated` on the method is enough. But I could add one.

src/Lexer.php
stof: the "at least" is covered by the fact that I'm left-trimming the step text.

src/Dialect/KeywordsDialectProvider.php
stof: I will implement that on Monday.

src/Dialect/GherkinDialect.php
acoulton: Ah, I hadn't realised phpstan wouldn't warn on unsafe access to the uninitialized property. In that case I agree, `?array = null` is good.

src/Lexer.php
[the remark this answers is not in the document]
acoulton: Good point, as you were.

src/Lexer.php
acoulton: I might be misunderstanding, but isn't this the "dead code" for BC with classes that have extended the Lexer with e.g. a custom `scanStep` that is still using the old `scanInputForKeywords` though? So they will still need the regex as before?

src/Lexer.php
```diff
@@ -737,27 +846,36 @@ protected function scanText()
...
 
-        if (empty($this->stepKeywordTypesCache)) {
-            $this->stepKeywordTypesCache = [
-                'Given' => explode('|', $this->keywords->getGivenKeywords()),
-                'When' => explode('|', $this->keywords->getWhenKeywords()),
-                'Then' => explode('|', $this->keywords->getThenKeywords()),
-                'And' => explode('|', $this->keywords->getAndKeywords()),
-                'But' => explode('|', $this->keywords->getButKeywords()),
-            ];
+        if (!isset($this->stepKeywordTypesCache[$native])) { // should not happen when the native keyword belongs to the dialect
```
acoulton, line 858: Am I right that we can only get here if the Lexer has found a line starting with a word / phrase that is configured as a step keyword in the Dialect, and $native will be the word that it found - which must, by definition, be part of the dialect?

So I think this is a "cannot happen" rather than a "should not happen". We could probably assert, or throw a LogicException here rather than returning `Given` since this implies an internal error in the Lexer?

tests/Cucumber/CompatibilityTest.php
```diff
@@ -54,13 +54,6 @@ class CompatibilityTest extends TestCase
         'incomplete_background_2.feature' => 'Background descriptions not supported',
     ];
 
-    /**
-     * @var array<string, string>
-     */
-    private array $parsedButShouldNotBe = [
-        'invalid_language.feature' => 'Invalid language is silently ignored',
-    ];
-
```
acoulton, line 63: I think we'd keep this, but based on #349, and just remove the invalid_language.feature from the `gherkin-32` set.

acoulton:
Looked over the rest of this now @stof, this is looking really good thanks.

src/Dialect/KeywordsDialectProvider.php
stof: I updated the PR to catch the NoSuchLanguageException to preserve the existing behavior of ignoring failure to change language.
A follow-up PR will enable the strict behavior based on the compatibility mode.

tests/Cucumber/CompatibilityTest.php
stof: I re-added this for now. It will handled in a follow-up PR (once both the dialect API and the compatibility mode are merged).

tests/TranslationTest.php
```diff
@@ -0,0 +1,173 @@
...
+                ];
+            }
+        }
+    }
+
+    /**
+     * @param string $language language name
+     * @param int $num Fixture index for that language
+     * @param FeatureNode $etalon etalon features (to test against)
+     * @param string $source gherkin source
```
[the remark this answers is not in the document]
stof: This is copied from our existing KeywordsTestCase which I used as basis for the new test based on the new API. I agree that those comments are redundant.

PR #353, opened 2025-05-24
Add native return type when possible

stof:
Native parameter types are added for all arguments.

Return types are added only for private methods as adding a returning type on an method that can be overriden is a BC break.

### DIFF
--- a/src/Cache/CacheInterface.php
+++ b/src/Cache/CacheInterface.php
@@ -27,7 +27,7 @@ interface CacheInterface
      *
      * @return bool
      */
-    public function isFresh($path, $timestamp);
+    public function isFresh(string $path, int $timestamp);
 
     /**
      * Reads feature cache from path.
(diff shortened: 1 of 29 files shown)


acoulton (approved):
Great, thanks @stof

uuf6429:
Oh, I wasn't aware that adding native parameters types is not a BC break. 😮

stof:
@uuf6429 this requires PHP 7.2+ to have the variance rules supported in PHP.
Those variance rules of PHP 7.2+ are also what allows a child class to add a return type even if the parent has not added it yet (making it the other side of the migration path by asking child classes to add a native return type in advance, which is implemented in the DebugClassLoader of symfony/error-handler)

Note that restricting a parameter type is a still BC break for callers. But our BC policy expects the phpdoc `@param` to be taken into account to be covered by BC, so we are not restricting the type supported for callers.

PR #354, opened 2025-05-25
change: Add ParserInterface, deprecate extending Lexer, Parser & nodes

### LINKED ISSUE #347: [RFC] Deprecate extending the lexer and parser
Currently, our lexer and parser are not final and contain a bunch of protected methods that could be called differently or overridden.

As the goal of the Parser class is to parse a file according to the Gherkin specification, I fail to see a use case for extending it. And the Lexer is a support class of the Parser to achieve that same goal.

I suggest tagging those 2 classes as `@final since 4.14.0` (or whatever minor version will be the one shipping that change) to communicate that projects using our package should not extend it (there is tooling for that: the DebugClassLoader of symfony/error-handler triggers a deprecation warning, phpstan reports an error, I haven't checked Psalm).

I suggest following the Symfony BC policy here for contributors: a `@final since ...` tag would still require us to maintain BC for child classes until the next major version. When doing the next major version, we can then migrate them to an actual final class or to a `@final` tag (where the BC guarantee is the same than the final keyword, making this mostly useful to allow mocking the class as mocks cannot be created for a final class).

Looking at the open-source dependents of `behat/gherkin` at https://packagist.org/packages/behat/gherkin/dependents?order_by=downloads, the 2 main users (`behat/behat` and `codeception/codeception`) are using our Parser without extending it. The packages coming next falls in a few categories:

- packages using Codeception that added an explicit requirement to restrict the gherkin version due to our 4.13 release changing the structure of the repo that broke locating the i18n.php file (most of them could remove their requirement again now that Codeception has released a compatible version). Those packages don't use the Gherkin code directly at all
- packages providing custom Behat formatters (which deal with the AST nodes only as parsing is done by Behat)
- helper packages performing assertions based on the multiline arguments, and so dealing with our TableNode, not the Parser.
- tools using the parser (`behat/behat` and `codeception/codeception` are technically in this category as well), for which the ones I looking at (on the first 2 pages ordered by downloads) are using our Parser directly

acoulton:
As per #347 there is no reason for end-users to extend these classes, and leaving them open for extension complicates maintenance. The ParserInterface provides for users who wish to mock (or entirely replace) the Parser.

They will be marked as `final` for the 5.0.0 release.

The only Node class I have not marked as `@final` is TableNode - we extend it ourselves for ExampleTableNode, and it contains more logic than a simple DTO. This should be considered separately for 5.0.

Fixes #347


stof:
`ExampleTableNode::getName` should mark that phpstan error as ignored until 5.0 (using a `string` return type makes sense for a method that is actually final, but we don't want to break BC yet).
`Parser::parseTags` should probably do the same.

acoulton:
Ah, of course, the phpstan errors are because the classes are now final - I did this as a quick chore in a gap at the weekend and wasn't quite concentrating. Thanks @stof.

I think it's safe to just change the phpdoc type on `Parser::parseTags` as that wouldn't be enforced on an extension class anyway?

PR #357, opened 2025-05-26
Reject invalid language in gherkin compatibility mode

### LINKED ISSUE #156: Upstream gherkin breaks if language is not recognised

stof:
Closes #156


acoulton (approved):
Great, thanks @stof

PR #361, opened 2025-05-26
[WIPfeat: Support parsing all describable nodes in `gherkin-32` mode

acoulton:


src/GherkinCompatibilityMode.php
```diff
@@ -35,14 +35,25 @@ public function shouldRemoveStepKeywordSpace(): bool
     /**
      * @internal
      */
-    public function shouldRemoveFeatureDescriptionPadding(): bool
+    public function shouldRemoveDescriptionPadding(): bool
```
acoulton, line 38: Renamed because in legacy mode, padding is also removed from descriptions that get combined into `title` on elements that did not previously support a description.

src/Parser.php
```diff
@@ -202,8 +203,7 @@ protected function parseFeature()
         \assert(\array_key_exists('keyword', $token));
         \assert(\array_key_exists('indent', $token));
 
-        $title = trim($token['value'] ?? '');
-        $description = null;
+        ['title' => $title, 'description' => $description] = $this->parseTitleAndDescription($token);
```
acoulton, line 206: Trimming and coalescing to & from null now all happens in `parseTitleAndDescription` depending on the compatibility mode.

src/Parser.php
```diff
@@ -216,25 +216,7 @@ protected function parseFeature()
...
-                    continue;
-                }
-
-                // It must be part of the feature description (text & newlines later in the document will be consumed as
-                // part of parsing Background / Scenario before execution returns to this loop).
-                $description .= $node;
-                if ($node !== "\n") {
-                    // Text nodes do not end with a newline, add one. The final trailing newline is rtrimmed below.
-                    $description .= "\n";
-                }
```
acoulton, line 237: Any `Text` lines that are validly part of the Feature description will have been consumed by `parseTitleAndDescription` before we enter the loop.

Once in the loop, the only string lines we can get are blank lines - everything else will be consumed (or rejected) as part of processing a `Background` or `Scenario / Scenario Outline` child.

src/Parser.php
```diff
@@ -299,27 +281,25 @@ protected function parseBackground()
         $steps = [];
         $allowedTokenTypes = ['Step', 'Newline', 'Text', 'Comment'];
         while (in_array($this->predictTokenType(), $allowedTokenTypes)) {
+            // NB: Technically, we do not support `Text` inside this loop. However, there is no situation where `Text`
+            // can be a direct child or immediately following a Scenario. Therefore, we consume it here as the most
+            // logical context for throwing an UnexpectedParserNodeException.
+
```
acoulton, line 287: For all of the child node types, as with `parseFeature` any `Text` lines that are validly part of the description will now be consumed before we enter the loop.

However, I have left the loops accepting `Text` nodes (which will then trigger the `UnexpectedParserNodeException`) for cases where there is text at an **invalid** position in the feature file e.g.:

```
# tests/Cucumber/extra_testdata/bad/unknown_step_type.feature
Feature:

    Scenario:
        Given some step
        Aaand some step
```

In cases like this:
a) IMO it is anyway clearer to throw `Expected Step, Examples table, or end of Scenario but got ...` as now, than exiting the loop and doing the throw from `parseFeature` with a message like `Expected Background, Scenario or Outline`
b) There is much less risk of unexpected impact (in particular on legacy mode) if we keep consuming and validating the same node types in each method, so that control flow moves between parser methods on the same lines of each file as it did before.

src/Parser.php
```diff
@@ -495,10 +470,11 @@ protected function parseExamples()
         \assert(\array_key_exists('keyword', $token));
         $keyword = $token['keyword'];
         $tags = empty($this->tags) ? [] : $this->popTags();
+        ['title' => $title, 'description' => $description] = $this->parseTitleAndDescription($token);
```
acoulton, line 473: Note, this will actually produce a small behaviour change in legacy mode.

Previously, we ignored any title on an Examples: line (the `$token['value']` was parsed but just discarded). Control flow then moves to `parseTableRows` which only accepts `TableRow|Newline|Comment`.

Therefore if there was text on the line below `Examples:` this would have caused a ParserException due to the unexpected `Text` content.

Now, even in legacy mode, those lines will be read and the ExampleTableNode will be created with a (possibly multiline) name and a null description (the same as we parse Scenario: etc in legacy mode).

I think this is OK since any files with syntax like that would have been invalid until now?

src/Parser.php
```diff
@@ -660,6 +636,64 @@ private function parseTableRows(): array
...
+            $token = $this->expectTokenType($predicted);
+
+            if ($token['type'] === 'Comment') {
+                continue;
+            }
+
+            $text = match ($token['type']) {
+                'Newline' => '',
+                'Text' => $token['value'],
+                default => throw new LogicException('Unexpected token type: ' . $token['type']),
```
acoulton, line 659: phpstan required this default in the match block, but it can't actually be triggered. It may be redundant once #344 is merged and phpstan has more information about the token types?

src/Parser.php
```diff
@@ -660,6 +636,64 @@ private function parseTableRows(): array
...
+            };
+
+            if ($this->compatibilityMode->shouldRemoveDescriptionPadding()) {
+                $text = preg_replace('/^\s{0,' . ($keywordToken['indent'] + 2) . '}|\s*$/', '', $text);
```
acoulton, line 663: This is the regex that the individual methods were originally using to un-indent / trim text lines when parsing their nodes.

src/Parser.php
acoulton: Ah, of course. I'll leave it as a throw then.

stof:
I think NameFilter needs to match the name and description combined, to match the same thing than before (when the name was multiline).

And we can think about the future handling of filters in Behat as a separate point.

acoulton:
@stof I've rebased this and finished the work on the filters : I think it is now complete.

acoulton:
@stof yeah, I'm hoping to get back to this asap - I also want to check if we have enough coverage of the parsing in `legacy` mode to be sure this hasn't changed anything...

src/Parser.php
```diff
@@ -679,6 +655,69 @@ private function parseTableRows(): array
...
+            $text = match ($token['type']) {
+                'Newline' => '',
+                'Text' => $token['value'],
+                default => throw new LogicException('Unexpected token type: ' . $token['type']),
+            };
+
+            // The only time we use $token['value'] is if we got a `Text` token.
+            // ->expectTokenType('Text') is tagged as returning a `TStringValueToken`, where 'value' cannot be null
+            // However PHPStan cannot follow the chain through predictTokenType -> expectTokenType -> $token['type']
+            assert($text !== null, 'Text token value should not be null');
```
acoulton: Without this, PHPStan was complaining that the `$text` argument to preg_replace could be null (when in fact it cannot be). I'm not sure if there's a cleverer way to solve this.

tests/Cucumber/extra_testdata/good/extra_blank_lines_everywhere.feature
```diff
@@ -0,0 +1,47 @@
...
+
+    Background: That is described
+
+        With multiple paragraphs
+
+        Of description
+
+        Given something
+
+        And something else
```
acoulton: This new feature file covers presence of blank lines between steps in a Background (and in all other parts of a gherkin file).

src/Parser.php
```diff
@@ -329,27 +311,25 @@ protected function parseBackground()
...
 
-            if ($steps === [] && is_string($node)) {
-                $text = preg_replace('/^\s{0,' . ($token['indent'] + 2) . '}|\s*$/', '', $node);
-                $title .= "\n" . $text;
-                continue;
-            }
-
             if ($node === "\n") {
                 continue;
             }
```
acoulton: This line is not covered by tests. 

I don't think it can actually ever happen:

* Before we get to this point, `parseTitleAndDescription` has already consumed any comments or newlines after the `Background:` line and before the first `Step`.
* Therefore, the only ones of our `$allowedTokenTypes` that we can actually receive for the first iteration is a `Step` or `Text`.
* If it is a `Text`, we throw.
* If it is a `Step`, we call `parseExpression()`, which internally calls `parseStep()`.
* [`parseStep()` itself consumes any comments or blank lines that follow the step as part of scanning ahead to find potential PyString / TableRow nodes](https://github.com/Behat/Gherkin/blob/9dec914fc40fdd4da3e4bdaa0792e76f423999a5/src/Parser.php#L496-L500).
* Therefore on our subsequent iterations the only `$allowedTokenTypes` we can actually receive are again `Step` or `Text`.

If I change `$allowedTypes` to just `['Step', 'Text']` and remove this line, all the tests still pass. 

I don't think there are any edge cases that could break this - we could potentially change it now, or keep it as un-covered code for absolute safety.

My temptation is to leave it as-is in this branch, then remove it in a separate PR for clarity.

PR #363, opened 2025-05-26
Improve the Lexer and Parser types

stof:
Building on top of #344, this fixes more errors related by phpstan at higher level for specific token types.


acoulton (approved):
Thanks @stof

uuf6429:
I had started working on this, but got stuck with the dynamic(generic) return type of `takeToken()`, so I started looking up defining generics alias types.

It didn't occur to me to have a simple template type on the method PHPDoc - very nice @stof!

I think I would have used something else than just `T`, but it's good as is nonetheless.

PR #364, opened 2025-05-27
Further PHPStan fixes and improvements

uuf6429:
The main motivation behind these changes is to be able to increase the PHPStan level, but there are some minor fixes that I came across in between, in which case I'll leave a comment.

### DIFF
--- a/src/Cache/FileCache.php
+++ b/src/Cache/FileCache.php
@@ -73,7 +73,7 @@ class FileCache implements CacheInterface
             return false;
         }
 
-        return filemtime($cachePath) > $timestamp;
+        return Filesystem::getLastModified($cachePath) > $timestamp;
     }
 
     /**
--- a/src/Cache/FileCache.php
+++ b/src/Cache/FileCache.php
@@ -31,7 +31,7 @@ class FileCache implements CacheInterface
      */
     private static function getGherkinVersionHash(): string
     {
-        $version = InstalledVersions::getVersion('behat/gherkin');
+        $version = InstalledVersions::getVersion('behat/gherkin') ?? 'unknown';
 
         // Composer version strings can contain arbitrary content so hash for filesystem safety
         return md5($version);
@@ -48,11 +48,11 @@ class FileCache implements CacheInterface
     {
         $this->path = rtrim($path, DIRECTORY_SEPARATOR) . DIRECTORY_SEPARATOR . self::getGherkinVersionHash();
 
-        if (!is_dir($this->path)) {
-            @mkdir($this->path, 0777, true);
+        if (!is_dir($this->path) && !mkdir($this->path, 0777, true) && !is_dir($this->path)) {
+            throw new CacheException(sprintf('Cache directory "%s" could not be created.', $this->path));
         }
 
-        if (!is_writeable($this->path)) {
+        if (!is_writable($this->path)) {
             throw new CacheException(sprintf('Cache path "%s" is not writeable. Check your filesystem permissions or disable Gherkin file cache.', $this->path));
         }
     }
@@ -89,7 +89,7 @@ class FileCache implements CacheInterface
     {
         $cachePath = $this->getCachePathFor($path);
         try {
-            $feature = unserialize(Filesystem::readFile($cachePath));
+            $feature = unserialize(Filesystem::readFile($cachePath), ['allowed_classes' => true]);
         } catch (FilesystemException $ex) {
             throw new CacheException("Can not load cache: {$ex->getMessage()}", previous: $ex);
         }
--- a/src/Filesystem.php
+++ b/src/Filesystem.php
@@ -67,4 +67,30 @@ final class Filesystem
 
         return $found;
     }
+
+    public static function getLastModified(string $path): int
+    {
+        $result = @filemtime($path);
+
+        if ($result === false) {
+            throw new FilesystemException("Cannot retrieve last modification time of file: $path");
+        }
+
+        return $result;
+    }
+
+    public static function getRealPath(?string $path): string
+    {
+        if ($path === null) {
+            throw new FilesystemException('Path must not be null');
+        }
+
+        $result = realpath($path);
+
+        if ($result === false) {
+            throw new FilesystemException("Cannot retrieve the real path of $path");
+        }
+
+        return $result;
+    }
 }
--- a/src/Filter/PathsFilter.php
+++ b/src/Filter/PathsFilter.php
@@ -10,6 +10,8 @@
 
 namespace Behat\Gherkin\Filter;
 
+use Behat\Gherkin\Exception\FilesystemException;
+use Behat\Gherkin\Filesystem;
 use Behat\Gherkin\Node\FeatureNode;
 use Behat\Gherkin\Node\ScenarioInterface;
 
@@ -33,21 +35,26 @@ class PathsFilter extends SimpleFilter
     public function __construct(array $paths)
     {
         foreach ($paths as $path) {
-            if (($realpath = realpath($path)) === false) {
-                continue;
+            try {
+                $realpath = Filesystem::getRealPath($path);
+                $this->filterPaths[] = rtrim($realpath, DIRECTORY_SEPARATOR)
+                    . (is_dir($realpath) ? DIRECTORY_SEPARATOR : '');
+            } catch (FilesystemException) {
+                // skip invalid path
             }
-            $this->filterPaths[] = rtrim($realpath, DIRECTORY_SEPARATOR)
-                . (is_dir($realpath) ? DIRECTORY_SEPARATOR : '');
         }
     }
 
     public function isFeatureMatch(FeatureNode $feature)
     {
-        foreach ($this->filterPaths as $path) {
-            if (($filePath = $feature->getFile()) !== null
-                && ($realPath = realpath($filePath)) !== false
-                && str_starts_with($realPath, $path)
-            ) {
+        try {
+            $realFeatureFilePath = Filesystem::getRealPath($feature->getFile());
+        } catch (FilesystemException) {
+            return false;
+        }
+
+        foreach ($this->filterPaths as $filterPath) {
+            if (str_starts_with($realFeatureFilePath, $filterPath)) {
                 return true;
             }
         }
--- a/src/Filter/PathsFilter.php
+++ b/src/Filter/PathsFilter.php
@@ -44,7 +44,10 @@ class PathsFilter extends SimpleFilter
     public function isFeatureMatch(FeatureNode $feature)
     {
         foreach ($this->filterPaths as $path) {
-            if (str_starts_with(realpath($feature->getFile()), $path)) {
+            if (($filePath = $feature->getFile()) !== null
+                && ($realPath = realpath($filePath)) !== false
+                && str_starts_with($realPath, $path)
+            ) {
                 return true;
             }
         }
--- a/src/Gherkin.php
+++ b/src/Gherkin.php
@@ -106,12 +106,14 @@ class Gherkin
         $filters = array_merge($this->filters, $filters);
 
         $matches = [];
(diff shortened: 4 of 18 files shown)


src/Cache/FileCache.php
```diff
@@ -86,7 +88,11 @@ public function isFresh(string $path, int $timestamp)
     public function read(string $path)
     {
         $cachePath = $this->getCachePathFor($path);
-        $feature = unserialize(file_get_contents($cachePath));
+        $feature = unserialize(
+            file_get_contents($cachePath)
+                ?: throw new CacheException(sprintf('Cache could not be loaded from file "%s".', $path)),
```
uuf6429, line 93: (PHPStan) file_get_contents might fail and return false. I think this change should be safe, we shouldn't get anything else equating to false.

src/Cache/FileCache.php
```diff
@@ -86,7 +88,11 @@ public function isFresh(string $path, int $timestamp)
...
+            ['allowed_classes' => true],
```
uuf6429, line 94: (IDE) this is more explicit, although maybe we might want to define exactly which classes for security reasons?

src/Filter/PathsFilter.php
```diff
@@ -44,7 +44,10 @@ public function __construct(array $paths)
     public function isFeatureMatch(FeatureNode $feature)
     {
         foreach ($this->filterPaths as $path) {
-            if (str_starts_with(realpath($feature->getFile()), $path)) {
+            if (($filePath = $feature->getFile()) !== null
+                && ($realPath = realpath($filePath)) !== false
```
uuf6429, line 48: (PHPStan) `realpath()` might fail and return false (which in turn doesn't work with `str_starts_with()`)

src/Gherkin.php
```diff
@@ -106,12 +106,14 @@ public function load($resource, array $filters = [])
         $filters = array_merge($this->filters, $filters);
 
         $matches = [];
-        if (preg_match('/^(.*):(\d+)-(\d+|\*)$/', $resource, $matches)) {
-            $resource = $matches[1];
-            $filters[] = new LineRangeFilter($matches[2], $matches[3]);
-        } elseif (preg_match('/^(.*):(\d+)$/', $resource, $matches)) {
-            $resource = $matches[1];
-            $filters[] = new LineFilter($matches[2]);
+        if (is_scalar($resource) || $resource instanceof \Stringable) {
```
uuf6429, line 109: (PHPStan) `$resource` is not always stringable.

I'm checking for `scalar` or `Stringable` for widest compatibility possible - is still a BC concern?

src/Loader/ArrayLoader.php
```diff
@@ -31,9 +31,9 @@
  * @phpstan-type TExampleHash array{table: TExampleTableHash, tags?: list<string>}|TExampleTableHash
  * @phpstan-type TExampleTableHash array<int<1, max>, list<string>>
  * @phpstan-type TStepHash array{keyword_type?: string, type?: string, text: string, keyword?: string, line?: int, arguments?: array<array-key, TArgumentHash>}
- * @phpstan-type TArgumentHash array{type: 'table', rows: TTableHash}|TPySstringHash
+ * @phpstan-type TArgumentHash array{type: 'table', rows: TTableHash}|TPyStringHash
```
uuf6429, line 34: Minor typo.

src/Loader/CucumberNDJsonAstLoader.php
```diff
@@ -57,6 +59,10 @@ public function supports(mixed $resource)
 
     public function load(mixed $resource)
     {
+        if (!(is_string($resource) || $resource instanceof Stringable)) {
```
uuf6429, line 62: (PHPStan) Similar to the previous case - this should guarantee compatibility.

src/Loader/CucumberNDJsonAstLoader.php
```diff
@@ -102,7 +109,7 @@ private static function getFeature(array $json, string $filePath): ?FeatureNode
     private static function getTags(array $json): array
     {
         return array_map(
-            static fn (array $tag) => preg_replace('/^@/', '', $tag['name']),
+            static fn (array $tag) => (string) preg_replace('/^@/', '', $tag['name']),
```
uuf6429, line 112: (PHPStan) not sure in which situation, but PHPStan thinks preg_replace might return null (perhaps in case of error?)

src/Loader/DirectoryLoader.php
```diff
@@ -59,6 +61,7 @@ public function supports(mixed $resource)
     public function load(mixed $resource)
     {
         $path = $this->getAbsolutePath($resource);
+        /** @var Traversable<string|SplFileInfo> $iterator */
```
uuf6429, line 64: (PHPStan) PHPStan doesn't infer the iterator correctly (I think there's a bug/feature issue about this somewhere?).

src/Loader/GherkinFileLoader.php
```diff
@@ -72,7 +73,7 @@ public function load(mixed $resource)
     {
         $path = $this->getAbsolutePath($resource);
         if ($this->cache) {
-            if ($this->cache->isFresh($path, filemtime($path))) {
+            if (($lastModified = filemtime($path)) !== false && $this->cache->isFresh($path, $lastModified)) {
```
uuf6429, line 76: (PHPStan) filemtime might fail and return false, but we need an int at this point.

src/Loader/GherkinFileLoader.php
```diff
@@ -95,6 +96,10 @@ protected function parseFeature(string $path)
     {
         $content = file_get_contents($path);
 
+        if ($content === false) {
+            throw new RuntimeException('Unable to read file: ' . $path);
```
uuf6429, line 100: (PHPStan) file might not be read.

src/Node/TableNode.php
```diff
@@ -42,11 +42,11 @@ public function __construct(
     ) {
         $columnCount = null;
 
-        foreach ($this->getRows() as $ridx => $row) {
+        foreach ($this->getRows() as $rowIndex => $row) {
```
uuf6429, line 45: Just renamed the variable to something more understandable.

src/Node/TableNode.php
```diff
@@ -58,27 +58,27 @@ public function __construct(
...
+                    $rowIndex,
                     $columnCount,
                     count($row)
                 ));
             }
 
-            foreach ($row as $column => $string) {
-                if (!isset($this->maxLineLength[$column])) {
-                    $this->maxLineLength[$column] = 0;
+            foreach ($row as $columnIndex => $cell) {
```
uuf6429, line 67: Renamed both variables to be more consistent.

src/Node/TableNode.php
```diff
@@ -58,27 +58,27 @@ public function __construct(
...
                 }
 
-                if (!is_scalar($string)) {
+                if (!is_scalar($cell)) {
                     throw new NodeException(sprintf(
-                        "Table cell at row '%s', col '%s' is expected to be scalar, got %s",
-                        $ridx,
-                        $column,
-                        gettype($string)
+                        "Table cell at row '%s', column '%s' is expected to be scalar, got %s",
```
uuf6429, line 74: Proper wording feels more appropriate here.

tests/Keywords/DialectKeywordsTest.php
```diff
@@ -35,26 +36,23 @@ protected static function getKeywords(): KeywordsInterface
...
+                    'scenario_outline' => implode('|', (array) ($dialect['scenarioOutline'] ?? [])),
+                    'examples' => implode('|', (array) ($dialect['examples'] ?? [])),
+                    'given' => implode('|', (array) ($dialect['given'] ?? [])),
+                    'when' => implode('|', (array) ($dialect['when'] ?? [])),
+                    'then' => implode('|', (array) ($dialect['then'] ?? [])),
+                    'and' => implode('|', (array) ($dialect['and'] ?? [])),
+                    'but' => implode('|', (array) ($dialect['but'] ?? [])),
+                ]
+                : throw new RuntimeException('Invalid data; expected array of keywords but got: ' . get_debug_type($dialect)),
+            (array) json_decode(Filesystem::readFile(__DIR__ . '/../../resources/gherkin-languages.json'), true, flags: \JSON_THROW_ON_ERROR),
```
uuf6429, line 54: (PHPStan) various issues about missing types or undefined indexes.

tests/Node/TableNodeTest.php
```diff
@@ -44,7 +44,7 @@ public function testConstructorExpectsTwoDimensionalArray(): void
     public function testConstructorExpectsScalarCellValue(): void
     {
         $this->expectExceptionObject(
-            new NodeException("Table cell at row '0', col '0' is expected to be scalar, got array")
+            new NodeException("Table cell at row '0', column '0' is expected to be scalar, got array")
```
uuf6429, line 47: Reflects the change in exception message.

src/Cache/FileCache.php
```diff
@@ -29,7 +29,8 @@ class FileCache implements CacheInterface
      */
     private static function getGherkinVersionHash(): string
     {
-        $version = InstalledVersions::getVersion('behat/gherkin');
+        $version = InstalledVersions::getVersion('behat/gherkin')
+            ?? throw new CacheException('Cannot detect behat/gherkin package version');
```
uuf6429, line 33: I cannot reasonable cover this with a test (it relies on a lot of static stuff), any ideas? Shall we close an eye on this (leave it as is or add codeCoverageIgnore)?

acoulton (changes requested):
Thanks @uuf6429, a few thoughts

src/Cache/FileCache.php
uuf6429: I also thought about it, but suspected it might not be desired. I'm fine doing it like that though.

src/Loader/DirectoryLoader.php
stof: why `string|SplFileInfo` ? As we use the default flags, it should be `SplFileInfo` only AFAICT.

anyway, this indeed looks like a case where the phpstan inference is not good.

tests/GherkinTest.php
```diff
@@ -237,6 +239,20 @@ public static function resourceLineFilterDataProvider(): iterable
...
+        $gherkin = new class($this->createMock(ParserInterface::class)) extends GherkinFileLoader {
+            public function parseInexistentFile(): void
+            {
+                $this->parseFeature('inexistent-file');
+            }
+        };
+
+        $this->expectExceptionObject(new RuntimeException('Cannot load feature from file; file could not be read: inexistent-file'));
+
+        $gherkin->parseInexistentFile();
```
stof, line 253: There is no reason to create a class extending `GherkinFileLoader`. Instead, use the actual GherkinFileLoader  class and call `$gherkin->parseFeature('inexistent-file')` in the test.

src/Cache/FileCache.php
uuf6429: I'm using 'unknown'. Please reopen if you change your mind.

src/Filesystem.php
```diff
@@ -67,4 +67,30 @@ public static function findFilesRecursively(string $path, string $pattern): arra
 
         return $found;
     }
+
+    public static function getLastModified(string $path): int
```
uuf6429, line 71: I'll move these two functions and the related changes into a separate PR - that should make things cleaner.

tests/GherkinTest.php
```diff
@@ -237,6 +240,21 @@ public static function resourceLineFilterDataProvider(): iterable
         ];
     }
 
+    public function testThatFileMustBeReadable(): void
```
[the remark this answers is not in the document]
uuf6429: Will be removed after the other PR is merged and this one is rebased.

uuf6429:
I'll jump ahead to PHPStan level 10 fixes, so this PR is no longer needed.

PR #365, opened 2025-05-29
Simplify FS operations

uuf6429:
What changed:
- moved `tests/Filesystem.php` to `src/` and made it `@internal`
- added `readJsonFile()` helper to `Filesystem`
- replaced/updated all usages of `file_get_contents(...)` or `json_decode(file_get_contents(...))`
- added `parseFile()` to the parser and parser interface
  - (which although technically a breaking change, we introduced the interface recently)

This is to:
- reduce duplicate code and checks
- less PHPStan warnings

### DIFF
--- a/bin/update_cucumber
+++ b/bin/update_cucumber
@@ -27,9 +27,8 @@ class {
     public function update(string $composerFile, string $outputVarsFile): void
     {
         assert(is_file($composerFile), "'$composerFile' should be a file");
-        $composerConfigStr = $this->getContents($composerFile);
 
-        $currentVersion = $this->findComposerGherkinVersion($composerConfigStr);
+        $currentVersion = $this->findComposerGherkinVersion($composerFile);
         echo "Current local version is {$currentVersion['hash']} (tagged {$currentVersion['tag_name']})\n";
 
         $releases = $this->listGitHubReleases('cucumber/gherkin');
@@ -45,7 +44,7 @@ class {
         }
 
         echo "Next upstream version is {$nextVersion['hash']} (tagged {$nextVersion['tag_name']})\n";
-        $this->updateComposerJsonFile($composerFile, $composerConfigStr, $currentVersion, $nextVersion);
+        $this->updateComposerJsonFile($composerFile, $currentVersion, $nextVersion);
         $this->writeGitHubOutputs($outputVarsFile, [
             'cucumber_updated' => 'yes',
             'cucumber_version' => $nextVersion['tag_name'],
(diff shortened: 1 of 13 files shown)


src/ParserInterface.php
```diff
@@ -26,4 +26,11 @@ interface ParserInterface
      * @throws ParserException
      */
     public function parse(string $input, ?string $file = null);
+
+    /**
+     * Parses a Gherkin file and returns features array.
```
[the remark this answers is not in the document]
uuf6429: Fixed, thanks.

bin/update_cucumber
```diff
@@ -65,11 +64,12 @@ class {
     /**
      * @phpstan-return TCurrentVersionArray
      */
-    private function findComposerGherkinVersion(string $composerConfigStr): array
+    private function findComposerGherkinVersion(string $composerFile): array
     {
-        $composerConfig = $this->jsonDecodeArray($composerConfigStr);
+        $composerConfig = Behat\Gherkin\Filesystem::readJsonFile($composerFile);
```
acoulton, line 69: I think the `$assoc` param has been missed here, so this will be returning `stdClass` rather than `array`?

I don't actually see any uses of `readJsonFile` that are / should be returning anything other than an associative array. I prefer to avoid boolean flags where possible : I'd suggest instead  `public function readJsonFileArray($file): array` which would simplify the interface and avoid the type-checking asserts following the calls.

I don't think we need a general-purpose "decode any JSON as array or stdClass or mixed" method if we're not using it?

tests/Cache/FileCacheTest.php
```diff
@@ -76,11 +76,20 @@ public function testBrokenCacheRead(): void
         // Now simulate the file being corrupted and attempt to read it
         file_put_contents($files[0], '');
 
-        $this->expectException(CacheException::class);
+        $this->expectExceptionObject(new CacheException('Can not load cache for a feature "broken_feature" from '));
```
acoulton, line 79: I find this use of `expectExceptionObject` (and the one below) a little confusing : on first read it suggests that the caught exception should be equal to the passed one, but in fact it is only looking for a partial match on the exception message.

For cases like this I think it's clearer to expect the class and partial message separately?

uuf6429: I did actually get confused when I changed the message globally by mistake, but tests still passed fine.

Then I found that `expectExceptionObject()` simply calls `expectExceptionMessage()` - and in fact it has always been a substring match!! Very confusing, but the code doesn't lie. :)

It worries me more that the existing tests are not checking the exact message string, which is mainly where we want this functionality. I can use `expectExceptionMessageRegExp()` for this case, but it feels like it just spreading a lie (or at least a misconception). 🙈

acoulton (approved):
Great, thanks @uuf6429

src/ParserInterface.php
```diff
@@ -26,4 +26,11 @@ interface ParserInterface
      * @throws ParserException
      */
     public function parse(string $input, ?string $file = null);
+
+    /**
+     * Parses a Gherkin file and feature (or null when none found).
```
stof, line 31:
```suggestion
     * Parses a Gherkin file and returns feature (or null when none found).
```

PR #366, opened 2025-06-03
Minor TableNode changes

uuf6429:
This will..
- rename some local variables
- improve an exception message


acoulton (approved):
Thanks @uuf6429

PR #367, opened 2025-06-03
Centralizse more FS operations

uuf6429:
This PR introduces two more FS helpers functions to decrease the amount of duplicate error handling code.

Extracted from #364.

### DIFF
--- a/src/Filesystem.php
+++ b/src/Filesystem.php
@@ -67,4 +67,30 @@ final class Filesystem
 
         return $found;
     }
+
+    public static function getLastModified(string $path): int
+    {
+        $result = @filemtime($path);
+
+        if ($result === false) {
+            throw new FilesystemException("Cannot retrieve last modification time of file: $path");
+        }
+
+        return $result;
+    }
+
+    public static function getRealPath(?string $path): string
+    {
+        if ($path === null) {
+            throw new FilesystemException('Path must not be null');
+        }
+
+        $result = realpath($path);
+
+        if ($result === false) {
+            throw new FilesystemException("Cannot retrieve the real path of $path");
+        }
+
+        return $result;
+    }
 }
--- a/src/Loader/AbstractFileLoader.php
+++ b/src/Loader/AbstractFileLoader.php
@@ -10,6 +10,9 @@
 
 namespace Behat\Gherkin\Loader;
 
+use Behat\Gherkin\Exception\FilesystemException;
+use Behat\Gherkin\Filesystem;
+
 /**
  * Abstract filesystem loader.
  *
@@ -29,7 +32,7 @@ abstract class AbstractFileLoader implements FileLoaderInterface
      */
     public function setBasePath(string $path)
     {
-        $this->basePath = realpath($path);
+        $this->basePath = Filesystem::getRealPath($path);
     }
 
     /**
@@ -57,20 +60,19 @@ abstract class AbstractFileLoader implements FileLoaderInterface
      */
     protected function findAbsolutePath(string $path)
     {
-        if (is_file($path) || is_dir($path)) {
-            return realpath($path);
-        }
+        try {
+            return Filesystem::getRealPath($path);
+        } catch (FilesystemException) {
+            if ($this->basePath === null) {
+                return false;
+            }
 
-        if ($this->basePath === null) {
-            return false;
+            try {
+                return Filesystem::getRealPath($this->basePath . DIRECTORY_SEPARATOR . $path);
+            } catch (FilesystemException) {
+                return false;
+            }
         }
-
-        if (is_file($this->basePath . DIRECTORY_SEPARATOR . $path)
-               || is_dir($this->basePath . DIRECTORY_SEPARATOR . $path)) {
-            return realpath($this->basePath . DIRECTORY_SEPARATOR . $path);
-        }
-
-        return false;
     }
 
     /**
--- a/tests/Loader/GherkinFileLoaderTest.php
+++ b/tests/Loader/GherkinFileLoaderTest.php
@@ -12,6 +12,7 @@ namespace Tests\Behat\Gherkin\Loader;
 
 use Behat\Gherkin\Cache\CacheInterface;
 use Behat\Gherkin\Dialect\CucumberDialectProvider;
+use Behat\Gherkin\Filesystem;
 use Behat\Gherkin\Lexer;
 use Behat\Gherkin\Loader\GherkinFileLoader;
 use Behat\Gherkin\Parser;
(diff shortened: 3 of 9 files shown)


tests/FilesystemTest.php
```diff
@@ -0,0 +1,36 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Tests\Behat\Gherkin;
+
+use Behat\Gherkin\Exception\FilesystemException;
+use Behat\Gherkin\Filesystem;
+use PHPUnit\Framework\TestCase;
+
+final class FilesystemTest extends TestCase
```
uuf6429, line 17: Somehow I'm not a fan of creating a new test class for this, but unfortunately the real-world cases are difficult to test (e.g. the `file_exists() ... filemtime()` case).

While it's possible to hook into vfsStream and do some fancy self-destroying files, it's not trivial and it would complicate tests.

src/Loader/AbstractFileLoader.php
```diff
@@ -57,20 +60,19 @@ protected function findRelativePath(string $path)
...
+            }
 
-        if ($this->basePath === null) {
-            return false;
+            try {
+                return Filesystem::getRealPath($this->basePath . DIRECTORY_SEPARATOR . $path);
+            } catch (FilesystemException) {
+                return false;
+            }
         }
```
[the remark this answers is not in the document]
uuf6429: I see what you mean, how about the following variation?
```php
        if (file_exists($path)) {
            return Filesystem::getRealPath($path);
        }

        if ($this->basePath === null) {
            return false;
        }

        try {
            return Filesystem::getRealPath($this->basePath . DIRECTORY_SEPARATOR . $path);
        } catch (FilesystemException) {
            return false;
        }
```
_(`file_exists`()` apples to both files and dirs)_

tests/Loader/GherkinFileLoaderTest.php
```diff
@@ -106,12 +107,18 @@ public function testBasePath(): void
         $features = $this->loader->load('features/pystring.feature');
         $this->assertCount(1, $features);
         $this->assertEquals('A py string feature', $features[0]->getTitle());
-        $this->assertEquals(realpath($this->featuresPath . DIRECTORY_SEPARATOR . 'pystring.feature'), $features[0]->getFile());
+        $this->assertEquals(
+            Filesystem::getRealPath($this->featuresPath . DIRECTORY_SEPARATOR . 'pystring.feature'),
```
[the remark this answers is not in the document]
uuf6429: I don't get what you mean about `$this->featuresPath``.. the reason we have/had `realpath()` is because we use double-dots when defining the base path:
```php
$basePath = __DIR__ . '/../Fixtures';
$this->loader->setBasePath($basePath);
```
I could put the `realpath()/getRealPath()` part there, but it would be the same situation you're describing.

If there are test scenarios that we're missing to ensure the stability of `getRealPath`, maybe we should just add such test scenarios?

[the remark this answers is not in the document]
uuf6429: Good point.

src/Loader/AbstractFileLoader.php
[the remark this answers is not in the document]
uuf6429: I went for my last suggestion. Please reopen if you disagree or have some feedback.

src/Loader/AbstractFileLoader.php
```diff
@@ -57,20 +60,19 @@ protected function findRelativePath(string $path)
      */
     protected function findAbsolutePath(string $path)
     {
-        if (is_file($path) || is_dir($path)) {
-            return realpath($path);
+        if (file_exists($path)) {
+            return Filesystem::getRealPath($path);
```
acoulton, line 64: I think for guaranteed BC, this also still needs to catch the exception and return false. 

This is because this method is used from `->supports()` (which just skips over any resource where this method returns `false`, so the later code that calls it and throws if it returns false is never called.).

To be honest I'm slightly struggling to think of examples where this would matter, and if starting from scratch I might be stricter than the old behaviour. But I do think there could be edge cases, maybe if a user has symlinks with strange permissions, that we currently ignore but would now throw on.

I still think [as my last comment](https://github.com/Behat/Gherkin/pull/367#discussion_r2127532007) that since this function explicitly follows the `realpath()` interface and the `false` return value is actually handled everywhere we call it, it is probably simpler to just keep using `realpath` directly rather than call our method just to catch the exception and return false again?

acoulton (changes requested):
Thanks @uuf6429 just a couple of small thoughts.

src/Loader/AbstractFileLoader.php
stof: > I still think [as my last comment](https://github.com/Behat/Gherkin/pull/367#discussion_r2127532007) that since this function explicitly follows the `realpath()` interface and the `false` return value is actually handled everywhere we call it, it is probably simpler to just keep using `realpath` directly rather than call our method just to catch the exception and return false again?

totally agree here

acoulton: Hmm, good point...

IMO, we should not suppress it, so that we are not modifying the existing behaviour (which might mean users see the warning, or get a failure from a custom error handler depending how they are running gherkin). Because I think it probably *is* unexpected that there's a file/directory inside your configured path(s) that exists but cannot be resolved to an absolute path. I just want to be cautious about breaking BC if someone is relying on that.

In a future major we could absolutely refactor this method to be stricter and to bubble the exception (in both cases) rather than returning `false`.

src/Filesystem.php
```diff
@@ -67,4 +76,53 @@ public static function findFilesRecursively(string $path, string $pattern): arra
...
+        try {
+            $result = self::callSafely(static fn () => filemtime($fileName));
+        } catch (ErrorException $e) {
+            throw new FilesystemException(
+                sprintf('Last modification time of file "%s" cannot be found: %s', $fileName, $e->getMessage()),
+                previous: $e,
+            );
+        }
+
+        assert($result !== false, 'realpath() should not return false without emitting a PHP warning');
```
acoulton, line 91:
```suggestion
        assert($result !== false, 'filemtime() should not return false without emitting a PHP warning');
```
Just a typo

src/Loader/AbstractFileLoader.php
```diff
@@ -57,20 +60,19 @@ protected function findRelativePath(string $path)
...
             return false;
         }
 
-        if (is_file($this->basePath . DIRECTORY_SEPARATOR . $path)
-               || is_dir($this->basePath . DIRECTORY_SEPARATOR . $path)) {
-            return realpath($this->basePath . DIRECTORY_SEPARATOR . $path);
+        try {
+            return Filesystem::getRealPath($this->basePath . DIRECTORY_SEPARATOR . $path);
+        } catch (FilesystemException) {
+            return false;
```
acoulton, line 74: Sorry if I wasn't clear before, but I think this usage should probably also just stay as realpath, since we don't gain anything by moving to the new helper here.

acoulton (changes requested):
Thanks for updating @uuf6429 this is very close

acoulton (approved):
Great, thanks @uuf6429

uuf6429:
OK, so can we merge this then, please? I'd like to continue with the next 2 PRs - we're almost there.

acoulton:
I was holding off for a review from @stof since this has changed a bit since his last approval. However he's commented agreement to the main changes I think so I'll merge and we can deal with any further comments in a follow-up PR if required.

PR #368, opened 2025-06-04
Fixes for and update to PHPStan level 10

uuf6429:
- This PR is for tracking the remaining changes to be able to switch to PHPStan level 10. (💯💥🎉)
- ~~Do not spend time reviewing this now - this PR will change for sure.~~
- ~~Depends on #364 (and possibly other PRs).~~

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
     "require-dev": {
         "symfony/yaml": "^5.4 || ^6.4 || ^7.0",
         "phpunit/phpunit": "^10.5",
-        "cucumber/gherkin-monorepo": "dev-gherkin-v33.0.0",
+        "cucumber/gherkin-monorepo": "dev-gherkin-v33.1.0",
         "friendsofphp/php-cs-fixer": "^3.77",
         "phpstan/phpstan": "^2",
         "phpstan/extension-installer": "^1",
@@ -56,16 +56,16 @@
             "type": "package",
             "package": {
                 "name": "cucumber/gherkin-monorepo",
-                "version": "dev-gherkin-v33.0.0",
+                "version": "dev-gherkin-v33.1.0",
                 "source": {
                     "type": "git",
                     "url": "https://github.com/cucumber/gherkin.git",
-                    "reference": "7cd0304c39a51ac139a22264ebe2a7ce6c68819d"
+                    "reference": "52c61370fdfa64bf750836d1a3cd306536f75a44"
                 },
                 "dist": {
                     "type": "zip",
-                    "url": "https://api.github.com/repos/cucumber/gherkin/zipball/7cd0304c39a51ac139a22264ebe2a7ce6c68819d",
-                    "reference": "7cd0304c39a51ac139a22264ebe2a7ce6c68819d"
+                    "url": "https://api.github.com/repos/cucumber/gherkin/zipball/52c61370fdfa64bf750836d1a3cd306536f75a44",
+                    "reference": "52c61370fdfa64bf750836d1a3cd306536f75a44"
                 }
             }
         }


src/Loader/ArrayLoader.php
```diff
@@ -28,11 +28,12 @@
  * @phpstan-type TBackgroundHash array{title?: string|null, keyword?: string, line?: int, steps?: array<int, TStepHash>}
  * @phpstan-type TScenarioHash array{type?: 'scenario', title?: string|null, tags?: list<string>, keyword?: string, line?: int, steps?: array<int, TStepHash>}
  * @phpstan-type TOutlineHash array{type: 'outline', title?: string|null, tags?: list<string>, keyword?: string, line?: int, steps?: array<int, TStepHash>, examples?: TExampleTableHash|array<array-key, TExampleHash>}
- * @phpstan-type TExampleHash array{table: TExampleTableHash, tags?: list<string>}|TExampleTableHash
+ * @phpstan-type TExampleHash TTaggedExampleTableHash|TExampleTableHash
  * @phpstan-type TExampleTableHash array<int<1, max>, list<string>>
+ * @phpstan-type TTaggedExampleTableHash array{table: TExampleTableHash, tags?: list<string>}
```
uuf6429, line 33: Can we find a better name for `TTaggedExampleTableHash`? I named it like that because it's a table "that could have tags" (and I needed a type to do some cleaning and referring to it).

But when you look at, for example, `isTaggedExampleTableHash()`, the name of the method kinda doesn't match with what the body is doing (namely just checking for 'table'). It would be best to keep the body as is, to stay backword compatible, but maybe consider finding a better name.

uuf6429: OK, then since this works for the moment, let's keep it as proposed in this PR and when PHPStan gets updated and starts complaining, we simply fix it. This would also applies to `mb_convert_encoding()`.

What do you think?

src/Loader/CucumberNDJsonAstLoader.php
```diff
@@ -57,14 +59,20 @@ public function supports(mixed $resource)
 
     public function load(mixed $resource)
     {
+        if (!(is_string($resource) || $resource instanceof Stringable)) {
+            return [];
+        }
+
```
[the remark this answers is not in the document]
uuf6429: Done: #372. Hopefully the generics stuff there proves that `$resource` has to be exactly a string, and eventually would make the code change here redundant.

src/Loader/ArrayLoader.php
```diff
@@ -280,30 +283,73 @@ protected function loadPyStringHash(array $hash, int $line = 0)
...
+    }
+
+    /**
+     * @param array<array-key, mixed> $exampleHash
+     *
+     * @phpstan-assert-if-true TExampleTableHash $exampleHash
+     */
+    private function isExampleTableHash(array $exampleHash): bool
+    {
+        return !$this->isTaggedExampleTableHash($exampleHash);
```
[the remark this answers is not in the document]
stof: See my comments in https://github.com/Behat/Gherkin/pull/368#discussion_r2139601340

src/Loader/ArrayLoader.php
uuf6429: @stof / @acoulton ok, so is it fine to rollback the helper methods I added and just tell PHPStan to ignore the errors? Or should I improve them as per [the other comment](https://github.com/Behat/Gherkin/pull/368/files#r2141124604)?

acoulton: I think if it only affects ArrayLoader then my vote is to just tell PHPStan to ignore the errors if that's easiest - thanks.

tests/Loader/ArrayLoaderTest.php
```diff
@@ -326,8 +326,8 @@ public function testLoadStepArguments(): void
                                         'type' => 'table',
                                         'rows' => [
                                             ['key', 'val'],
-                                            [1, 2],
-                                            [3, 4],
+                                            ['1', '2'],
+                                            ['3', '4'],
```
uuf6429, line 330: We've said at some point that the table structure should only contain strings, which conflicts with an alias type defined in `ArrayLoader` (note that I removed the/an `|int`) and then of course this test (and therefore the change above).

I hope this change is correct.

src/Loader/ArrayLoader.php
uuf6429: Done, please reopen if you have any further feedback.

tests/Keywords/DialectKeywordsTest.php
```diff
@@ -35,26 +36,23 @@ protected static function getKeywords(): KeywordsInterface
...
+                    'examples' => implode('|', (array) ($dialect['examples'] ?? [])),
+                    'given' => implode('|', (array) ($dialect['given'] ?? [])),
+                    'when' => implode('|', (array) ($dialect['when'] ?? [])),
+                    'then' => implode('|', (array) ($dialect['then'] ?? [])),
+                    'and' => implode('|', (array) ($dialect['and'] ?? [])),
+                    'but' => implode('|', (array) ($dialect['but'] ?? [])),
+                ]
+                : throw new RuntimeException('Invalid data; expected array of keywords but got: ' . get_debug_type($dialect)),
+            Filesystem::readJsonFileHash(__DIR__ . '/../../resources/gherkin-languages.json'),
+        );
```
acoulton, line 55: Rather than changing this implementation, is it possible to solve it by telling PHPstan the JSON has the (known) structure as in the CucumberDialectProvider e.g. 

https://github.com/Behat/Gherkin/blob/1325278942943332a6bcdc02b99fab76fa20bfb4/src/Dialect/CucumberDialectProvider.php#L30-L36

I see we also do that in the TranslationTest below.

Or does that not work in this context?

tests/Loader/ArrayLoaderTest.php
acoulton: Indeed, we did agree that.

acoulton (commented):
Thanks for this @uuf6429 this is looking great. I only have one remaining question (partly for my own understanding, it may be that your change is correct).

tests/Keywords/DialectKeywordsTest.php
uuf6429: Considering that this is a test data provider, I suppose it's ok to ignore the warnings (and let if fail at runtime, in case some problems comes up from the upstream cucumber repo).

uuf6429: PS: `$data` is not compatible with `TDialectData`, e.g. `scenarioOutline` vs `scenario_outline`

acoulton: That would also be fine, the JSON file can only change in the job that updates to a newer upstream and if that fails it will be fairly clear what's going on. Plus I assume their dialect schema is fairly well fixed now.

uuf6429: I've applied an update - please reopen this for further feedback.

acoulton (approved):
Great, thanks for all your work on this

carlos-granados (commented):
Some small comments/questions, nothing major

src/Keywords/CachedArrayKeywords.php
```diff
@@ -31,6 +31,7 @@ public static function withDefaultKeywords(): self
      */
     public function __construct(string $file)
     {
+        // @phpstan-ignore argument.type
```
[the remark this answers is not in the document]
acoulton: Not in the most recent master, I think?
https://github.com/Behat/Gherkin/blob/ca766ccb3a0d12f72e3ce4b5f52fcdbb05d2fd2a/src/Keywords/ArrayKeywords.php#L72-L77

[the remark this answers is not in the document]
carlos-granados: Argh, you are right, my local branch was not updated :-( Then feel free to ignore this comment

tests/Keywords/CachedArrayKeywordsTest.php
```diff
@@ -24,6 +24,7 @@ protected static function getKeywords(): KeywordsInterface
 
     protected static function getKeywordsArray(): array
     {
+        // @phpstan-ignore return.type
```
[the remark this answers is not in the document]
uuf6429: It's not "just an array", so the assertion would be much more complex than that.

To be honest I don't like ignoring errors like this, but it seems the most reasonable solution. :/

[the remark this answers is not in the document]
carlos-granados: Yeah, same here, I was checking this on an outdated version of the code, so I did not have the complete type, please just ignore this

carlos-granados (approved):
All good now, thanks @uuf6429

uuf6429:
Took some time, but finally this chapter's over. 🎉 Thanks guys!

PR #370, opened 2025-06-08
fix: Use default file cache key if `behat/gherkin` version is unknown

uuf6429:
As per PHPStan, it is possible that the version of the `behat/gherkin` package, (which we use as cache key), cannot be retrieved. This PR makes it default to `'unknown'` in that case.

This change was extracted from #364.


acoulton (approved):
Thanks @uuf6429

PR #372, opened 2025-06-12
Improve loaders load() handling

uuf6429:
- `load()` is now guarded by a call to `supports()` (throwing a `LogicException` when that returns `false`)
- improves narrowing the real type of `mixed $resource` with the use of generics


src/Gherkin.php
```diff
@@ -139,9 +139,11 @@ public function load($resource, array $filters = [])
     /**
      * Resolves loader by resource.
      *
-     * @param mixed $resource Resource to load
+     * @template TResourceType of mixed
```
stof, line 142: `of mixed` is useless, as `mixed` is the top type in PHP.

src/Loader/LoaderInterface.php
```diff
@@ -24,14 +26,16 @@ interface LoaderInterface
      *
      * @param mixed $resource Resource to load
      *
+     * @phpstan-assert-if-true LoaderInterface<TResourceType> $this
```
stof, line 29: this looks weird to me. This is always the type of `$this` due to the generic type.

`@phpstan-assert-if-true =TResourceType $resource` would make more sense (the `=` prefix indicates that returning `false` does not mean that the resource is not of that type, as we might have additional checks, for instance that the string corresponds to a valid file path)

uuf6429: I was doing that specifically for `resolveLoader()` to work well (in PHPStan's perspective).

Using your code breaks it, but now I found out it also breaks on PHPStan level 9.

I did make it work by combining your code with mine though:

https://phpstan.org/r/862e103a-f937-4713-b339-ea212db5649d

(the key difference is the PHPDoc of `LoaderInterface::supports()`)

tests/Loader/GherkinFileLoaderTest.php
```diff
@@ -103,6 +103,7 @@ public function testBasePath(): void
         $this->assertFalse($this->loader->supports('tables.feature'));
         $this->assertTrue($this->loader->supports('features/tables.feature'));
 
+        $this->assertTrue($this->loader->supports('features/pystring.feature'));
```
uuf6429, line 106: Pretty cool that with the new changes, it actually detects that this test was slightly wrong in principle, since `supports()` should be called with the same argument as when (and before) calling `load()`.

acoulton (approved):
Looks good to me, thanks @uuf6429

uuf6429:
@stof could you or @carlos-granados review & approve this and #373 please? 🙏

carlos-granados (approved):
Thanks @uuf6429

PR #374, opened 2025-07-16
ci: Fix error in cucumber update job due to non-autoloadable classes

acoulton:
#365 extracted some filesystem logic from these scripts to the new Filesystem helper class. This broke the job, because the scripts are unable to locate the class. This is because there is no autoloader registered in this environment (and, in the case of the main update_cucumber script, we have not yet run composer install).

Fixed by manually requiring the class for update_cucumber, and registering the autoloader for the update_i18n (which runs after the composer install).

### DIFF
--- a/bin/update_cucumber
+++ b/bin/update_cucumber
@@ -18,6 +18,9 @@ if (ini_get('zend.assertions') !== '1') {
     exit(1);
 }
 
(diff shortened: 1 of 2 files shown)


acoulton:
As this only affects the CI scripts used by the scheduled cucumber_update action, I have merged it without a review so we can get that job working again.

PR #379, opened 2025-07-31
ci: Fix string interpolation in update_cucumber job

acoulton:
#368 accidentally converted this variable to a single quoted string, breaking variable interpolation and causing it to generate an incorrect URL for the github API request.

This previously happened in #365 (which had been picked out of #368) and was fixed in #375, however I didn't spot that the patch had then been reapplied in the final PR.

Change from interpolation to concatenation to minimise the risk of recurrence.


carlos-granados (approved):
Thanks @acoulton

PR #380, opened 2025-08-04
ci: Fix incorrect assumption about github releases in update_cucumber

acoulton:
#368 converted the previous `readJsonFileArray` utility method to a stricter `readJsonFileHash` to improve phpstan checking. This was based on the belief that we were always currently using that method to parse JSON data with an object structure at the root.

However, in the case of the update_cucumber script, the github releases API returns a JSON array of objects. The strict check was therefore causing that job to fail.

I have introduced the looser `readJsonFileArray` for this purpose and confirmed that the task now works locally.

### DIFF
--- a/bin/update_cucumber
+++ b/bin/update_cucumber
@@ -116,7 +116,7 @@ class {
     {
         // GitHub requires API requests to send a user_agent header for tracing
         ini_set('user_agent', 'https://github.com/Behat/Gherkin updater');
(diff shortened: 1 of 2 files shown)


carlos-granados (approved):
Good spot, thanks @acoulton

acoulton:
I've managed to configure github to now send the alerts to me instead of Ciaran if the scheduled action fails (notifications go to the last person to edit the cron, or to disable & then re-enable the job in the web interface) so at least I can now keep on top of any issues :)

PR #383, opened 2025-08-18
style: Fix comma position after heredoc call argument

acoulton:
The previous version (with the comma on the following line) dates back to PHP < 7.3.0 when the heredoc terminator had to be the only thing on the line.

php-cs-fixer 3.86.0 updates the symfony ruleset to require that the comma be on the same line as the heredoc terminator (which we already do in most places), meaning this line is now reported as a violation.


carlos-granados (approved):
Good spot, thanks @acoulton

stof (approved):

PR #388, opened 2025-11-03
feat: Support PHP 8.5

acoulton:
Adds PHP 8.5 to our build matrix and supported versions.


acoulton:
It's unclear why codecov is seeing a drop in coverage - this is based directly off the last commit in master and has no changes to code or tests... Seems like an issue with codecov AFAICS.

carlos-granados (approved):
Thanks @acoulton

stof:
It looks like some lines containing closing braces (at the end of empty constructors doing only property promotion for instance, or the end of a multiline function call) are not marked as covered anymore while they were before: https://app.codecov.io/gh/Behat/Gherkin/pull/388/indirect-changes

Maybe this is a change in the way XDebug or PHPUnit treats those lines.

PR #401, opened 2025-12-02
docs: Add funding links and information

acoulton:
Add information on current maintainers, and how users can support the project.

Completes Behat/Behat#1695 and matches Behat/Behat#1704


carlos-granados (approved):
Looking good @acoulton

stof (approved):

PR #402, opened 2025-12-02
fix: Add `Stringable` to classes implementing __toString()

acoulton:
php-cs-fixer 3.91.0 added a new rule to the `@Symfony` set that adds the `Stringable` interface to classes implementing a `__toString()` method.

This commit applies that rule to our codebase to resolve the current build failure.


carlos-granados (approved):
Thanks @acoulton

PR #404, opened 2025-12-04
Fix the implementation of the default dialect for the keywords provider

### LINKED ISSUE #403: Changed behavior in v4.15 when parsing test files with different languages
I noticed a failing CI job in https://github.com/Codeception/Codeception/actions/runs/19833673725/job/56825691142 when `behat/gherkin:4.15` is used.

The same tests are green when using `behat/gherkin:4.14`.

Codeception uses [`Behat\Gherkin\Keywords\CachedArrayKeywords::withDefaultKeywords()`](https://github.com/Codeception/Codeception/blob/71087f3e643119fae473644ffd5f73ece113b80d/src/Codeception/Test/Loader/Gherkin.php#L73) when initializing the lexer.

The test tries to load all gherkin files.

File 1 and 3 are using the default language "en" whereas the language "ru" is defined for file 2.

File 1

```gherkin
Feature:
  In order to know that the right step will be executed
  As a user
  I should be notified about steps that are ambiguous

  Scenario: We have multiple step definitions that match a given step
    Given a step which matches multiple step definitions
```

File 2

```gherkin
#language: ru

Функционал: Генерация сниппетов из сценариев на других языках
  Как пользователь
  Чтобы выполнить проверку сценариев на родном языке заказчика
  Я хочу иметь возможность генерировать сниппеты из них 

  Сценарий: Генерация сниппета из сценария на русском языке
    Допустим я написал сценарий на языке "RU_ru"
```

File 3

```gherkin
Feature: Run gherkin
  In order to test a feature
  As a user
  I need to be able to have subfolders

  Scenario: Describe a feature in subfolder
    Given I have a feature in a subfolder
```

In `behat/gherkin:4.14` this works without problems.
However, in `behat/gherkin:4.15` the lexer seems to use the language "ru" for file 3 and thus parsing the file fails.

The test becomes green when adding `#language: en` to file 3.

I'm pretty sure this is related to https://github.com/Behat/Gherkin/pull/350 because before this PR, "en" was used as second parameter when calling `Lexer::analyse()` inside `Parser::parse()`.

Is the behavior change intended or does codeception just use `behat/gherkin` the wrong way?

stof:
The legacy KeywordsInterface was stateful, so we cannot assume that the _current_ language is the default one. We need to read the _initial_ language instead.

Closes #403

### DIFF
--- a/src/Dialect/KeywordsDialectProvider.php
+++ b/src/Dialect/KeywordsDialectProvider.php
@@ -21,9 +21,13 @@ use Behat\Gherkin\Keywords\KeywordsInterface;
  */
 final class KeywordsDialectProvider implements DialectProviderInterface
 {
+    private readonly string $defaultLanguage;
+
     public function __construct(
         private readonly KeywordsInterface $keywords,
     ) {
+        // Assume a default dialect of `en` as the KeywordsInterface does not allow reading its language but returns the current data
+        $this->defaultLanguage = $this->keywords instanceof ArrayKeywords ? $this->keywords->getLanguage() : 'en';
     }
 
     public function getDialect(string $language): GherkinDialect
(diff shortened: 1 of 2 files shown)


acoulton (approved):
Thanks @stof

PR #405, opened 2025-12-04
Fix trimming of padding in table cells

stof:
This trims Unicode spaces (as done upstream) without removing newlines produces by escape sequences.

Closes #395

This does not change the legacy mode (the `parseTableCell` private method is not used by the legacy parsing mode).


src/Lexer.php
```diff
@@ -788,18 +788,21 @@ protected function scanTableRow()
 
     private function parseTableCell(string $cell): string
     {
+        $trimmedCell = preg_replace('/^[ \\t\\n\\x0B\\f\\r\\x85\\xA0]++|[ \\t\\n\\x0B\\f\\r\\x85\\xA0]++$/u', '', $cell);
```
acoulton, line 791: I noticed this is slightly different to cucumber/gherkin, they [call their `ltrimKeepNewLines` and `rtrimKeepNewLines` utils](https://github.com/cucumber/gherkin/blob/6b9e4c16c0a659396d86c60d8d49a5362c3675f2/php/src/StringGherkinLine.php#L90-L92) which use a regex the same as yours but without the `\n` character.

I think that's OK though - since we're parsing content from a single line of the file, there can't actually be any literal newlines in the string at this point, so there's no particular reason why gherkin need to use `..trimKeepNewLines` over `..trim` here.

But just thought worth sense checking that assumption.

src/Lexer.php
acoulton: Should we rename the `shouldSupportNewlineEscapeSequenceInTableCell` flag, since that's now covering two slightly different variances?

src/Lexer.php
stof: There is indeed a guarantee that the string is single-line in those cases, so it makes no difference.

src/Lexer.php
stof: `shouldSupportNewlineEscapeSequenceAndUnicodePaddingInTableCell` would be the name covering all differences with the legacy mode, but this is crazy long.

acoulton: It is indeed. Maybe we go more vague - `shouldUseNewTableCellParsing()` or something? It's not a huge deal, the methods are only really for our benefit, but it might be easier to understand in future with a name that's not explicit than one that is explicit but incomplete?

stof: I renamed the method with your suggestion.

acoulton (approved):
Thanks very much @stof

PR #406, opened 2025-12-05
docs: Add changelog for 4.16.0

acoulton:

PR #407, opened 2025-12-08
fix: Reinstate support for legacy tag filters without `@` prefix

acoulton:
As reported by @carlos-granados in #406, the 4.16.0 release broke tag filtering where the tag expression does not contain `@` symbols [e.g. in Behat](https://github.com/Behat/Behat/actions/runs/20028377501/job/57431075976).

Behat's filter expression in the failing feature is arguably incorrect:

* It uses `new TagFilter('tag2')` (note no leading `@`) - the PHP config object is obviously new but the value has been the same since [the feature was first added in 2014](https://github.com/Behat/Behat/pull/434/files#diff-ca77b0b1e7860ced7eb08cc40f9d6648acc0760a0be45c5fd76a8a19992adbebR92)
* All the Behat documentation shows expressing tag filters with the leading `@` e.g. https://docs.behat.org/en/latest/user_guide/configuration/suites.html#suite-filters
* The actual implementation of tag filtering is done in Behat/Gherkin. Gherkin's [TagFilterTest only tests tag expressions where the leading `@` is included](https://github.com/Behat/Gherkin/blame/297297343c125d058e18b3f03a24ebb32f70b69d/tests/Filter/TagFilterTest.php) and that appears to have been the case from the beginning.

As far as I can see, `TagFilter` has therefore never been meant to accept that expression, and it's been working by accident.

However, we obviously don't validate the expression, and if even Behat has been relying on it then I think we should reinstate support for this for now.

I have started from the 4.15.0 tag to prove the legacy behaviour was working, and will then add more test cases and the fix.

I will open a separate issue to add a deprecation for this usage and remove it in due course. I don't want to add a deprecation now as that could also cause build failures for existing users until we have a proper way to report it in Behat.


acoulton:
@carlos-granados annoyingly my first commit (d50df91) based off the v4.15.0 tag to prove these were passing originally doesn't pass here because of course pull requests are built on the merge of the PR back to master, which already contains the broken implementation.

However I can confirm that all of those tests were passing locally before I merged the changes from 4.16.0 into the branch.

I've added a changelog in this PR so we can get a bugfix release out as soon as it's merged.

carlos-granados (approved):
Just a tiny change needed

acoulton:
> Just a tiny change needed

Sorry @carlos-granados GH isn't showing me that - just your approval?

tests/Filter/TagFilterTest.php
```diff
@@ -279,6 +279,54 @@ public function testFilterFeatureWithTaggedExamples(): void
         $this->assertEquals([$exampleTableNode3], $scenarioInterfaces[1]->getExampleTables());
     }
 
+    /**
+     * @phpstan-return list<array{string, list<string>, bool}>
+     */
+    public static function provderMatchWithNoPrefixInFilter(): array
```
carlos-granados: Should be `providerMatchWithNoPrefixInFilter` (missing `i` in `provider`)

carlos-granados: Somehow GitHub did not register my comment

acoulton: Thanks, good spot. Fixed now

carlos-granados (approved):

PR #413, opened 2026-03-24
fix: Phpstan errors in current master build

acoulton:
The master branch is failing static analysis due to changes in recent phpstan versions.

One of these I think is a valid issue that had not been detected previously, though fixing it involved working around a limitation in narrowing the types of arrays that I've reported to phpstan in https://github.com/phpstan/phpstan/issues/14360

The other is to do with type narrowing / confusion in how our LoaderInterface and the GherkinFileLoader are tagged as generics. We were previously making an assumption that wasn't guaranteed to be correct. I've updated the generics tagging to fix this.

### DIFF
--- a/bin/update_i18n
+++ b/bin/update_i18n
@@ -24,6 +24,17 @@ foreach ($gherkinLanguages as $lang => $keywords) {
             $words = [$words];
         }
 
+        assert(array_find($words, fn ($word) => is_string($word)) === null, 'Expected $words to be an array of strings');
+        /**
+         * Here we force phpstan to recognise that we have narrowed the type of $words to a list of strings with the
+         * assertion above. Otherwise it will report an error from the later `implode()` call that $words must be a
+         * list<string>.
+         *
+         * There does not currently seem to be a way to force phpstan to detect this type narrowing at runtime
+         * https://github.com/phpstan/phpstan/issues/14360
+         *
+         * @var array<string> $words
+         */
         if ($type === 'scenarioOutline') {
             $type = 'scenario_outline';
         }
@@ -32,7 +43,6 @@ foreach ($gherkinLanguages as $lang => $keywords) {
             $formattedKeywords = [];
 
             foreach ($words as $word) {
-                assert(is_string($word));
                 $formattedWord = trim($word);
 
                 if ($formattedWord === $word) {
@@ -45,12 +55,14 @@ foreach ($gherkinLanguages as $lang => $keywords) {
             $words = $formattedKeywords;
         }
 
-        usort($words, static function ($type1, $type2) {
-            assert(is_string($type1));
-            assert(is_string($type2));
+        foreach ($words as $word) {
+            assert(is_string($word));
+        }
 
-            return [mb_strlen($type2, 'utf8'), $type1] <=> [mb_strlen($type1, 'utf8'), $type2];
-        });
+        usort(
+            $words,
+            static fn (string $type1, string $type2) => [mb_strlen($type2, 'utf8'), $type1] <=> [mb_strlen($type1, 'utf8'), $type2]
+        );
 
         $langMessages[$type] = implode('|', $words);
     }
--- a/bin/update_i18n
+++ b/bin/update_i18n
@@ -24,7 +24,7 @@ foreach ($gherkinLanguages as $lang => $keywords) {
             $words = [$words];
         }
 
-        assert(array_find($words, fn ($word) => is_string($word)) === null, 'Expected $words to be an array of strings');
+        assert(array_find($words, fn ($word) => !is_string($word)) === null, 'Expected $words to be an array of strings');
         /**
          * Here we force phpstan to recognise that we have narrowed the type of $words to a list of strings with the
          * assertion above. Otherwise it will report an error from the later `implode()` call that $words must be a
--- a/src/Gherkin.php
+++ b/src/Gherkin.php
@@ -151,6 +151,12 @@ class Gherkin
     {
         foreach ($this->loaders as $loader) {
             if ($loader->supports($resource)) {
+                /**
+                 * If the loader supports the provided resource, then it can safely be typed as
+                 * LoaderInterface<ResourceType>.
+                 *
+                 * @var LoaderInterface<TResourceType> $loader
+                 */
                 return $loader;
             }
         }
--- a/src/Loader/LoaderInterface.php
+++ b/src/Loader/LoaderInterface.php
@@ -24,11 +24,9 @@ interface LoaderInterface
     /**
      * Checks if current loader supports provided resource.
      *
-     * @template TSupportedResourceType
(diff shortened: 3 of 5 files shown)


tests/Loader/GherkinFileLoaderTest.php
```diff
@@ -93,26 +99,76 @@ public function testParsingCachedFeature(): void
...
-        $this->loader->setBasePath($this->featuresPath . '/../');
-        $this->assertFalse($this->loader->supports('features'));
-        $this->assertFalse($this->loader->supports('tables.feature'));
-        $this->assertTrue($this->loader->supports('features/tables.feature'));
+        $actual = [];
+        foreach (array_keys($expected) as $resource) {
+            $actual[$resource] = $this->loader->supports($resource);
+        }
 
-        $this->assertTrue($this->loader->supports('features/pystring.feature'));
```
acoulton, line 106: On the first `assertTrue($this->loader->supports(...))` call, phpstan was narrowing the type of `$this->loader` to `Behat\Gherkin\Loader\GherkinFileLoader&Behat\Gherkin\Loader\LoaderInterface<'features/tables.feature'>`.

The second call then caused phpstan to decide that `$this->loader` was a `*NEVER*` type.

And this then caused it to decide that the next call to `$this->loader->load(...)` would return `*ERROR*`. That then caused an argument type error when we passed it to `assertCount()`.

This appears to be because there's a conflict between:
* the `GherkinFileLoader` being tagged as `@extends AbstractFileLoader<string>`
* the `LoaderInterface::supports` being tagged with `@phpstan-assert-if-true =LoaderInterface<TSupportedResourceType> $this`

It's fine if you call the `->supports()` with a string variable, because that causes phpstan to infer it as `LoaderInterface<string>` which matches the class definition. But it causes a problem if you call it with hardcoded values, since phpstan then treats it as `LoaderInterface<'value of first string passed'>`.

I've created a simplified reproduction case in the phpstan playground 
https://phpstan.org/r/6250ecaa-ed3e-48dd-81cf-80d3cf25db3c but I'm not confident enough in phpstan generics to say whether this is a phpstan bug or an issue with our tagging. @stof, do you have an opinion?

stof: I think the issue is our generic type. `supports` should refine the type of `$resource` (to tell phpstan it is compatible with the type of the loader), not of `$this`.

acoulton: @stof thank you!

That makes sense. If I modify the example case to assert the type of the resource rather than the instance, that looks like it works:

https://phpstan.org/r/091134b1-4a3e-4514-a63d-17b7c6ccb795

bin/update_i18n
```diff
@@ -24,6 +24,17 @@ foreach ($gherkinLanguages as $lang => $keywords) {
             $words = [$words];
         }
 
+        assert(array_find($words, fn ($word) => is_string($word)) === null, 'Expected $words to be an array of strings');
```
stof: this assertion looks weird to me. Shouldn't it assert that it cannot find any non-string value instead ?

src/Loader/LoaderInterface.php
```diff
@@ -24,11 +24,9 @@ interface LoaderInterface
     /**
      * Checks if current loader supports provided resource.
      *
-     * @template TSupportedResourceType
+     * @param mixed $resource Resource to load
      *
-     * @param TSupportedResourceType $resource Resource to load
-     *
-     * @phpstan-assert-if-true =LoaderInterface<TSupportedResourceType> $this
+     * @phpstan-assert-if-true =TResourceType $resource
```
acoulton, line 29: As @stof identified, calling `supports('foo')` doesn't guarantee that the instance **only** accepts 'foo', so we can't safely refine the instance itself to `LoaderInterface<'foo'>`. Our instance could as easily be (probably is) `LoaderInterface<string>` or an even wider type.

However we can safely refine `$resource` - if it passes `supports()` then it must be a type of `TResourceType` and safe to pass to `->load($resource)`.

src/Gherkin.php
```diff
@@ -151,6 +151,12 @@ public function resolveLoader(mixed $resource)
     {
         foreach ($this->loaders as $loader) {
             if ($loader->supports($resource)) {
+                /**
+                 * If the loader supports the provided resource, then it can safely be typed as
+                 * LoaderInterface<ResourceType>.
+                 *
+                 * @var LoaderInterface<TResourceType> $loader
+                 */
```
acoulton, line 159: We still need to narrow the type returned from `resolveLoader`, because `$this->loaders` is `LoaderInterface<*>` and the method needs to return `LoaderInterface<TResourceType>`.

Here, we assume that the caller will call `->resolveLoader($resource)->load($resource)` for each resource - they won't attempt to reuse the loader for other resources.

Therefore, it doesn't matter if the type we return is stricter than the actual implementation of the loader we're returning, so long as it accepts the type of resource that was passed in. So the easiest way to do this is just to explicitly tag it the specialised type for phpstan.

acoulton:
@stof I've reverted the test refactoring and (I think) fixed the generic / phpstan tagging instead.

bin/update_i18n
```diff
@@ -24,6 +24,17 @@ foreach ($gherkinLanguages as $lang => $keywords) {
...
+        /**
+         * Here we force phpstan to recognise that we have narrowed the type of $words to a list of strings with the
+         * assertion above. Otherwise it will report an error from the later `implode()` call that $words must be a
+         * list<string>.
+         *
+         * There does not currently seem to be a way to force phpstan to detect this type narrowing at runtime
+         * https://github.com/phpstan/phpstan/issues/14360
+         *
+         * @var array<string> $words
```
stof, line 36: this `@var` should probably go near the assignment of the value (on line 24) instead.

bin/update_i18n
acoulton: Urgh, yeah of course it should. I'd tried a whole range of options for trying to get phpstan to narrow the type itself without success, and mangled the one I'd settled on when I added it back to this file. Will fix.

bin/update_i18n
acoulton: Unfortunately it can't go higher up because if it's before the `assert(array_filter(...))` then phpstan will complain that the assertions are redundant https://phpstan.org/r/5e86b039-21e8-4bb9-82a1-4877aef0c1c2

Unless we only tag it, and don't assert at runtime at all, but that feels a bit like cheating

bin/update_i18n
acoulton: Fixed.

acoulton:
@stof when you have a moment please could you review my responses / fixes to your comments here?

I could do with getting master green again as I've hit a challenge with the compatibility mode that I think I'll need to solve in this repo.

stof (approved):

acoulton:
Thanks @stof

PR #416, opened 2026-04-22
fix: Ignore whitespace around operators in tag filter strings

### LINKED ISSUE #410: Issue with multiple tag-filters
Recently when upgrading from `4.15.0` to `4.16.1` I noticed that behat was ignoring all of my tests when running them directly.

My behat-configuration looked similar to this in order to ignore all tests with the tags `@foo` and `@bar`:
```yaml
default:
    ...
    gherkin:
        filters:
            tags: "~@foo && ~@bar"
```

On `4.15.0` this seems to be working fine, but after the upgrade this seems to filter out tests that don't have either tag set.
I've created a minimal example:

Suppose I have the following .feature file:

```cucumber
# features/test.feature
@mink
Feature: Public Login
  Scenario: Logging in
    Given I am on "/en/login"
```
When running it via `vendor/bin/behat features/test.feature --tags "~@foo && ~@bar"` this seems to ignore all Scenarios in `4.16.1` but works fine with `4.15.0`.

Is this an issue on my side (e.g. configuration that could be wrong) or a bug?

acoulton:
Whitespace is allowed within a tag filter expression either side of the `&&` or `,` operators.

This reinstates historical behaviour where these strings were allowed.

The deprecation on filters containing whitespace is now only triggered if the tags themselves contain whitespace, after trimming any space around operators.

I began this PR with some refactoring and expansion of the test cases, to make it easier to see exactly what tag syntax is tested / supported, and to cover cases and edge cases that weren't previously tested.

Fixes #410

### DIFF
--- a/tests/Filter/TagFilterTest.php
+++ b/tests/Filter/TagFilterTest.php
@@ -76,6 +76,12 @@ class TagFilterTest extends TestCase
             ['@wip,@vip&&@user', ['vip'], false],
             ['@wip,@vip&&@user', ['wip', 'user'], true],
             ['@wip,@vip&&@user', ['vip', 'user'], true],
+
+            // `&&` with negated tag matches if only the first tag is present
+            ['@wip&&~@slow', [], false],
+            ['@wip&&~@slow', ['wip'], true],
+            ['@wip&&~@slow', ['wip', 'fast'], true],
+            ['@wip&&~@slow', ['wip', 'slow'], false],
         ];
     }
 
@@ -105,6 +111,25 @@ class TagFilterTest extends TestCase
             // `&&` matches if one tag present on the feature and one on the scenario
             ['@feature-tag&&@user', ['feature-tag'], ['wip', 'user'], true],
             ['@feature-tag&&@user', ['feature-tag'], ['wip'], false],
+
+            // Does not match if the feature matches a `~` tag
+            ['@user&&~@feature-tag', [], [], false],
+            ['@user&&~@feature-tag', ['feature-tag'], ['user'], false],
+            ['@user&&~@feature-tag', ['other-feature'], ['user'], true],
+            ['@user&&~@feature-tag', ['other-feature'], ['api'], false],
+
+            // Matches if the feature or the scenario matches an OR expression
+            ['@api,@browser', [], [], false],
+            ['@api,@browser', ['api'], [], true],
+            ['@api,@browser', ['browser'], [], true],
+            ['@api,@browser', [], ['api'], true],
+            ['@api,@browser', [], ['browser'], true],
+            ['@api,@browser', ['api'], ['browser'], true],
+            ['@api,@browser', ['browser'], ['api'], true],
+
+            // Not affected if same tag is present on Feature and Scenario
+            ['@api', ['api'], ['api'], true],
+            ['@api', ['slow'], ['slow'], false],
         ];
     }
 
@@ -139,6 +164,10 @@ class TagFilterTest extends TestCase
                 '@wip',
                 true,
             ],
+            'no match if the Outline does not match regardless of Examples' => [
+                '@etag2&&~@wip',
+                false,
+            ],
             'match if tags present on Outline & ANY Examples' => [
                 '@wip&&~@etag3',
                 true,
@@ -146,6 +175,10 @@ class TagFilterTest extends TestCase
             'match if tags present on Feature, Outline & ANY Examples' => [
                 '@feature-tag&&@etag1&&@wip',
                 true,
+            ],
+            'no match if the Feature does not match regardless of Examples' => [
+                '@etag2&&~@feature-tag',
+                false,
             ],
             'match if tags present on Feature & Outline & ALL Examples match the NOT filter' => [
                 '@feature-tag&&~@etag11111&&@wip',
@@ -159,6 +192,10 @@ class TagFilterTest extends TestCase
                 '@feature-tag&&@etag2',
                 true,
             ],
+            'match if tags present on Feature & ANY Examples' => [
+                '@feature-tag&&@etag3',
+                true,
+            ],
             'no match if ALL Examples match ONE of the NOT filters' => [
                 '~@etag1&&~@etag3',
                 false,
@@ -167,6 +204,10 @@ class TagFilterTest extends TestCase
                 '@etag1&&@etag3',
                 false,
             ],
+            'match if ANY Examples match an OR filter' => [
+                '@etag1,@etag3',
+                true
+            ]
         ];
     }
 
--- a/tests/Filter/TagFilterTest.php
+++ b/tests/Filter/TagFilterTest.php
@@ -47,147 +47,140 @@ class TagFilterTest extends TestCase
         $this->assertSame([$matchedScenario], $filteredFeature->getScenarios());
     }
 
-    public function testIsFeatureMatchFilter(): void
+    /**
+     * @return array<array{string, list<string>, bool}>
+     */
+    public static function providerFeatureMatches(): array
     {
-        $feature = new FeatureNode(null, null, [], null, [], '', '', null, 1);
+        return [
+            // Single tag matches if tag is present
+            ['@wip', [], false],
+            ['@wip', ['wip'], true],
 
-        $filter = new TagFilter('@wip');
-        $this->assertFalse($filter->isFeatureMatch($feature));
+            // Negated `~` tag matches if tag is NOT present
+            ['~@done', ['wip'], true],
+            ['~@done', ['wip', 'done'], false],
 
-        $feature = new FeatureNode(null, null, ['wip'], null, [], '', '', null, 1);
-        $this->assertTrue($filter->isFeatureMatch($feature));
+            // Or `,` matches if ANY of the list of tags is present
+            ['@tag5,@tag4,@tag6', ['tag1', 'tag2', 'tag3'], false],
+            ['@tag5,@tag4,@tag6', ['tag1', 'tag2', 'tag3', 'tag5'], true],
+            ['@tag5,@tag4,@tag6', ['tag1', 'tag2', 'tag3', 'tag5'], true],
 
-        $filter = new TagFilter('~@done');
-        $this->assertTrue($filter->isFeatureMatch($feature));
+            // And `&&` matches if ALL of the list of tags is present
+            ['@wip&&@vip', ['wip', 'done'], false],
+            ['@wip&&@vip', ['wip', 'done'], false],
+            ['@wip&&@vip', ['wip', 'done', 'vip'], true],
 
-        $feature = new FeatureNode(null, null, ['wip', 'done'], null, [], '', '', null, 1);
-        $this->assertFalse($filter->isFeatureMatch($feature));
-
-        $feature = new FeatureNode(null, null, ['tag1', 'tag2', 'tag3'], null, [], '', '', null, 1);
-        $filter = new TagFilter('@tag5,@tag4,@tag6');
-        $this->assertFalse($filter->isFeatureMatch($feature));
-
-        $feature = new FeatureNode(null, null, [
-            'tag1',
-            'tag2',
-            'tag3',
-            'tag5',
-        ], null, [], '', '', null, 1);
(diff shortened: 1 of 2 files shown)


tests/Filter/TagFilterTest.php
```diff
@@ -47,147 +47,140 @@ public function testFilterFeature(): void
...
-        ], null, [], '', '', null, 1);
-        $this->assertTrue($filter->isFeatureMatch($feature));
-
-        $filter = new TagFilter('@wip&&@vip');
-        $feature = new FeatureNode(null, null, ['wip', 'done'], null, [], '', '', null, 1);
-        $this->assertFalse($filter->isFeatureMatch($feature));
-
-        $feature = new FeatureNode(null, null, ['wip', 'done', 'vip'], null, [], '', '', null, 1);
-        $this->assertTrue($filter->isFeatureMatch($feature));
+        return [
```
stof: I suggest using a generator instead of returning an array. IMO it makes the data provider more readable as the `yield` keyword marks the start of a new case.

tests/Filter/TagFilterTest.php
```diff
@@ -76,6 +76,12 @@ public static function providerFeatureMatches(): array
             ['@wip,@vip&&@user', ['vip'], false],
             ['@wip,@vip&&@user', ['wip', 'user'], true],
             ['@wip,@vip&&@user', ['vip', 'user'], true],
+
+            // `&&` with negated tag matches if only the first tag is present
```
stof: this comment is confusing:

- the negated tag could be the first one
- it matches only if positive tag is present **and** negative tag is absent (i.e. both conditions match, which is the normal behavior of a `AND` operator)

tests/Filter/TagFilterTest.php
```diff
@@ -105,6 +111,25 @@ public static function providerScenarioMatches(): iterable
             // `&&` matches if one tag present on the feature and one on the scenario
             ['@feature-tag&&@user', ['feature-tag'], ['wip', 'user'], true],
             ['@feature-tag&&@user', ['feature-tag'], ['wip'], false],
+
+            // Does not match if the feature matches a `~` tag
```
stof: I would call that a negated condition. There is nothing like a `~ tag`. `~` is the `NOT` operator of our expressions.

tests/Filter/TagFilterTest.php
```diff
@@ -417,16 +455,42 @@ public function testItMatchesTagsParsedWithoutRemovingPrefix(string $filter, arr
...
-    private function expectDeprecationError(): void
+    /**
+     * @template T
+     *
+     * @param Closure():T $callable
+     * @param non-empty-string|false $expectDeprecation
+     *
+     * @return T
+     */
+    private function assertWhetherTriggersDeprecation(string|false $expectDeprecation, Closure $callable): mixed
```
stof: It would be better to use the native PHPunit deprecation handling here:

- use a dedicated test for cases expecting deprecations (instead of putting them in the same test that others)
- put the `IgnoreDeprecations` attribute on the test so that it does not report deprecations triggered during this test (as they are expected)
- use `$this->expectUserDeprecationMessage` or `expectUserDeprecationMessageMatches` to assert the deprecation

tests/Filter/TagFilterTest.php
```diff
@@ -394,6 +417,50 @@ public static function providerWhitespaceDeprecated(): array
...
+                '@tag1, @tag2 && ~@tag3',
+                'expectMatch' => true,
+                'expectDeprecation' => false,
+            ],
+            'allows all whitespace around operators' => [
+                // Very much an edge case, but the legacy implementation would have allowed this as it always just used
+                // `trim`. And arguably someone *could* have a config file with an indented multiline filter expression.
+                "\t@tag1,\n\t@tag2  &&  ~@tag3\n",
+                'expectMatch' => true,
+                'expectDeprecation' => false,
```
stof: Why is this going in a data provider named `providerWhitespaceDeprecated` when it covers a non-deprecated case ?

tests/Filter/TagFilterTest.php
acoulton: You're right, that's badly worded - started from a description of the example below rather than the behaviour. will fix.

tests/Filter/TagFilterTest.php
acoulton: So (and this answers your other point below about the phpunit deprecation handling) I much prefer to explicitly test both what *does* trigger a deprecation and what *does not* trigger it.

I also want to prove that even though the deprecation is triggered, the result is not affected.

Particularly in this case, where we are triggering a deprecation on whitespace but only in certain places in the input value, I think it is much easier to follow if there is a single test and set of examples that shows that.

I could of course have one test that tests values that trigger deprecations, and then rely on there being examples to some other test that don't trigger deprecations (and phpunit being configured to fail if they unexpectedly trigger the deprecation). But I think that's less clear.

You would not split `it returns true` or `it returns false` into separate tests (if using data providers). Therefore IMO you should not split `it does emit a deprecation` and `it does not emit a deprecation`.

I note BTW we were also not using PHPUnit's native deprecation handling before this PR either.

stof: for the fact that we don't have deprecations in tests that must not emit them, the native solution is to make the test suite fail on deprecations (which is also a native PHPUnit feature)

stof: > I note BTW we were also not using PHPUnit's native deprecation handling before this PR either.

that's probably related to the fact that those tests were written before our migration to PHPUnit 11+, and so in versions that did not have those handlings.



> You would not split `it returns true` or `it returns false` into separate tests (if using data providers). Therefore IMO you should not split `it does emit a deprecation` and `it does not emit a deprecation`.

We always split tests between "it returns a result" and "it throws an exception" to be able to use the native PHPUnit exception expectations. That's not different to me when wanting to expect deprecations.
Also, when wanting to fail on unwanted deprecations, that's necessary as we need to apply the `IgnoreDeprecations` attribute on tests for which deprecations must be ignored by the system deciding the test result.

And another reason for that split: when working on the next major version, tests covering deprecations will be changed in 2 potential ways (depending on the case):
- deleting them
- turning them into tests expecting an error (which require splitting them)

Having deprecated cases split from the start instead of being interleaved in the same data provider makes such maintenance work easier (we learned it in Symfony when working on removing deprecated layers every 2 years)

tests/Filter/TagFilterTest.php
acoulton: Fixed

tests/Filter/TagFilterTest.php
acoulton: Fixed

tests/Filter/TagFilterTest.php
stof: @acoulton why have you edited my comment, removing some parts of it and adding your own parts ?

Note that I restored my original comment.

acoulton: @stof apologies - I genuinely didn't think I had, but I can see I did. 

I responded from GitHub mobile, I am fairly sure that I'd done "Quote Reply" and looking at the flow again now I don't see how I can have confused that with "Edit" even if my thumb hit the wrong thing, as I would have had to add / change all the `>` quote markers and I don't remember doing that. 

But it doesn't seem that plausible that the GitHub app has posted an edit rather than a reply as that would be quite a fundamental bug.

So I think potentially I must have. I was rushing a bit. It was certainly not intentional. Thanks for spotting and restoring your comment, I'll post my reply below.

acoulton: >> You would not split `it returns true` or `it returns false` into separate tests (if using data providers). Therefore IMO you should not split `it does emit a deprecation` and `it does not emit a deprecation`.

> We always split tests between "it returns a result" and "it throws an exception" to be able to use the native PHPUnit exception expectations. That's not different to me when wanting to expect deprecations.

I would usually, but not always.

IMO for something like data validation where the result depends on possibly-subtle differences in the input value, I find it a lot clearer and more maintainable to have the valid and invalid cases together.

I might then conditionally call `->expectException` or I might just wrap the code in a try/catch, depending on how much detail I need to assert about the exception itself.

This also makes it trivial to refactor to add e.g. an `isValid($input): bool` method in future, because there is already a single data provider with all the cases so the only change is a single test method.

> Also, when wanting to fail on unwanted deprecations, that's necessary as we need to apply the `IgnoreDeprecations` attribute on tests for which deprecations must be ignored by the system deciding the test result.

The problem with `IgnoreDeprecations` is there's then a risk you don't detect **unexpected** deprecations. I realise in newer PHPUnit versions you can specify a matching pattern to reduce this risk, but you can (for now) only specify one, so that can be problematic if there are multiple possible deprecations for one set of input values.

Ultimately both `IgnoreDeprecations` and the global `failOnDeprecations` configuration rely on behaviour at a distance from the code under test. 

IMO that is fine if the deprecation is a side-effect of the test (expected or unexpected). But I prefer a more imperative / explicit assertion if the deprecation itself is the behaviour that is being tested.

> And another reason for that split: when working on the next major version, tests covering deprecations will be changed in 2 potential ways (depending on the case):
> - deleting them
> - turning them into tests expecting an error (which require splitting them)

> Having deprecated cases split from the start instead of being interleaved in the same data provider makes such maintenance work easier (we learned it in Symfony when working on removing deprecated layers every 2 years)

Not necessarily.

In the next major we will want to assert that filters that were previously deprecated now throw, but we will still also want to know that similar-but-valid filters (e.g. with whitespace in expected locations) are expected.

We can easily do that by changing the test method to assert an exception (either by conditionally calling `expectException` or by modifying the `assertWhetherTriggersDeprecation` to `assertWhetherThrowsInvalidFilter`).

The test cases do not need to change at all. That PR is then very easy to review, because it will be obvious that the valid / invalid syntax hasn't changed, only the behaviour when it is encountered.

---
tl;dr - IMO this is a difference of opinion rather than a case where either of us is definitively correct. I'm happy with the style I've used, which has worked well for me in the past, but if you feel strongly I am happy to take a different approach.

stof: the drawback of the current approach is that it replaces the PHPunit error handler by your own. So if the code triggers a PHP warning, it won't be handled by PHPUnit anymore.
There can be only one active error handler at a time in PHP.

> In the next major we will want to assert that filters that were previously deprecated now throw, but we will still also want to know that similar-but-valid filters (e.g. with whitespace in expected locations) are expected.

Whitespaces in expected locations are *not* about deprecated cases though.

acoulton: @stof I have updated this PR.

> Whitespaces in expected locations are not about deprecated cases though.

IMO there's an argument either way but I don't want to get bogged down, I have moved / extended the examples of supported whitespace to the main `providerFeatureMatches`.

> It would be better to use the native PHPunit deprecation handling here:

I've looked at this, and we can't do what you're suggesting.

`expectUserDeprecationMessage()` was added in phpunit 11.0.0. and we are on 10.x. 

Additionally IMO the code I have here is significantly clearer about **which** part of the code triggers the deprecation. The native alternative below doesn't specify (or prove) whether you get the deprecation when you create the filter, or only when you call `->isFeatureMatch`.

```php
$this->expectUserDeprecationMessage('Tags with whitespace');,
$tagFilter =  new TagFilter($filterString);
$feature = new FeatureNode(null, null, ['tag with space', 'tag1', 'tag2'], null, [], '', '', null, 1);
$this->assertSame($expectMatch, $tagFilter->isFeatureMatch($feature), 'Expected correct matching behaviour');
```

I suppose that could be split into two tests, one that only creates the TagFilter and a separate one with `IgnoreDeprecations` that asserts the result, but IMO that's also less clear and more complex than the solution in this PR (and risks ignoring unexpected deprecations as well as the expected one).

> the drawback of the current approach is that it replaces the PHPunit error handler by your own. So if the code triggers a PHP warning, it won't be handled by PHPUnit anymore. There can be only one active error handler at a time in PHP.

After 20-something years, I'm very aware that PHP only allows one error handler at a time :laughing: 

The error handler here will throw an exception if it gets anything other than the expected deprecation. Since we have PHPUnit configured to fail on PHP warnings, if the code does trigger a warning the outcome of the test run will be identical (except that the failure will be displayed / presented slightly differently).

stof (approved):

PR #417, opened 2026-04-22
fix: Emit deprecation if filter string does not contain @ prefixes 

### LINKED ISSUE #408: [BC break] Drop support for tag filter expressions without `@` prefixes
As per #407, we have always officially documented and expected that tag filter expressions would contain the `@` prefix e.g. `@wip&&~@slow`. This was the only syntax that was ever tested.

However, we do not validate that we can parse the tag expression, and the implementation prior to 4.16.0 had coincidentally accepted filters with or without the prefix.

We have reinstated backwards compatibility with those for the time being, but we should deprecate support for invalid tag expressions and officially remove it in the next major.

acoulton:
This PR builds on #416 - only 77b57cb8a7ed77eb66263cd054654482ad6c2002 is new in this branch.

When we added the BC layer to reintroduce support for filter strings without `@` prefixes, we didn't add a runtime deprecation because Behat didn't have a way to capture and report these.

However, since Behat does now have a deprecation collector / reporter (which does not fail the build by default) we can now safely emit the deprecation.

Fixes #408

### DIFF
--- a/src/Filter/TagFilter.php
+++ b/src/Filter/TagFilter.php
@@ -29,16 +29,11 @@ class TagFilter extends ComplexFilter
     public function __construct(string $filterString)
     {
         $filterString = trim($filterString);
-        $fixedFilterString = $this->fixLegacyFilterStringWithoutPrefixes($filterString);
-        // @todo trigger a deprecation here $filterString !== $fixedFilterString
-        $this->filterString = $fixedFilterString;
-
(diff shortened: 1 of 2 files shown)


stof (approved):

acoulton:
Force pushed just the new commit rebased after the squash-merge of #416, no changes to the original diff of this PR.

PR #418, opened 2026-04-22
refactor: Only parse tag filter expression once

acoulton:
Previously we were parsing (and normalising) the filter string in the constructor, and then again on every call to `isTagsMatchCondition`.

As well as being inefficient, this meant there was some duplication and coupling between the different places that we dealt with a string filter expression.

Instead, parse the filter once into an internal data structure, checking and normalising it at the same time.

This requires a small safety layer to ensure that behaviour does not change for child classes that are extending `TagFilter`, because the `filterString` property is protected and theoretically could be modified at runtime by a child class.

### DIFF
--- a/src/Filter/TagFilter.php
+++ b/src/Filter/TagFilter.php
@@ -27,7 +27,7 @@ class TagFilter extends ComplexFilter
     protected $filterString;
 
     /**
-     * @var array{all?: list<array{any: list<array{not: string}|array{has: string}>}>, filterString: string}
+     * @var array{all?: list<array{any: list<array{tag: string, hasTag: bool}>}>, filterString: string}
      */
     private array $parsedFilter;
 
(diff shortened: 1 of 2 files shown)


src/Filter/TagFilter.php
```diff
@@ -26,37 +26,61 @@ class TagFilter extends ComplexFilter
      */
     protected $filterString;
 
+    /**
+     * @var array{all?: list<array{any: list<array{not: string}|array{has: string}>}>, filterString: string}
```
stof: Instead of `list<array{not: string}|array{has: string}>}>`, we could use `list<array{tag: string, expected: bool}>}>`, which will allow simpler code during matching (your existing implementation basically recreates this tag name and boolean value from the structure).

acoulton: I actually started with something like that, but I wondered if `expected: false` was clear enough that we're explicitly saying it **must not** be present?

I'm also not sure `expected` is quite the right naming, as this is a filter rather than e.g. an assertion?

`list<array{tag: string, hasTag: bool}>` might work?

stof: `expected` was a quick proposal, but better names might be possible. `expectTag` or `hasTag` might be better indeed.

acoulton: Fixed

acoulton:
Updated this branch based on #416 and #417

acoulton:
@stof this is now rebased following the merge of #416 and #417 (no changes to the diff here) and ready to review.

stof (approved):

PR #419, opened 2026-04-27
fix: Wrap NoSuchLanguageException in a parser exception

### LINKED ISSUE #1800: Improve error output when feature file contains invalid language tag
In the new `gherkin-32` compatibility mode, the parser throws an exception if a feature file contains a `# language:` tag for an unknown language (previously, it silently fell back to English).

As seen in #1799, the error output in this case is not ideal:

```
In KeywordsDialectProvider.php line XX:

    Language not supported: martian
``` 

We would ideally catch this `NoSuchLanguageException` and render an error message that includes the path to the invalid feature file. In an ideal world, this would be treated like a failure of a single feature (rather than an outright fatal error) but I suspect that will be difficult to achieve.

stof:
This error provides better error reporting about where the error happens (similar to what happens for other invalid states).

Closes https://github.com/Behat/Behat/issues/1800

### DIFF
--- a/src/Parser.php
+++ b/src/Parser.php
@@ -14,6 +14,7 @@ use Behat\Gherkin\Exception\FilesystemException;
 use Behat\Gherkin\Exception\InvalidTagContentException;
 use Behat\Gherkin\Exception\LexerException;
 use Behat\Gherkin\Exception\NodeException;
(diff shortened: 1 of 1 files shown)


acoulton (approved):
Thanks @stof

PR #421, opened 2026-05-17
feat: Add a `getFullText()` method to a StepNode

acoulton:
This provides a gherkin representation of the step text, including the keyword. This can for example be used by formatters that need to render the step, without them needing to take account of whether the keyword should be followed by a space character in the current language.

The result will be consistent in all gherkin parsing modes.

### DIFF
--- a/src/Node/StepNode.php
+++ b/src/Node/StepNode.php
@@ -23,6 +23,8 @@ class StepNode implements NodeInterface
 {
     private readonly string $keywordType;
 
+    private readonly string $fullText;
+
     /**
      * @param ArgumentInterface[] $arguments
      */
@@ -32,6 +34,7 @@ class StepNode implements NodeInterface
         private readonly array $arguments,
         private readonly int $line,
         ?string $keywordType = null,
+        ?string $fullText = null,
     ) {
         if (count($arguments) > 1) {
             throw new NodeException(sprintf(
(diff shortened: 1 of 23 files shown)


src/Node/StepNode.php
```diff
@@ -43,6 +46,7 @@ public function __construct(
         }
 
         $this->keywordType = $keywordType ?: 'Given';
+        $this->fullText = $fullText ?? $keyword . ' ' . $text;
```
acoulton, line 49: This provides a sane default for backwards compatibility if external code creates a `StepNode`. It will match the actual value we would parse in almost all languages.

carlos-granados (approved):
Looks good and well tested, thanks @acoulton

src/Node/StepNode.php
stof: it will not actually match in gherkin-32 mode, which does not trim keywords

acoulton: Correct, but in gherkin-32 mode we are creating the StepNode and passing in the `$fullText` taking account of that.

This is just a fallback for BC with third-party code etc (which is unlikely to know anything about `gherkin-32` mode, and where we have no knowledge or control over what is actually in the strings they're passing in). 

I don't expect it to run IRL and if it did worst-case the string value would contain two spaces instead of one.

stof (approved):

Issue #362, opened 2025-05-26
[RFC] Deprecate the YamlFileLoader and the ArrayLoader


stof:
Loading features from Yaml files is not something that is needed by Behat. And the format of those Yaml files (which is the array-based format of the ArrayLoader used internally by the YamlFileLoader) is not specified (nor documented) anyway.

AFAICT, this loader is mostly used by our testsuite, to load the reference files for comparison (which means it should probably never have been part of the package but only of the testsuite...).

I suggest deprecating those 2 classes.

Our tests fixtures used by ParserTest should first be migrated to the gherkin testdata format (which will also help for upstreaming them)

acoulton:
Thanks, I was going to ask about that as I wasn't sure of the history of ArrayLoader but it looked like it was indeed mostly a test class.

I'd agree with migrating the parser examples to NDJson and deprecating them, I can do that.

PR #427, opened 2026-08-19
test: Migrate ParserTest fixtures to the cucumber gherkin testdata format

### LINKED ISSUE #389: Migrate all ParserTest fixtures to the gherkin testdata format
Now that we have support for using custom tests written using the cucumber gherkin testdata format, all our existing fixtures used by ParserTest should be converted to the gherkin testdata format.

This will unblock #362 (and might unblock contributing some cases upstream to increase coverage)

Amoifr:
Closes #389.

This converts all 34 `ParserTest` fixture pairs (`tests/Fixtures/features/*.feature` + `tests/Fixtures/etalons/*.yml`) to the cucumber gherkin testdata format in `tests/Cucumber/extra_testdata/good/`, following the approach of #322:

- the `.ast.ndjson` expectations were generated with the official cucumber `gherkin-javascript` CLI (`--predictable-ids --no-source --no-pickles`, piped through `jq --sort-keys -c` to match the existing file format), so they are real cucumber parser output, not dumps of our own parser;
- the official parser accepts every one of the 34 features, so they all live in `good/`;
- two fixtures collided with upstream testdata names and were renamed after what they actually test: `background.feature` -> `background_and_untitled_scenario.feature` and `empty.feature` -> `comment_without_trailing_newline.feature`;
- 7 fixtures parse differently in `legacy` mode only (zero divergence in `gherkin-32`): multiline titles folded into the title (5) and trailing whitespace trimmed from description lines (2). They are documented in `$notParsingCorrectly` and their `expected_variants` were recorded with `RE_RECORD_EXPECTATIONS`; I checked the recorded variants against the old YAML etalons, the semantics are identical.

Each fixture is now exercised in both compatibility modes instead of one, and the previously etalon-covered behaviour is protected either by the cucumber expectations or by the recorded variants.

The loader tests used to share the fixtures folder, so they now have their own minimal `tests/Loader/Fixtures/` (three features and the `addition.yml` etalon they assert against). With that, `tests/Fixtures/features` and `tests/Fixtures/etalons` are gone and `ParserTest` no longer uses `YamlFileLoader` at all, which unblocks #362.

Note: PHPStan currently reports a pre-existing `offsetAccess.invalidOffset` on `src/Node/TableNode.php:156` that is unrelated to this change (master has not been rebuilt since the gherkin v40 / dependency bumps); happy to address it separately if useful.


tests/Cucumber/expected_variants/legacy/hashes_in_quotes.feature.expected.yaml
```diff
@@ -0,0 +1,93 @@
+Behat\Gherkin\Node\FeatureNode:
+    title: "Some '#quoted' string"
```
acoulton, line 2: This expectation fails in the [PHP 8.1 / symfony/yaml:^5.4](https://github.com/Behat/Gherkin/actions/runs/32265036546/job/99916004344#step:5:598) job - I am guessing that is a difference between the YAML dumping between symfony versions rather than our coding. Presumably need to see if we can force the older symfony/yaml version to encode it consistently in the captured expectations?

acoulton (commented):
Thanks @Amoifr this is great.

Do you have time to look at how to get the failing test to pass on all supported symfony versions?

I see that the coverage reduction is all in the ArrayLoader that we'll be deprecating, so that's no issue.

It looks like the phpstan issue you mentioned has resolved itself as that is passing in CI.

tests/Cucumber/expected_variants/legacy/hashes_in_quotes.feature.expected.yaml
Amoifr: Your guess was right. Reproduced locally with `composer update --with=symfony/yaml:^5.4`, which is exactly what the job installs, and it is the only failure. 5.4 dumps a string containing a `#` as `'Some ''#quoted'' string'`, later versions as `"Some '#quoted' string"`. Both parse to the same value, and `hashes_in_quotes` is the only migrated fixture whose title contains a `#`.

I could not find a way to force the older dumper into the newer style: the quoting is chosen internally by `Inline::dump()` and no `Yaml::DUMP_*` flag reaches it. So rather than pin the expectation to one version, I normalise: `ParserResultDumper::normalize()` re-dumps the recorded file with whichever YAML component is installed, so both sides of the comparison go through the same dumper and the quoting style stops being significant. No expectation file changes.

One gotcha worth knowing if you touch this later: `replaceAbsolutePaths()` runs after the dump, and `file: {BASEDIR}/vendor/...` is not valid inline YAML, so parsing a recorded file straight away fails with `Malformed inline YAML string`. `normalize()` puts the real path back before parsing and the placeholder is reapplied on the way out.

In the test, `assertStringEqualsFile` became `assertFileExists` plus `assertSame`, which keeps the `RE_RECORD_EXPECTATIONS` flow working for a missing file.

The trade off, so you can overrule me: the comparison no longer locks the exact formatting of the recorded file, only its content. An expectation reformatted by hand but semantically identical would now pass. That felt right for a test that is about the parser result rather than about the dumper, but say the word if you would rather keep it byte for byte and solve it another way.

Green on symfony/yaml 5.4.53, 6.4.45 and 7.4.18 locally, 1099 tests, and PHPStan is clean.

acoulton (approved):
Thanks @Amoifr - normalising the YAML is a good solution. I don't think it matters that the format isn't guaranteed to match the file on disk - we're really only concerned about the values inside our objects. I just matched on the YAML content as it's a good format for serialising our data and seeing the diff.

carlos-granados (approved):
Thanks @Amoifr

PR #430, opened 2026-09-03
feat: Support `Rule` nodes in GHERKIN_32 mode (backwards compatible)

acoulton:
> [!CAUTION]
> This PR needs to go in the next release, as it contains breaking changes to the protected methods introduced in #436 which is now merged.

This adds support for the `Rule:` keyword/node when using the opt-in GHERKIN_32 parsing mode. This addresses #140 and also Behat/Behat#1451.

The implementation is fully backwards compatible - test runners that call `FeatureNode::getScenarios` expecting an array of `ScenarioInterface[]` will still receive a flat array of `ScenarioNode|OutlineNode`. They will never receive the new `RuleNode` objects from any existing API method.

Instead, `FeatureNode::getScenarios()` hoists any Scenario / Outline from inside a Rule as though it had been written as part of the top-level Feature. As part of this hoisting, any steps from a Rule Background are collapsed into (duplicated across) the start of each Scenario.

## Backwards compatibility - test running

Existing test runners will not see the detail of the Rule groupings, or the division between the Background and Scenario parts of the Rule (the Feature background is still separate, for BC).

This has no real impact on running tests - it just takes us in the same direction as Cucumber, where the `pickles` passed to the test runner contain the combined list of steps to run for each test case.

All hoisted nodes contain the same line number as the original node.

Hoisted scenarios merge in the tags from their parent Rule. This is to provide BC for things like the Behat/MinkExtension logic that selects a driver based on Feature or Scenario tags - the user would expect these to also apply tags from the Rule.

## Backwards compatability - output formatting

Formatters that render e.g. JUnit / JSON / lists of tests will not be affected. They will not have access to `Rule` information, but will receive exactly the same dataset of scenarios, steps and line numbers as now.

Formatters that render a gherkin-style output (e.g. Behat's `pretty` formatter) will render without breaking. However, the grouping of Scenarios and Steps will not match the gherkin source file.

It is possible to upgrade an individual formatter to display Rule information even if the test runner itself is using hoisted nodes. The formatter just needs to call `FeatureNode::getChildren` and then match Scenarios by e.g. the line number in the source. Again, this is similar to how official Gherkin formatters have to match the execution results from `pickles` back to the (separate) `AST` from the parser.

## Backwards compatibility - filtering

Filters implemented within this package will be updated:

* [x] `TagFilter` will take account of any tags on a `Rule` parent when filtering Scenarios, Outlines & ExampleTables in `filterFeature`.
* [x] `LineFilter` will include all Scenarios from a `Rule` if given the line number of the Rule.
* [x] `LineRangeFilter` will include all Scenarios from a `Rule` if they are in the range - the Rule's own line number is considered during `filterFeature`.
* [x] `NameFilter` will match against the Rule title & description, as well as the feature / scenario.
* [x] `NarrativeFilter` currently only filters on Feature description, so does not change.
* [x] `PathsFilter` currently only filters on Feature file path, so does not change.
* [x] `RoleFilter` only filters on Feature description, so does not change.

End-user Filters will work as previously, they just won't see or be able to filter against anything related to the `Rule`. If they call `FeatureNode::getScenarios()` to filter individual scenarios then they will receive (and return) the hoisted scenarios, dropping the Rule node entirely.

External filter implementations can migrate to using `FeatureNode::getChildren()` to get the new node tree including Rule nodes.

## Edge case - custom node classes

Our API uses `ScenarioInterface` to typehint Feature children, rather than the built-in `ScenarioNode|OutlineNode` implementations. We do not expect end-users to provide their own implementations of this interface, but technically speaking they could. 

However, the hoisting logic in `getScenarios` relies on being able to create modified versions of the original nodes. We can't do that with classes we don't control.

Therefore if the user enables the new Gherkin mode, and they have a scenario using Rules, and they are parsing to a custom `ScenarioInterface` for some reason, `FeatureNode::getScenarios` may throw an exception. I do not expect this to happen in reality.

## Edge case - `isScenarioMatch` does NOT consider parent Rules

Attempting to reverse from an arbitrary `ScenarioInterface` back to its parent `RuleNode` is complex, because:

* We can't guarantee everyone is using our Parser, so we can't rely on tracking the tree at parse time.
* There are lots of places (in our own code, and in external code) that create a new Scenario from an existing one, so we can't be sure to track this or match on object identity.
* We don't have a unique `id` for Scenarios that relates them to their source. Even in methods that are called with a `FeatureNode` and a `ScenarioInterface`, we can't *guarantee* that the Scenario is a child of the Feature. So it's tricky to iterate the Feature and find a matching child - line number isn't necessarily safe to match on.

Since we can't conclusively guarantee that a filter's `isScenarioMatch($scenario)` or `isScenarioMatch($feature, $scenario)` will **always** be able to find the Rule parent, it is safer to define that it will **never** consider the Rule when called standalone.

Instead, we only consider the Rule when in `filterFeature` - in this case we're iterating the full tree so we can be certain of parentage.

This is slightly inconsistent, but it matches e.g. the current behaviour of `NameFilter`, which considers the parent Feature text during `filterFeature` but does *not* consider it in `isScenarioMatch`.


src/Node/FeatureNode.php
```diff
@@ -125,11 +127,82 @@ public function hasScenarios()
...
+                throw new UnexpectedValueException('Cannot merge rule background and scenario steps for custom ScenarioInterface ' . $child::class);
+            },
+            $node->getChildren())
+        );
+    }
+
+    /**
+     * @return list<BackgroundNode|RuleNode|ScenarioInterface>
+     */
+    public function getChildren(): array
```
acoulton: I went for `getChildren()` for the new API for both `FeatureNode` and `RuleNode` as this matches Cucumber's own semantics - both a feature and a rule have a single list of `children`.

The `children` can include zero or one `BackgroundNode` - but it must be the first item in the list.

In practice this makes the `getChildren` API a bit awkward to use. Almost everywhere ends up having to filter / typecheck it to separate out `BackgroundNode` from any scenarios / rules - in some cases just to satisfy phpstan in cases where I know we can't in reality hit a BackgroundNode.

It would probably be easier for everyone to have separate getters for the Background (if any) and for any other type of children. This would also reduce the number of repeated iterations of the same arrays.

I just need to come up with some sensible naming for that getter that doesn't conflict with the existing methods, and doesn't box us in for the future.

src/Node/OutlineNode.php
```diff
@@ -47,6 +47,24 @@ public function __construct(
         $this->tables = is_array($tables) ? $tables : [$tables];
     }
 
+    /**
+     * @param list<StepNode> $steps
+     *
+     * @internal
```
acoulton, line 51: I've realised these `withXX` methods can't be `@internal` as a third-party filter implementation would also need to be able to call them.

tests/Filter/TagFilterTest.php
```diff
@@ -375,6 +380,43 @@ public function testFilterFeatureWithTaggedExamples(): void
         $this->assertEquals([$exampleTableNode3], $scenarioInterfaces[1]->getExampleTables());
     }
 
+    public function testFilterFeatureKeepsMatchingEmptyOutline(): void
+    {
+        // Note the existing / historic behaviour:
+        // - if an Outline had examples, but none matched the filter, the whole Outline is removed
+        // - if an Outline had no examples, but itself matches the filter, the empty Outline is kept
```
acoulton, line 384: I found this slightly-surprising behaviour of the existing TagFilter implementation during refactoring, so added a test to make sure it didn't / doesn't change unexpectedly.

tests/Filter/TagFilterTest.php
```diff
@@ -375,6 +380,43 @@ public function testFilterFeatureWithTaggedExamples(): void
...
+                title: 'Outline with examples',
+                tags: [],
+                tables: [
+                    StubNode::exampleTable(
+                        tags: ['@bar'],
+                        name: 'Example table'
+                    ),
+                ]
+            ),
+        ]);
```
acoulton, line 401: I've introduced the `StubNode::{type}` helper methods in this PR. There's a factory method for each type of node, which mirrors the constructor signature but with all properties optional and provided with sane defaults.

I like this style of creating DTOs in tests because:

* It focuses the test example on the DTO properties that are actually relevant, rather than lots of noise for parameters that might be required but don't affect behaviour.
* It decouples the tests from the DTO constructor signature - so if we add or change properties in future, we only have to update tests where that is relevant rather than amending every `new ScenarioNode(...)` line throughout test cases.

Hopefully you like it. If so, I can do a separate PR to use this for creating DTOs in the existing tests after this merges. If not, I can obviously refactor to just use `new XXXNode(...)` everywhere in these tests but they'll be a lot more verbose.

tests/Filter/TagFilterTest.php
```diff
@@ -545,6 +587,159 @@ public function setFilterString(string $filterString): void
...
+        ];
+
+        yield 'matches by tags on child scenarios' => [
+            '@slow',
+            <<<'EOS'
+            Rule: Admin rule
+              - Background: Admin background
+              - Scenario: Slow admin scenario
+            Rule: Users rule
+              - Scenario: Slow user scenario
```
acoulton: I think asserting against a simple stringified version of the expected feature children feels like a more immediately obvious way of showing which nodes should have made it through filtering than e.g. passing in a full tree of objects to compare. It also makes the diff much less verbose and therefore more obvious if the assertion fails.

acoulton:
@stof @carlos-granados this is NOT ready for any detailed review - it's not finished and there's definitely things in the implementation I want to improve.

However if you have time I'd appreciate any high-level opinions on the direction I'm taking this / the decisions I've pulled out in comments above :)

src/Node/FeatureNode.php
```diff
@@ -125,11 +127,82 @@ public function hasScenarios()
...
+     * @deprecated use getChildren() for first-class handling of Rule nodes
      */
     public function getScenarios()
     {
-        return $this->scenarios;
+        // It is safe for this cache to be static, because the result is always the same for a given RuleNode,
+        // and the WeakMap will automatically clean up any references to RuleNodes that are no longer in use.
+        // Making it static and private to this method also eliminates any potential issues with serialising
+        // FeatureNode e.g. for caching or tests.
+        static $ruleScenariosCache;
```
stof: instead of a static inline variable which is easy to miss, I would make it a private static property of the class, which makes it easier to notice we have such static state (and is equivalent in term of lifetime)

src/Filter/TagFilter.php
```diff
@@ -131,28 +133,72 @@ private function parseFilterString(): void
      */
     public function filterFeature(FeatureNode $feature)
     {
-        $scenarios = [];
-        foreach ($feature->getScenarios() as $scenario) {
-            if (!$this->isScenarioMatch($feature, $scenario)) {
+        $filteredChildren = [];
+        foreach ($feature->getChildren() as $child) {
```
stof: All those changes could be reverted if the flattening of rules in `FeatureNode::getScenarios` was merging the rule tags in the scenario tags.

src/Node/FeatureNode.php
acoulton: Good point, thanks

src/Filter/TagFilter.php
acoulton: This is kind-of true, except:

* I might need to do similar things to support the other filters - we're not e.g. going to merge the Rule's line number or description into the hoisted nodes. So hopefully I can make this cleaner.
* Our filters work by returning a new FeatureNode with the filtered children, which then passes to the next filter and eventually to the caller. If the filter uses `FeatureNode::getScenarios` then the returned `FeatureNode` will only contain the hoisted scenarios, and will lose all the Rule information. Therefore I think all our in-built filters need to be refactored to use `FeatureNode::getChildren` so that we're still returning the full node structure at the point the object leaves our library.

acoulton:
I've stubbed a very initial proof-of-concept (literally just switch to my fork and add a few features to prove it works without runtime changes) in Behat/Behat#1879

src/Node/FeatureNode.php
acoulton: I've gone with a dedicated method for children that represent one or more executable scenarios (e.g. not including `Background`, which is not executable in its own right).

*`FeatureNode::getExecutableChildren(): list<ScenarioInterface|RuleNode>`
* `RuleNode::getExecutableChildren(): list<ScenarioInterface>`

This naming should cover us if we either need to introduce new kinds of executable children in future, or if Feature or Rule ever gets more children that are *not* executable.

acoulton:
I have a few small bits on this still to finish / tidy up - but I'm now confident that we can implement `Rule` support in a way that is fully BC for existing consumers.

However, I'm now getting quite blocked on the other open PRs as they all have an impact one way or another on this branch.

@carlos-granados @stof if one of you has some time to review other Behat/Gherkin PRs so we can get them merged (or changed) that would be really helpful. I'll then be able to get this finished off and ready for review.

src/Filter/TagFilter.php
```diff
@@ -90,10 +100,15 @@ public function isFeatureMatch(FeatureNode $feature)
      */
     public function isScenarioMatch(FeatureNode $feature, ScenarioInterface $scenario)
     {
-        // Note, we can't refactor this method / filterFeature to avoid iterating the tables twice because that
-        // would break end-user assumptions about the relationship between these two methods if they have extended
-        // either method.
-        $tags = [...$feature->getTags(), ...$scenario->getTags()];
+        // Note, this will only consider Rule tags when filtering features using our implementation of `filterFeature`.
+        // It will not consider Rule tags if:
+        // - this method is called from outside our main `filterFeature` loop (e.g. to check a tagged hook in Behat)
```
acoulton, line 105: As it stands with this PR:

* A tag filter works as expected (e.g. considers Rule-level tags) when filtering which scenarios should be executed.
* It **does not** consider Rule-level tags if `isScenarioMatch` is called directly with the ScenarioNode that was parsed directly by us.

Behat calls this method to check whether to run tagged hooks - in practice at the moment, that will still match the ScenarioInterface that Behat passes in. That's because Behat is using the "hoisted" scenarios from the BC layer, so any `RuleNode` tags are merged into the `ScenarioNode` before execution.

I have a couple of ideas for improving this in future, but this PR is already quite big and I think we should review & merge it as is and I can iterate on this separately.

src/Node/FeatureNode.php
```diff
@@ -175,7 +236,7 @@ public function getLine()
     /**
      * Returns a copy of this feature, but with a different set of scenarios.
      *
-     * @param array<array-key, ScenarioInterface> $scenarios
+     * @param array<array-key, ScenarioInterface|RuleNode> $scenarios
      */
     public function withScenarios(array $scenarios): self
```
acoulton, line 241: Arguably, this method should be renamed (deprecated & replaced with a new one) - but I think that could be a separate PR - the internal `scenarios` property should potentially also be renamed. There are a few naming / logic things around whether that should take the Background and other children in one set, or whether it should be just the "executable" children. IMO it'd be easier to discuss & review that all separately, rather than blocking this.

src/Node/RuleNode.php
```diff
@@ -0,0 +1,111 @@
...
+
+    /**
+     * @param list<string> $tags
+     * @param list<BackgroundNode|ScenarioInterface> $children
+     */
+    public function __construct(
+        private readonly ?string $title,
+        private readonly ?string $description,
+        private readonly array $tags,
+        private readonly array $children,
```
acoulton, line 25: Having a single `RuleNode->children`, which may or may not include a Background, is similar to the convention that cucumber/gherkin uses.

However it does slightly complicate typehinting and runtime type checks - it might be better to have separate `?BackgroundNode $background` and `ScenarioInterface $scenarios` as we do on `FeatureNode`?

On the other hand, using `$children` gives us more flexibility if Cucumber introduces other kinds of valid Rule children in future...

tests/Cucumber/CompatibilityTest.php
```diff
@@ -63,11 +63,6 @@ class CompatibilityTest extends TestCase
             'extra_blank_lines_everywhere.feature' => 'Background / Examples descriptions not supported',
         ],
         'gherkin-32' => [
-            'complex_background.feature' => 'Rule keyword not supported',
-            'rule.feature' => 'Rule keyword not supported',
-            'rule_with_tag.feature' => 'Rule keyword not supported',
-            'tags.feature' => 'Rule keyword not supported',
-            'rule_without_name_and_description.feature' => 'Rule is wrongly parsed as Description',
```
acoulton, line 70: All of the official cucumber/gherkin examples for `Rule` now parse as expected.

acoulton:
I've updated Behat/Behat#1879 to show that this is all still working as expected without any changes to Behat's runner.

acoulton:
This PR is now rebased onto current master and is ready for review.

src/Node/FeatureNode.php
```diff
@@ -125,11 +140,57 @@ public function hasScenarios()
     /**
      * Returns feature scenarios.
      *
+     * To provide backwards compatibility, this method will hoist any scenarios nested insideRules (losing any
+     * information about the Rule in the process). If the Rule had a Background, then any steps from that will be cloned
+     * and merged into each Scenario.
+     *
      * @return ScenarioInterface[]
+     *
+     * @deprecated use getExecutableChildren() for first-class handling of Rule nodes
```
carlos-granados: If we deprecate this function, users who are still in legacy mode should move to use `getExecutableChildren()`. But that function now returns list<RuleNode|ScenarioInterface>, so if you are in legacy mode and just want a list of scenarios you need to start filtering your list with instanceof or you'll start getting PHPStan errors. Not sure what is the right solution for this problem

src/Node/FeatureNode.php
carlos-granados: The PR description ask formatter authors to call FeatureNode::getChildren() but that function does not exist

src/Filter/AbstractFeatureFilter.php
```diff
@@ -90,5 +91,24 @@ private function filterFeatureChildren(FeatureNode $feature): FeatureNode
...
+        if ($filteredChildren === []) {
+            // Drop the rule, no scenarios match
+            return false;
+        }
+
+        if ($rule->hasBackground()) {
+            array_unshift($filteredChildren, $rule->getBackground());
+        }
+
+        return $rule->withChildren($filteredChildren);
```
carlos-granados: We should probably only do this if the list of filtered children does not match the list of executable children

src/Node/RuleNode.php
```diff
@@ -0,0 +1,111 @@
...
+ * This file is part of the Behat Gherkin Parser.
+ * (c) Konstantin Kudryashov <ever.zet@gmail.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Behat\Gherkin\Node;
+
+class RuleNode implements KeywordNodeInterface, DescribableNodeInterface, TaggedNodeInterface
```
carlos-granados: Should this class be final? 🤔

carlos-granados (commented):
@acoulton I was finally able to review this. Looking mostly good, just a few questions

src/Node/FeatureNode.php
acoulton: Updated in the description, it should be `FeatureNode::getExecutableChildren()`

src/Node/RuleNode.php
acoulton: Good catch, yes it should - fixed.

src/Filter/AbstractFeatureFilter.php
acoulton: Good spot - fixed and proved in the tests.

src/Node/FeatureNode.php
acoulton: IMO this is OK...

The caller does not usually know what gherkin parsing mode has been enabled, nor should they be depending on it. A formatter or filter provided by an extension, for example, could be used by some end-users in legacy mode and some in gherkin-32. They also don't know whether this specific feature contains any `Rule` nodes.

Broadly speaking, our current API relies on the caller understanding / following the tree structure of whatever Gherkin we parsed:

* The existing `getScenarios` doesn't actually give you **all** the "Scenarios" that the Feature will execute - it just gives you the Feature's direct children.
* Some of those might be `OutlineNode`. Arguably for most purposes they are **not** actually a "Scenario" - they're a group of "Scenarios" that will be executed independently. To find all the "Scenarios", you have to know that and call `OutlineNode::getExamples()` to traverse down to the `ExampleNode` children.
* You can only traverse the tree top-down. So if your code elsewhere accepts a `ScenarioInterface` and you get an `ExampleNode`, the only way to understand where it came from is to follow `FeatureNode` -> `OutlineNode` -> `ExampleNode`.
* All of this means there are actually very few cases where it's useful to iterate `FeatureNode::getScenarios()` without checking the explicit types of the `ScenarioInterface` you receive.

See for example the duplicate / fiddly logic in Behat's [TagExpressionFilter](https://github.com/Behat/Behat/blob/2b8d289917bd0bede95821530617bb4809578821/src/Behat/Behat/Gherkin/Filter/TagExpressionFilter.php#L46-L73)

The typehints also mean this isn't that clear to new / external users - `FeatureNode::getScenarios()` is typed as `ScenarioInterface[]`. But in reality it is only ever expected to return `list<ScenarioNode|OutlineNode>` - they are the only `ScenarioInterface` that are valid as direct children of a Feature.

So if anything, there's a good argument already for deprecating `getScenarios()` - it really only means `getExecutableChildren()`.

IMO we should also tighten the node types / use of interfaces throughout. But that would have to be for a major version because we first have to lock down the types you can use to create a `FeatureNode` - in theory at the moment someone with a custom `Parser` implementation could legally be passing some other type of custom `ScenarioInterface` implementation - although most code would have no idea what to do with that. Hence for now at least `getExecutableChildren` also has to be `array<ScenarioInterface|RuleNode>`.

If a caller sticks with the deprecated `getScenarios` for now, it will work as before (in either legacy or gherkin-32 parsing mode) - the only tradeoff is that if the Scenario was part of a Rule the caller will not get that information.

If a caller switches to `getExecutableChildren` then:
* Yes, as part of that their code will have to expect and possibly traverse into the new `RuleNode`. 
* But it's likely they already have to check for and traverse into `OutlineNode`, so this isn't a major new concept.
* It doesn't really matter what parsing mode is being used - if the file was parsed in `legacy` mode then we know it won't contain RuleNode, but it also wouldn't contain RuleNode if it was parsed in `gherkin-32` but didn't contain any Rules. So that shouldn't cause any problems.

There is probably a strong argument for a new interface / method / traverser to extract and iterate all the "leaf" Scenario nodes as a flat list, so that callers don't all have to know how to traverse the gherkin tree. That would be another step towards the cucumber "pickles" concept which exists for exactly this reason. But that would have to be a separate feature - and almost certainly a different kind of result object that carries some information about the node's parent(s).

Not sure how clearly I've explained all that, but hopefully makes sense?

acoulton:
Thanks for the review @carlos-granados - appreciate this is a chunky PR. I think I've answered all your points but any more questions / feedback very welcome :)

src/Node/FeatureNode.php
carlos-granados: Understood now, well explained

carlos-granados (approved):
With the changes that you added this is now looking good to me, thanks @acoulton

Harrisonbro:
Thanks so much for your work here, @acoulton 👏  As a user, I'm looking forward to being able to use `Rule`. Appreciate you and the Behat team.

Apologies if this is an obvious question, but do you have a sense of when this will be in a public release and available for use in projects?

acoulton:
@Harrisonbro thanks :)

I dislike being too specific about timings of open source releases, it's always hard to predict what might come up in "real life" and how that will impact availability to work on open source. Case in point, I was going to do a bit more Behat this evening then got sidetracked by my teenager managing to install an infostealer malware and get a whole bunch of his accounts compromised through cookie exfiltration... 😠

That said, there's only a couple of small PRs left before we can do a Gherkin release so I'd hope that'll be by early next week.

As soon as we release Gherkin, this will be available in Behat 3.x if you opt-in to the GHERKIN_32 parser compatibility mode. It'll also be available by default with Behat 4.0.0-alpha or 4.x-dev. 

The only caveat being that for the moment the Behat formatters (particularly pretty) won't know about / display Rule nodes, but the scenarios within Rules will all execute as expected. I hope it won't take too long to add support for that, but we'll probably prioritise getting 4.0.0 out first.

PR #431, opened 2026-09-09
test: Cover and document how multiple filters combine

### LINKED ISSUE #428: Behaviour with multiple filters is undefined and untested
As identified in [this discussion on Behat](https://github.com/Behat/Behat/pull/1863#discussion_r3902270095), our `Gherkin` class accepts multiple filters to use when loading features from a file. These can be injected into the class, or passed to the `load` method and merged in.

However, we don't appear to document or specify the expected behaviour when multiple filters are provided - is the filtering expected to be an AND or an OR?

I have only been able to find [one unit test that covers multiple filters](https://github.com/Behat/Gherkin/blob/master/tests/GherkinTest.php#L27), and it only really asserts the mocks are called, not the variants of how the filters could interact.

The implementation appears to be an AND condition, as the feature is skipped as soon as any filter doesn't match: https://github.com/Behat/Gherkin/blob/a30421614d271aa677033afe386758aa5e4841fa/src/Gherkin.php#L125-L138

We should improve test coverage and document this behaviour.

Amoifr:
Fixes #428. Tests and docblocks only, no behaviour change.

I measured the behaviour with real filters rather than reading it off the loop, which turned up four things worth pinning down:

1. Filters combine with a logical **AND**: `@smoke` and `Login` together keep only the scenario both match.
2. The result does not depend on the order the filters are applied in.
3. Filters passed to `load()` are merged with the injected ones and behave identically.
4. The part your issue does not mention: a filter that leaves no scenarios only drops the feature when it does not match the feature itself. And that reprieve is per filter, so a feature kept by a matching `@feature-tag` is still dropped by a later filter that finds nothing.

Eight cases in a data provider cover those, two of them for the fourth point. They use a real loader rather than mocks, since asserting the mocks were called is exactly what the existing `testLoader` already does and what you pointed at in the issue.

The docblocks of `load()` and `addFilter()` now state the same thing.

#### Testing

1066 tests against 1058 on `master`, same single skip and 22 incomplete. PHPStan reports no errors, php-cs-fixer is clean at 0 of 111.


acoulton (approved):
Great, thanks @Amoifr

stof (approved):

PR #433, opened 2026-09-11
refactor: Refactor filter implementations and tests, cover missed test cases

acoulton:
The tests for filtering were hard to follow, and in particular hard to see exactly which cases were and were not covered. Additionally, some tests were not sufficiently robust to detect unexpected differences in the structure / properties of the filtered nodes.

This becomes more of an issue once we introduce Rule support, where we need to know that the filter has not accidentally triggered the BC behaviour and stripped out the Rule nodes.

This branch refactors the TagFilter implementation, so that we can have one set of tests covering all the edges and legacy anomalies of our filter expressions. These can be separate from the filter itself, which really only needs to prove that the filter conditions take account of tags merged down the hierarchy.

Then, it refactors the tests to a new structure that:

* Uses data providers rather than asserting different filter patterns in a single test.
* Ensures that the behaviour of `filterFeature` and `isScenarioMatch`   are kept in sync, and any differences are made clear in the tests.
* Guarantees that every test covers the full shape and properties of the filtered node tree.
* Organises the examples around human-readable representations of the filtering behaviour.

Finally, it adds coverage for some cases that were not previously tested.

I will follow this with a separate PR that fixes some bugs that I found when reviewing and adding coverage for missing cases.

Once this is merged, the changes to filters (and their tests) in #430 will be much easier to implement and to review for potential BC impact.

### DIFF
--- a/src/Filter/TagFilter.php
+++ b/src/Filter/TagFilter.php
@@ -26,16 +26,11 @@ class TagFilter extends ComplexFilter
      */
     protected $filterString;
 
-    /**
-     * @var array{all?: list<array{any: list<array{tag: string, hasTag: bool}>}>, filterString: string}
-     */
-    private array $parsedFilter;
+    private TagFilterMatcher $filterMatcher;
 
     public function __construct(string $filterString)
     {
-        $this->filterString = trim($filterString);
-        $this->parseFilterString();
-
+        $this->filterMatcher = new TagFilterMatcher($filterString);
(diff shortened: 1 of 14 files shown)


acoulton:
The CI failure is down to an accidental use of `null` as an array key when a test scenario has no title - I'll fix that.

carlos-granados (approved):
great job @acoulton. I have not been able to review each individual tests but the general direction and reorganisation, with the use of generators and data providers, looks great and reads much more easily than the previous implementation

PR #434, opened 2026-09-11
fix: LineFilter and LineRangeFilter should handle varied Example tables

acoulton:
These filters still use the deprecated `OutlineNode::getExampleTable()` method, which merges all example tables from the Outline.

As a result, they throw if an Outline has tables with different column structures, despite this being valid Gherkin. 

This PR adds tests to prove this, before adding a fix.

The only other core filter that filters tables is TagFilter - that is currently working, I have added a testcase to prove it works into the future.


carlos-granados:
@acoulton now that #433 is approved, could you please merge it and rebase this PR? Thanks!

acoulton:
@carlos-granados I'm out today, had hoped to be able to rebase from the app but forgot you can't do that. I'll rebase later on when I'm home. Thanks for the reviews.

acoulton:
@carlos-granados this is rebased now, thanks

carlos-granados (approved):
Looks ok, with good test coverage, thanks @acoulton

PR #435, opened 2026-09-12
fix: Don't use our own deprecated code

acoulton:
There are some parts of the library that are calling (our own) methods that we have already deprecated. These may now be causing bugs - e.g. the filtering bug fixed in #434.

We should have phpstan check for and fail on use of deprecated methods.

Since we have very few external dependencies, it should be safe to include this in the main CI pass/fail result - it is fairly unlikely to fail unexpectedly.


acoulton:
Blocked on #434

carlos-granados (approved):
Makes sense, thanks @acoulton

PR #436, opened 2026-09-15
refactor: Internal implementations of Filter classes

acoulton:
This PR refactors out duplication and shared logic between our existing Filter implementations. The new internal API then makes it easier to modify the filtering logic to support Rule tags in the next PR.

The external API is fully backwards compatible - including for classes that

I have extracted this initial work to make the Rule branch diff easier to review.

Note that the Rule branch includes breaking changes to some of these new internal method signatures, so this PR needs to be in the same release as the Rule PR #430


src/Filter/AbstractFeatureFilter.php
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        return $this->filterFeatureChildren($feature);
+    }
+
+    private function shouldSkipFilteringChildrenIfFeatureMatches(): bool
+    {
+        // BC: The filterChildrenIfFeatureMatches property may not be initialised if a class has extended
+        // SimpleFilter or ComplexFilter before it had a constructor, because they will not be calling
+        // parent::__construct();
```
acoulton, line 46: This is slightly ugly, but is the only way to guarantee a backwards-compatible implementation for any third-party classes that are extending `ComplexFilter`, `SimpleFilter` or any of our own Filter classes.

Since the base classes never had a constructor in the past, implementations will not automatically be calling `parent::__construct` so we need to allow for the possibility that our private property has never been initialised and set it to a sane default based on the past behaviour of `SimpleFilter` or `ComplexFilter`.

src/Filter/AbstractFeatureFilter.php
```diff
@@ -0,0 +1,94 @@
...
+
+            if ($filteredChild !== false) {
+                $filteredChildren[] = $filteredChild;
+            }
+        }
+
+        return $originalChildren === $filteredChildren ? $feature : $feature->withScenarios($filteredChildren);
+    }
+
+    abstract protected function filterScenario(FeatureNode $feature, ScenarioInterface $scenario): ScenarioInterface|false;
```
acoulton, line 93: In some cases, `filterScenario` only calls `isScenarioMatch`. In others, the `filterFeature` loop also had custom logic for filtering out Example table nodes. 

In the next PR, we need to be able to run this logic for both `FeatureNode` children and `RuleNode` children - extracting it to a standalone `filterScenario` method makes that easier to achieve without `filterFeature` having to vary in every child class.

Existing (e.g. third-party) classes that have customised either `filterFeature` or `isScenarioMatch` will still produce the same results.

acoulton:
The lines missing coverage in AbstractFeatureFilter are the ones that throw exceptions in unexpected / essentially impossible edge cases if third-party code uses the class in an illegal way.

carlos-granados (approved):
As you said, this implementation is a bit ugly but agree that it's needed to keep BC. Thanks @acoulton

acoulton:
@carlos-granados thanks - I've updated this branch to resolve the conflicts with / fix the phpstan-ignore tags after merging #435

carlos-granados (approved):

